PR #501, opened 2026-07-22
fix(coding-agent): open resumed sessions with right arrow

snimu:
## Why

The resume-session picker supports left arrow as back, but right arrow does not open the highlighted session. This makes horizontal navigation inconsistent with the rest of the session-management UI.

## What changes

- treats right arrow at the end of the search field as selection, equivalent to Enter
- preserves ordinary right-arrow cursor movement while editing within a search query
- extends the existing searchable-selector navigation scenario rather than adding a duplicate selector fixture

## Validation

- `npx vitest --run test/selector-search-reset.test.ts` — 5 passed
- `npm run check`


snimu:
Superseded by a clean unified searchable session-view branch that will merge daemon-resident agents with saved sessions rather than extending the separate resume picker.

PR #548, opened 2026-07-28
docs(coding-agent): guide completed subagent cleanup

sethkarten:
## Summary

- tell the model to retain completed direct children when they may be reused
- tell the model to delete a completed direct child only once it is definitively no longer needed
- keep lifecycle cleanup explicit rather than automatically deleting reusable sessions

## Tests

- `test/system-prompt.test.ts`
- `npm run check`

Stacked on #547.

---

> [!NOTE]
> **Low Risk**
> Prompt and changelog-only change; no runtime or API behavior is modified.
> 
> **Overview**
> **Subagent lifecycle guidance** in `buildSubagentGuidance` now tells the model to **keep finished direct children** when they might be reused (e.g. via `agent_message`), and to call **`await rlm.delete_subagent(child)`** only after they are clearly no longer needed—making cleanup explicit instead of implying automatic removal of reusable sessions.
> 
> The **[Unreleased]** changelog entry and a **system-prompt test** assert the new sentence is present in the assembled prompt.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b8365ac821bfb3ddb8acf3078abce6f3585f9f2f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Retry deleted subagent runtime cleanup after compaction and guide subagent lifecycle
> - Adds `_reapDeletedRlmSubagentRuntimesAfterCompaction` to [`AgentSession`](https://github.com/PrimeIntellect-ai/prime-agent/pull/548/files#diff-a72e8a0b5f514ca1def756ee863a6e8a556df115e7167a50ca4f5e66b14448ae), which retries `deleteRlmSubagent` for any pending retryable deletions using `Promise.allSettled` after each compaction.
> - Updates [`buildSubagentGuidance`](https://github.com/PrimeIntellect-ai/prime-agent/pull/548/files#diff-64c77a41480409e73a6d87d381a618f83c44304c3677d755b3ef66d7a5ddce29) to advise retaining reusable completed direct children and explicitly deleting them with `rlm.delete_subagent` when no longer needed.
[...]

PR #549, opened 2026-07-28
fix(coding-agent): bound large session load memory

sethkarten:
## Summary

- stream daemon JSONL loads at 128 MiB and above instead of retaining the complete input Buffer beside the parsed entry graph
- keep the faster Buffer path for ordinary sessions
- preserve malformed-line handling, CRLF/UTF-8 parsing, final unterminated records, and full session-tree semantics

## Measured result

On a closed 623.9 MiB, 80,441-row session, using the production loader in isolated processes:

| path | load time | peak RSS |
| --- | ---: | ---: |
| Buffer | 1.05 s | 1.462 GB |
| streaming | 2.12 s | 616 MB |

The large-file path reduces peak RSS by 846 MB (58%). It trades clean-cache parse latency for protection from GC/swap/OOM slowdown; files below 128 MiB retain the current fast path.

`buildSessionContext()` measured 21 ms on the same session, so it was not changed. Completed-child JSONL parsing measured 33 ms total for the target parent and was also not changed.

## Tests

- `test/session-manager/file-operations.test.ts`
- `npm run check`

Stacked on #548.

---

> [!NOTE]
> **Medium Risk**
> Touches core session load and list-scan paths used by the daemon; behavior is heavily tested for parity, but very large sessions pay ~2× parse time for much lower RSS.
> 
> **Overview**
> **Large daemon session loads** no longer read the entire JSONL file into memory before parsing. `loadEntriesFromFileAsync` uses a **128 MiB** size threshold: smaller files still use `readFile` plus async buffer parsing with periodic `setImmediate` yields; at or above the threshold it **streams** lines via a new `readLinesAsBuffers` generator so the full file buffer is not held alongside the parsed entry graph.
> 
> Parsing is refactored around shared **`appendEntryFromBuffer`** / **`parseEntriesFromBufferAsync`**, and **`scanSessionInfo`** (session list metadata) now uses the same line streaming instead of `createReadStream` + `readline` UTF-8 strings.
> 
> Tests cover streaming parity with the sync loader, LF-only record boundaries, multi-megabyte records, event-loop yielding, and pending-chunk cleanup in `readLinesAsBuffers`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit dfffe35de80405e033acb077776e1162e53c168a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Bound memory usage during large session loads in coding agent
> - `loadEntriesFromFileAsync` in [session-manager.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/549/files#diff-e514ecbe1ca80031cc52600ee3863d993a99616f116355f3fec33c92b532d2cf) now streams JSONL lines via a new `readLinesAsBuffers` generator for files above a configurable byte threshold, avoiding holding both the full input buffer and parsed entries in memory simultaneously.
> - A new async generator `readLinesAsBuffers` in [file-lines.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/549/files#diff-29279bea377c244202a98bab59966dd405681b64b4e72e18716015effe92b8a3) yields each line as a `Buffer` from a `createReadStream`, handling multi-chunk and EOF-terminated lines.
> - `AgentSession.compact` now retries pending subagent runtime deletions after compaction via a new `_reapDeletedRlmSubagentRuntimesAfterCompaction` helper, fixing a case where explicitly deleted subagents were not cleaned up.
> - The subagent system prompt now instructs models to retain reusable completed direct children and delete them explicitly once no longer needed.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized dfffe35.</sup>
> 
>
>


packages/coding-agent/src/utils/file-lines.ts
```diff
@@ -33,3 +33,34 @@ export function readFirstLineSync(filePath: string, maxBytes = 64 * 1024): strin
...
+			if (end === -1) {
+				const part = buffer.subarray(start);
+				pendingParts.push(part);
+				pendingBytes += part.length;
+				break;
+			}
+			if (pendingParts.length > 0) {
+				const part = buffer.subarray(start, end);
+				pendingParts.push(part);
+				yield Buffer.concat(pendingParts, pendingBytes + part.length);
```
snimu: [prime-agent review] **High — large individual records retain two raw copies while parsing**

When a JSONL record spans stream chunks, the generator pauses at `yield` before `pendingParts` is cleared. During the consumer’s UTF-8 decode and `JSON.parse`, the original chunk slices remain referenced alongside the concatenated line buffer. The EOF path has the same behavior.

For a session dominated by one enormous tool result/message, peak memory can therefore include the original chunks, a concatenated raw copy, the decoded string, and the parsed object graph, substantially weakening this PR’s OOM protection.

Please assign the concatenated buffer to a local, clear `pendingParts`/`pendingBytes`, and only then yield it (including the EOF path).

snimu (commented):
[prime-agent review] Reviewed the current #549 head for streaming-loader correctness and memory behavior. One issue found inline.

packages/coding-agent/src/utils/file-lines.ts
sethkarten: Fixed in `1345bca84`: chunk references are cleared before yielding concatenated records, including the EOF path, with focused lifetime regression coverage.

snimu (approved):

PR #560, opened 2026-07-30
fix(coding-agent): stop blocking startup on private Prime Inference model authorization

xeophon:
## Problem

Startup was spending ~350-600ms (up to the 10s request timeout on a slow network) on the critical path of session creation: `refreshAvailableModels()` awaited a live request to the Prime Inference catalog to authorize private models on every call, including the initial model resolution during startup. It also ignored `PI_OFFLINE`.

Measured with `PI_TIMING=1` and the `scripts/profile-coding-agent-node.mjs` RPC benchmark:

- before: `findInitialModel` ~370ms of a ~920ms startup
- after: ~12ms; benchmark median 917.7ms -> 552.3ms (unbundled `dist/cli.js`), bundled CLI wall time ~700ms -> ~330ms

## Change

The authorization decision is now persisted next to `models.json` (`prime-inference-private-models.json`), fingerprinted by API key + team id:

- matching, fresh cache (5 min TTL): served without any network call
- matching, stale cache: ids are served immediately from cache and a deduplicated background refresh (3s timeout) updates the in-memory set and the cache file for subsequent lookups
- missing/mismatched cache (first login, key/team switch): synchronous fetch, same as before, then persisted
- `PI_OFFLINE`: never fetches; serves the persisted cache if credentials match

Failure semantics are unchanged: on fetch error the previous in-process ids are kept for the same team.

## Tests

- added `test/model-registry-private-cache.test.ts` covering cold-start persistence, fresh cache hit without network, stale-cache background refresh, credential-mismatch invalidation, and offline behavior
- existing ENG-4645 private-model regression suite and model-resolver tests pass

Two `daemon-mode` session-lease tests fail identically on pristine HEAD in this environment; unrelated to this change.

---

> [!NOTE]
> **Medium Risk**
> Changes which private models appear in pickers and at cold start may briefly use stale authorized ids until background refresh completes; cache files hold model id lists tied to credential fingerprints on disk.
> 
> **Overview**
> Startup and model refresh no longer **always await** the Prime Inference `/models` call to decide which `internal/*` models are authorized. That decision is **persisted** in `prime-inference-private-models.json` next to `models.json`, keyed by a SHA-256 fingerprint of API key and team id.
> 
> On a **matching cache**: authorized ids load synchronously from disk. If the entry is still within the **5-minute TTL** (or **`PI_OFFLINE`** is set), no network runs. **Stale** matches still apply cached ids immediately and kick off a **deduplicated background refresh** with a **3s** timeout; successful refreshes update memory and rewrite the cache atomically (`0o600` temp + rename). **Missing or mismatched** cache (first login, key/team change) still does a **blocking fetch** like before, then persists. Offline with no valid cache clears private authorization instead of fetching.
> 
> `fetchAuthorizedPrivatePrimeInferenceModelIds` now takes an optional **`timeoutMs`** (default unchanged at 10s) so background refresh can use the shorter limit. Fetch failures still fall back to the previous in-process ids for the same team when applicable.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 784d1a0dc22e045a35caba1805f013c3b50c8c42. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Stop blocking startup on private Prime Inference model authorization
> - Authorization results are now cached to disk (adjacent to `models.json`) with a 5-minute TTL. On startup, cached ids are served immediately if the credentials fingerprint matches, avoiding a blocking network call.
> - Stale caches trigger a background refresh with a 3-second timeout, de-duplicated by credential fingerprint so concurrent callers share one in-flight request.
> - `PI_OFFLINE` env var (accepts `1`/`true`/`yes`) skips all network calls; if a matching cache exists it is used, otherwise authorization is cleared.
> - On fetch failure, the registry falls back to previously authorized ids for the same team rather than clearing state, unless the team has changed.
> - The cache file is written atomically via a PID-suffixed temp file and `renameSync`, with mode `0600`.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 784d1a0.</sup>
> 
>


xeophon:
Addressed both cc/macroscope findings:

- Background-refresh dedupe is now keyed by credential fingerprint; a refresh for changed credentials is chained after the in-flight one instead of being dropped.
- The apply guard now compares the full API key + team fingerprint (recomputed at completion) instead of just the team id, so a key rotation mid-flight can't have a stale result overwrite the newer authorization or cache.

The models.generated.ts churn was already reverted in 7ef74b58.

snimu (approved):

PR #592, opened 2026-08-03
fix(coding-agent): echo salvaged command id in daemon parse failures

snimu:
## Summary

Sandbox validation for the v0.5.0 release (#591) found that a v0.4.0 client talking to a protocol-7 daemon hangs for 30s with a generic timeout instead of surfacing the protocol rejection.

The daemon does reply promptly with `Daemon commands require protocol 7 or newer`, but the parse-failure path writes `failure(undefined, "parse", ...)`. Old clients only resolve pending requests by response id (`daemon-client.ts` ignores id-less responses), so the rejection is never correlated and the request waits out its full timeout.

## Change

- `salvageDaemonCommandId(line)` in `daemon-protocol.ts`: best-effort id extraction from the raw line, deliberately ignoring protocol-version validity, returning `undefined` for malformed JSON or missing/non-string ids.
- Both parse-failure paths (`daemon-supervisor.ts` `handleLine` and `daemon-mode.ts` `handleLine`) echo the salvaged id instead of `undefined`.

No wire-shape change: `id` is already an optional `DaemonResponse` field that all released clients correlate on. No protocol or schema bump needed.

## Tests

- `daemon-protocol.test.ts`: unit coverage for the salvage helper (v6 envelope, bare command, non-string id, missing type, non-object, malformed JSON).
- `daemon-supervisor-side-question.test.ts`: the two existing protocol-6 rejection tests now assert the echoed id.

With this, a 0.4.0 client hitting a 0.5.0 daemon sees the protocol error in milliseconds instead of a 30s timeout, making the 0.5.0 changelog's "rejected cleanly at connect" true in both directions.

---

> [!NOTE]
> **Low Risk**
> Small behavioral fix on error paths only; no change to successful command handling or protocol version.
> 
> **Overview**
> Fixes a compatibility gap where **daemon parse failures** (e.g. protocol-6 clients on a protocol-7 daemon) were answered promptly but with **no command `id`**, so older clients ignored the response and **timed out** instead of showing the protocol error.
> 
> Adds **`salvageDaemonCommandId`** to best-effort read a string `id` from the raw JSON line even when the command is otherwise invalid. **`AgentDaemon.handleLine`** and **`DaemonSupervisor.handleLine`** now pass that id into `failure(..., "parse", ...)` instead of `undefined`. No wire or schema bump—`id` was already optional on responses.
> 
> Tests cover the helper and assert echoed ids on existing protocol-6 rejection scenarios; changelog documents the fix.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 15ee7ab7d37a4ea8eee72d48e416a85f9a2ee64e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Echo salvaged command id in daemon parse failure responses
> When a daemon socket line fails JSON parsing, the failure response previously dropped the command id. Now both `AgentDaemon.handleLine` and `DaemonSupervisor.handleLine` call the new `salvageDaemonCommandId` utility to best-effort extract the `id` field from the raw line and include it in the failure response when recoverable.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 15ee7ab.</sup>
> 
>

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [Unreleased]
 
+- Fixed daemon parse rejections dropping the command id, which left older clients waiting for a timeout instead of seeing the protocol error.
 - Changed large daemon session loads to stream JSONL history and avoid retaining a second full-file copy in memory.
 - Changed the agents view to render explicit session names in bold and the "(no messages)" placeholder in italics.
 - Fixed the blank line between the recap and the working hint so they render directly above each other.
--- a/packages/coding-agent/src/modes/daemon/daemon-mode.ts
+++ b/packages/coding-agent/src/modes/daemon/daemon-mode.ts
@@ -149,6 +149,7 @@ import {
 	isDaemonCommandEnvelope,
 	isDaemonDialogExtensionUiRequest,
 	isDaemonMutatingCommand,
+	salvageDaemonCommandId,
 	success,
 	UPDATE_RESTART_DRAIN_COMMANDS,
 } from "./daemon-protocol.js";
@@ -2660,7 +2661,7 @@ export class AgentDaemon {
 			}
 			command = parsed as DaemonCommand;
 		} catch (error) {
-			this.write(client, failure(undefined, "parse", error, serializeDaemonError(error)));
+			this.write(client, failure(salvageDaemonCommandId(line), "parse", error, serializeDaemonError(error)));
 			return;
 		}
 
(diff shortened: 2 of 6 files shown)


xeophon:
`salvageDaemonCommandId` still drops a valid request id when `type` is missing or non-string. Since this helper runs after command parsing or validation has already failed, an invalid `type` is one of the cases that should remain correlatable. For example, `{"id":"daemon_1","type":null}` gets an id-less failure from the supervisor, so a client waiting on `daemon_1` still times out.

Could we return the id whenever `typeof candidate.id === "string"` and update the `no-type` expectation accordingly? The exact v0.4 envelope is fixed as-is, but this would make the helper match its stated parse-failure behavior.

snimu:
Agreed - the type check contradicted the helper's purpose. It only runs after parsing/validation already failed, so the id is the one field that matters for correlation.

Fixed in 15ee7ab7d: the helper now returns the id whenever `typeof candidate.id === "string"`, the `{"type":null,"id":...}` and no-type cases now salvage, and the doc comment states the id-only contract.

xeophon (approved):

PR #593, opened 2026-08-03
fix(coding-agent): surface seeded --goal objective to the model

snimu:
## Summary

Headless testing during the v0.5.0 release validation found that `-p --goal "x" "prompt"` runs with an active goal the model knows nothing about: a follow-up `-p -c "Is there an active goal?"` answers No.

Root cause: the model's only source of goal visibility is the `goal_context` message (there is no goal block in the system prompt). `/goal` injects it via `_runOrQueueGoalContext` after `_startGoal`. CLI `--goal` seeding in the `AgentSession` constructor calls only `_startGoal` - state is persisted, accounting runs, budget limits fire, the interactive footer shows the goal - but no context message is ever injected, so the model never sees the objective on the first turn or any continuation.

## Change

Constructor seeding now pushes the same `goal_context` message (kind `continuation`, same as first-time `/goal`) into `_pendingNextTurnMessages`, the existing next-turn context bucket already used from the constructor by `_restoreLateIpythonSentAgentMessages`. It is construction-safe (no action admission mid-constructor), delivered with the first turn under the turn's execution policy, and persisted to the transcript as a `custom_message` entry - which is what makes the goal visible to `-c` continuations too (verified against a reopened session file).

One line of behavior plus a comment; no changes to `/goal`, goal continuation, or budget paths.

## Tests

Extended the existing `seeds an active goal from initialGoal config` suite test: after the first prompt, a `goal_context` message containing the objective must appear before the first assistant message and be part of the agent context. Verified the assertion fails without the fix. Full goal suite passes (42), `npm run check` clean.

## Note

Flagged as high-priority pre-work for the recursive-agent-harness plan: its PR 3 copies the goal persistence/seeding pattern for `/rlm-max-depth`, so the pattern should be correct before it is cloned (see `features/bug-reports/headless-cli-issues.md` #7).

> [!NOTE]
> ### Fix `--goal` sessions to surface the seeded objective to the model
> - When an `initialGoal` seeds a top-level `AgentSession`, a `goal_context` message is now pushed to `_pendingNextTurnMessages` so the model sees the objective before the first assistant reply.
> - `_clearQueuedGoalContexts` is extended to also filter pending goal context messages from `_pendingNextTurnMessages`, so clearing a goal before the first prompt prevents delivery.
> - Tests in [agent-session-goal.test.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/593/files#diff-f831cdd3806262a4b6e6b6cb82279e590e00695be22ea012ba4b216eb4b5df59) verify ordering, content, inclusion in `currentAgentContext`, and the cleared-goal case.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 3287807.</sup>
> 
>

---

> [!NOTE]
> **Low Risk**
> Narrow goal-seeding and queue-cleanup change with targeted tests; no auth, persistence, or scheduling core changes.
> 
> **Overview**
> **Fixes `--goal` sessions where the model never saw the objective.** Goal state was persisted and shown in the UI, but unlike `/goal`, CLI seeding did not inject the `goal_context` custom message the model relies on.
> 
> After `_startGoal` in the constructor, the same `goal_context` message used elsewhere (`continuation` kind) is pushed onto `_pendingNextTurnMessages` so it ships with the first turn and lands in the transcript for `-c` continuations. **`_clearQueuedGoalContexts`** now drops those pending messages too when the goal is cleared before the first prompt.
> 
> Tests assert the objective appears in context before the first assistant reply and that clearing the goal removes pending goal context.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3287807eb345d343b1aca14c94f5b3923b03cd0f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [Unreleased]
 
 - Fixed daemon parse rejections dropping the command id, which left older clients waiting for a timeout instead of seeing the protocol error.
+- Fixed `--goal` sessions never showing the objective to the model, which made seeded goals invisible to first turns and continuations.
 - Changed large daemon session loads to stream JSONL history and avoid retaining a second full-file copy in memory.
 - Changed the agents view to render explicit session names in bold and the "(no messages)" placeholder in italics.
 - Fixed the blank line between the recap and the working hint so they render directly above each other.
--- a/packages/coding-agent/src/core/agent-session.ts
+++ b/packages/coding-agent/src/core/agent-session.ts
@@ -1625,6 +1625,9 @@ export class AgentSession {
 	}
 
 	private _clearQueuedGoalContexts(): void {
+		this._pendingNextTurnMessages = this._pendingNextTurnMessages.filter(
+			(message) => message.customType !== GOAL_CONTEXT_CUSTOM_TYPE,
+		);
 		this.agent.removeQueuedMessages(
 			(message) => message.role === "custom" && message.customType === GOAL_CONTEXT_CUSTOM_TYPE,
 		);
--- a/packages/coding-agent/src/core/agent-session.ts
+++ b/packages/coding-agent/src/core/agent-session.ts
@@ -1248,13 +1248,8 @@ export class AgentSession {
 		// or restart/rehydration of a session that already has messages or a goal.
 		if (this._rlmDepth === 0 && config.initialGoal && this._isBranchSeedable()) {
 			this._goalState = this._startGoal(config.initialGoal.objective, config.initialGoal.tokenBudget);
-			// The goal-context message is the only way the model learns a goal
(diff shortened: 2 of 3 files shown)


snimu:
Valid - both this and the Macroscope finding. All stale-context triggers (clear, pause, replace, host completion) funnel through `_clearQueuedGoalContexts`, which swept the agent queue and session actions but not `_pendingNextTurnMessages`, where constructor seeding parks the context.

Fixed in 3287807eb: the sweep now also filters `goal_context` messages out of `_pendingNextTurnMessages`, with a regression test covering clear-before-first-prompt.

alexzhang13 (approved):

PR #598, opened 2026-08-04
fix(coding-agent): derive refine output budgets from the model

sethkarten:
`/refine` failed with:

```
Error: Refinement failed: Expected ',' or ']' after array element in JSON at position 8741 (line 31 column 6)
```

## Cause

`planRefinement` builds its request from four blocks and then caps the reply at a fixed 4096 tokens:

| Block | Bound | Measured on a real harness (59 entries, 51 refinements) |
|---|---|---|
| system prompt | fixed | ~1,188 tok |
| `<current_harness_state>` | 40 entries/kind, 240 chars each | ~5,691 tok |
| `<refinement_history>` | last 20 records | ~207 tok |
| `<conversation>` | `.slice(-80_000)` chars | ~20,000 tok |

That is ~27k tokens of input against a 4,096-token output cap, a 6.6:1 ratio. Every input block grows with harness size; the output cap was a constant chosen when the harness was empty.

Reconstructing per-edit JSON from real entries: median 202 tok, p90 347 tok, max 737 tok. Typical proposals carry 1-3 edits and fit comfortably. Large ones do not: 5 max-sized edits land at ~3.9k, 8 at ~6.1k. The cap does not bite on routine refinements, only on the consequential multi-edit ones.

Two things then turned truncation into an unreadable error:

1. `stopReason === "length"` was never checked. Only `"error"` was handled, so a truncated-but-successful reply went straight to parsing.
2. `extractJsonObject`'s brace-slicing fallback is meant to recover JSON wrapped in prose. On a reply truncated inside the `edits` array it slices to the *previous* edit's closing brace, producing a fragment that is genuinely invalid JSON. The reported parse error describes that fragment, not the truncation.

## Changes

- Derive both budgets from the model: `min(model.maxTokens, 32_000)` for the proposal, `min(model.maxTokens, 4_096)` for the auto-refine review. 32k is ~5x the realistic worst case and ~25% of a 128k-output model. `maxTokens` is a ceiling, not a reservation, so unused budget costs nothing.
- Throw a clear error on `stopReason === "length"` at both call sites.
- Make `extractJsonObject` validate its brace-sliced candidate instead of returning a fragment that cannot parse.

## Tests

`test/refinement.test.ts`, 57 passed. Three new cases: the budget is capped by the model's own `maxTokens`; a truncated reply reporting `length` produces the budget error; a truncated reply that never reports `length` is rejected by the parser rather than surfacing a fragment error. One existing assertion moved from the `4096` literal to the model-derived `8192`.

`npm run check` passes.

## Not included

This fixes the output side only. The input side has a related problem worth a separate change: the 40-entry cap is already reached at 40 memories, and the 240-char truncation hides ~62% of entry content from the refiner, which is how near-duplicate entries get created. The better fix is an index-style overview plus direct harness lookups during refinement rather than a larger one-shot dump.

---

> [!NOTE]
> **Low Risk**
> Localized to refinement LLM calls and JSON parsing; behavior change is larger output budgets on capable models and clearer errors on truncation, with no auth or data-path impact.
> 
> **Overview**
> Fixes `/refine` failures where large harness proposals hit a **fixed 4096-token output cap** and surfaced misleading `JSON.parse` errors instead of explaining truncation.
> 
> **Output budgets** in `planRefinement` and auto-refine review now use `min(model.maxTokens, ceiling)` — **32k** for refinement proposals and **4k** for the review gate — replacing hardcoded **4096** and **1024**.
> 
> **Truncation handling** treats `stopReason === "length"` as an explicit exhausted-budget error at both LLM call sites. **`extractJsonObject`** gains incomplete-JSON detection (`isIncompleteJson`, `parseJsonCandidate`) so cut-off replies and brace-slicing artifacts report budget exhaustion or invalid JSON distinctly, not opaque parse failures on partial objects.
> 
> Tests cover model-derived caps, the 32k ceiling on wide models, length stop reason, silent truncation, and malformed vs truncated JSON.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ae16e3085ff2fd888ac29fc5be566680c34d9f0d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Fix `/refine` output budget to derive from model capacity instead of fixed 4096-token cap
> - Replaces the hardcoded 4096-token output limit in `planRefinement` with a model-derived cap (up to 32000 tokens), and the 1024-token limit in auto-refine review with a model-derived cap (up to 4096 tokens), both in [refinement.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/598/files#diff-17b614d876315ad8ac0165562a8f07a5c4ada3b28928eb02d90f6b93a96ea91d).
> - Adds `isIncompleteJson` and `parseJsonCandidate` helpers to distinguish truncated JSON from malformed JSON, so truncation now surfaces a clear "exhausted budget" error instead of an opaque parse failure.
> - Adds a `stopReason === 'length'` check in both refine paths to throw an explicit truncation error before attempting to parse the response.
> - Behavioral Change: models with fewer than 32000 max tokens will now use their actual limit; replies that previously failed silently with a JSON parse error will now throw a descriptive truncation message.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized ae16e30.</sup>
> 
>

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- Fixed `/refine` failing with an opaque JSON parse error when the refiner exceeded a fixed 4096-token output cap; output budgets now derive from the selected model, and a truncated reply reports the exhausted budget directly.
+
 ## [0.5.0] - 2026-08-03
 
 ### Breaking Changes
--- a/packages/coding-agent/src/core/refinement/refinement.ts
+++ b/packages/coding-agent/src/core/refinement/refinement.ts
@@ -184,6 +184,26 @@ Return JSON only:
   "instructions": "optional concise instructions for /refine if shouldRefine is true"
 }`;
 
+/**
+ * Output budgets are derived from the selected model instead of fixed literals.
+ * /refine input scales with harness size (entry overview, refinement history, and
+ * the trajectory slice), so a constant output cap silently truncates exactly the
+ * large multi-edit proposals that matter most. Math.min keeps small models honest.
+ */
+const REFINEMENT_MAX_OUTPUT_TOKENS = 32_000;
+const AUTO_REFINE_REVIEW_MAX_OUTPUT_TOKENS = 4_096;
+
+const TRUNCATED_JSON_ERROR =
+	"the model stopped before completing its JSON object. This usually means the output budget was exhausted; retry with a smaller request.";
+
+function refinementMaxOutputTokens(model: Model<any>): number {
+	return Math.min(model.maxTokens, REFINEMENT_MAX_OUTPUT_TOKENS);
+}
+
+function autoRefineReviewMaxOutputTokens(model: Model<any>): number {
+	return Math.min(model.maxTokens, AUTO_REFINE_REVIEW_MAX_OUTPUT_TOKENS);
+}
+
 function now(): string {
 	return new Date().toISOString();
 }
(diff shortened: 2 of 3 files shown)


sethkarten:
Both findings were valid and are fixed in ae16e308.

The Cursor finding was the more serious of the two: a reply truncated after a nested closing brace takes the `startsWith`/`endsWith` fast path, so it never reached the new guard and still produced the original `Expected ',' or ']' after array element` error. Reproduced against the real function before fixing.

The Macroscope finding is also correct: the fallback blamed the output budget for any parse failure, so a complete-but-malformed reply like `Here is the result: {"edits": [oops]}` would wrongly tell the user to retry with a smaller request.

Both come from the same root cause: the code was inferring the reason from *where* parsing failed rather than from the candidate itself. It now scans for an unterminated string or unclosed object/array. Incomplete reports the exhausted budget; balanced-but-invalid reports malformed JSON. Applied on every path, not just the slicing fallback.

Two regression tests added: a JSON-only reply truncated after a nested brace, and a balanced malformed reply. 59 tests pass, `npm run check` clean.

alexzhang13 (approved):

PR #605, opened 2026-08-04
fix(coding-agent): say something when print mode gets no prompt

snimu:
## What was broken

Running `prime-agent -p` with nothing on stdin (for example `echo -n | prime-agent -p`) printed nothing at all and exited successfully. If a script accidentally fed an empty prompt, there was no hint that anything went wrong - it just looked like a run that produced no output.

## The fix

Print mode now writes one line to stderr - `No prompt provided (stdin was empty); nothing to do.` - and still exits with code 0. The exit code stays 0 on purpose: existing tests pin that behavior three times over, and scripts may rely on it. Stdout stays empty so pipelines are unaffected.

## Testing

New test asserts the warning appears on stderr, stdout stays empty, and the exit code stays 0. Verified by hand: before the fix, total silence; after, the warning. The three existing empty-input tests pass unchanged.

---

> [!NOTE]
> **Low Risk**
> Small startup guard with no model or session logic changes; behavior is additive (stderr message) while preserving exit code 0 and empty stdout.
> 
> **Overview**
> **Fixes silent success when headless runs have nothing to send.** After `prepareInitialMessage`, **print**, **json**, and daemon-backed non-RPC clients bail out early if there is neither an `initialMessage` nor CLI `messages`, instead of continuing into a no-op or replaying prior output.
> 
> The user sees **`No prompt provided (stdin was empty); nothing to do.`** on **stderr**; **stdout** stays empty and the process still exits with **code 0** so existing scripts and the empty-input regression matrix stay unchanged.
> 
> The same guard is applied on both the **daemon client** path and the **in-process** print/json path in `main.ts`. New tests cover **`--print`** with empty stdin and **`--print --continue`** after a real turn, asserting the warning and that the last assistant reply is not printed again.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3cad67b2c1216a5e9cfec4974d61b7feef96f2aa. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Fix print mode silently exiting when no prompt is provided
> Adds an early guard in `runPrintModeWithConnectionInternal` that detects when neither `initialMessage` nor `messages` are present. Instead of silently exiting, it now writes a warning to stderr (`No prompt provided (stdin was empty); nothing to do.`) and returns exit code 0. Two regression tests cover both `--print` with no prompt and `--print --continue` with no prompt.
>
> 
> #### Changes since #605 opened
>
> - Added early-return guards in the `main` function to validate empty prompts [3cad67b]
> - Removed empty-input validation from `runPrintModeWithConnectionInternal` function [3cad67b]
> 
>
> 
>
> <details>
> <summary>📊 <a href="https://app.macroscope.com">Macroscope</a> summarized 3cad67b. 1 file reviewed, 0 issues evaluated, 0 issues filtered, 0 comments posted</summary>
>
> ### 🗂️ Filtered Issues
> No issues evaluated.
>
>
> </details>
>
>

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## [Unreleased]
 
+- Fixed print mode silently exiting when no prompt was provided through arguments or stdin.
+
 ## [0.5.1] - 2026-08-04
 
 ### Fixed
--- a/packages/coding-agent/src/main.ts
+++ b/packages/coding-agent/src/main.ts
@@ -1477,6 +1477,10 @@ export async function main(args: string[], options?: MainOptions) {
 			stdinContent,
 		);
 		time("prepareInitialMessage");
+		if (appMode !== "rpc" && !initialMessage && parsed.messages.length === 0) {
+			console.error("No prompt provided (stdin was empty); nothing to do.");
+			return;
+		}
 		initTheme(settingsManager.getTheme(), false);
 		time("initTheme");
 
@@ -1571,6 +1575,10 @@ export async function main(args: string[], options?: MainOptions) {
 		stdinContent,
 	);
 	time("prepareInitialMessage");
+	if ((appMode === "print" || appMode === "json") && !initialMessage && parsed.messages.length === 0) {
+		console.error("No prompt provided (stdin was empty); nothing to do.");
+		return;
+	}
 	initTheme(settingsManager.getTheme(), appMode === "interactive");
 	time("initTheme");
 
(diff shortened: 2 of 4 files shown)


xeophon:
prime-agent -p -c or -p -r … with empty stdin bypasses the warning and may print the previous assistant response as if it were new output.

xeophon (approved):

snimu:
Confirmed - reproduced exactly that: `-p "Say X"` then `echo -n | prime-agent -p -c` printed X again as if it were fresh output. The guard's history condition (added so resumed sessions would "still work") was wrong: with no new prompt there is nothing to run, and what looked like working was the completion path re-printing the previous run's terminal message.

Fixed in 2faaba536: the warning now applies regardless of session history (the history check is simply dropped, which also makes the guard synchronous). Regression test covers the continue case: first run produces output, `--continue` with empty stdin produces empty stdout, the warning, and exit 0.

PR #608, opened 2026-08-04
fix(coding-agent): show why the daemon crashed instead of a generic timeout

snimu:
## What was broken

When the background daemon crashed immediately on startup, the person launching it saw only `Timed out waiting for daemon to start` - after waiting the full 30 seconds. The actual error was thrown away: the daemon was started with its output discarded, and it crashed before it ever got to write a log line. A recent bug report shows the cost: a user whose daemon was crashing on startup had to bisect the problem themselves because every command just said "timed out".

## The fix

Two small changes:

1. The daemon now writes startup failures to its log before exiting, so crashes leave a trace on disk.
2. The launcher watches whether the spawned daemon is still alive. If it dies during startup, the launcher waits a further two seconds in case a concurrently started daemon is about to take over the socket (in which case the exit is harmless), and otherwise fails right away - no 30-second wait - with the exit code and the tail of the daemon's log. Only log content written after this launch is shown, so an old crash is never blamed for a new one. If startup genuinely times out, the same log tail is appended to the timeout message.

An earlier version of this PR captured the daemon's stderr through a pipe instead; that turned out to break healthy daemons (writing to the closed pipe killed them once the CLI exited) and was replaced by the log-based approach.

## Testing

New tests: a daemon that exits immediately produces a launcher error containing the exit code and its log output well before the timeout; a daemon that crashes before logging anything reports that honestly instead of showing a previous run's log (including with multi-byte content in the old log); a daemon that loses the startup race to an already-serving daemon does not fail the launch. Verified by hand with a deliberately broken setup: before the fix, 31 seconds of silence and a generic timeout; after it, the real error ("Invalid daemon supervisor owner record: <path>") in about 2 seconds. Full launcher test file passes (25/25).

---

> [!NOTE]
> **Low Risk**
> Localized to daemon launch/readiness and supervisor startup error logging; behavior change is clearer errors with a short race grace window.
> 
> **Overview**
> Fixes **daemon cold start** so a crash no longer surfaces only after a 30s `Timed out waiting for daemon to start`.
> 
> The launcher records the daemon log size before spawn, listens for **spawn errors** and **early child exit**, and after a **2s grace** (so a losing concurrent launcher can still win the socket) fails fast with **exit code/signal** or spawn error plus a **log tail from bytes written this attempt only**—so stale log lines from prior runs are not blamed. Genuine timeouts append the same tail.
> 
> The **daemon supervisor** now logs startup failures (stack/message) to the rotating daemon log before cleanup/rethrow.
> 
> **Tests** cover log tail on exit, silent crash messaging, spawn ENOENT, startup race success, and stale-log isolation.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 68930748e744c0601f7fad87ac4d7e1037f67acf. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ## What was broken
>
> When the background daemon crashed immediately on startup, the person launching it saw only `Timed out waiting for daemon to start` - after waiting the full 30 seconds. The actual error was thrown away: the daemon was started with its output discarded, and it crashed before it ever got to write a log line. A recent bug report shows the cost: a user whose daemon was crashing on startup had to bisect the problem themselves because every command just said "timed out".
>
> ## The fix
>
> Two small changes:
>
> 1. The daemon now writes startup failures to its log before exiting, so crashes leave a trace on disk.
> 2. The launcher watches whether the spawned daemon is still alive. If it dies during startup, the launcher waits a further two seconds in case a concurrently started daemon is about to take over the socket (in which case the exit is harmless), and otherwise fails right away - no 30-second wait - with the exit code and the tail of the daemon's log. Only log content written after this launch is shown, so an old crash is never blamed for a new one. If startup genuinely times out, the same log tail is appended to the timeout message.
>
> An earlier version of this PR captured the daemon's stderr through a pipe instead; that turned out to break healthy daemons (writing to the closed pipe killed them once the CLI exited) and was replaced by the log-based approach.
>
> ## Testing
>
> New tests: a daemon that exits immediately produces a launcher error containing the exit code and its log output well before the timeout; a daemon that crashes before logging anything reports that honestly instead of showing a previous run's log (including with multi-byte content in the old log); a daemon that loses the startup race to an already-serving daemon does not fail the launch. Verified by hand with a deliberately broken setup: before the fix, 31 seconds of silence and a generic timeout; after it, the real error ("Invalid daemon supervisor owner record: <path>") in about 2 seconds. Full launcher test file passes (25/25).
>
> 
> ---
>
> > [!NOTE]
> > **Low Risk**
> > Localized to daemon launch/readiness and supervisor startup error logging; behavior change is clearer errors with a short race grace window.
> > 
> > **Overview**
> > Fixes **daemon cold start** so a crash no longer surfaces only after a 30s `Timed out waiting for daemon to start`.
> > 
> > The launcher records the daemon log size before spawn, listens for **spawn errors** and **early child exit**, and after a **2s grace** (so a losing concurrent launcher can still win the socket) fails fast with **exit code/signal** or spawn error plus a **log tail from bytes written this attempt only**—so stale log lines from prior runs are not blamed. Genuine timeouts append the same tail.
> > 
> > The **daemon supervisor** now logs startup failures (stack/message) to the rotating daemon log before cleanup/rethrow.
> > 
> > **Tests** cover log tail on exit, silent crash messaging, spawn ENOENT, startup race success, and stale-log isolation.
> > 
> > <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 68930748e744c0601f7fad87ac4d7e1037f67acf. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>
> 
> #### Changes since #608 opened
>
> - Modified `ensureDaemonRunning` function in `coding-agent` to track both spawn errors and exit events through a unified failure tracker, surfacing spawn errors immediately with the error message and daemon log tail instead of timing out, and including daemon log tail in exit-based failure messages [6893074]
> - Updated CHANGELOG.md for `coding-agent` package to document the daemon startup crash fix [6893074]
> 
>


snimu:
Both findings were valid, and CI exposed a third, worse problem with the same root cause: piping the detached daemon's stderr ties the daemon's lifetime to the short-lived CLI. Once the CLI exits and destroys its end of the pipe, the daemon's next `console.error` (the supervisor logs through it for its whole life) dies with EPIPE - that is what killed the daemon mid-test in `daemon-serialized-refine-process` and the 4685 daemon suites.

Fixed in d99cf6354 by dropping the pipe entirely: the child is spawned with `stdio: "ignore"` again, and on early exit or timeout the launcher reads the tail of the daemon's own rotating log, which the supervisor now writes startup failures to before rethrowing. That removes the exit-before-drain race (the log is complete by the time the process exits), the open-pipe leak on the timeout path, and the EPIPE coupling, in one move. Tests cover both the log-tail and the crashed-before-logging cases.

snimu:
Valid - the per-socket log path is stable across runs and appendRotatingLog preserves old content, so a daemon that crashed before logging would surface a previous run's tail as the current crash cause.

Fixed in 40150041f: the launcher snapshots the log size before spawning and reports only content written past that offset (falling back to the full file if a rotation shrank it below the offset, so fresh content is never hidden). The no-new-content message now reads "The daemon wrote nothing to its log" instead of claiming no log exists. Regression test: pre-existing stale log content is not attributed to a daemon that crashes before logging.

snimu:
Valid - the offset comes from statSync (bytes) but was applied with string slice (characters), so any multi-byte content shifted the window: stale lines could resurface past the guard, or fresh content could be hidden.

Fixed in 1e3545040: the file is read as a raw buffer, both the offset and the tail cap are applied with subarray in byte space, and only the result is decoded. The stale-log regression test now seeds multi-byte content (umlaut path) - verified it fails against the previous string-slice version and passes with the byte-space version.

snimu:
Valid - with fail-fast on child exit, a spawn that lost the socket to a concurrent launcher failed the ensure even though a healthy daemon was coming up (the winner may still be booting behind the socket lock, so probe-before-throw alone is not enough either).

Fixed in b0a82686b: a child exit no longer fails the launch immediately; it starts a 2s grace window during which the launcher keeps probing for a healthy daemon, and only then reports the exit with the log tail. Genuine crashes now fail in ~2s instead of instantly (still well under the 30s timeout). Regression test: child exits immediately, the winning daemon starts answering 300ms later, ensure resolves - verified the test fails against the previous fail-fast ordering.

packages/coding-agent/src/cli/daemon-launch.ts
```diff
@@ -365,30 +367,83 @@ async function ensureDaemonRunning(socketPath: string, spawnCwd?: string): Promi
...
 	const deadline = Date.now() + DAEMON_STARTUP_TIMEOUT_MS;
-	while (Date.now() < deadline) {
+	let exitDeadline: number | undefined;
+	while (Date.now() < Math.min(deadline, exitDeadline ?? Number.POSITIVE_INFINITY)) {
 		const started = await probeDaemonVersion(socketPath);
 		if (started.status === "current") {
 			return;
 		}
+		if (childExit) {
+			exitDeadline ??= Date.now() + DAEMON_STARTUP_EXIT_GRACE_MS;
```
alexzhang13, line 421: This is kind of sus to me, this seems a little arbitrary to me (the 2 second window)? What if recovery takes longer than 2s (or is this not a case to be concerned about).

packages/coding-agent/src/cli/daemon-launch.ts
```diff
@@ -365,30 +367,83 @@ async function ensureDaemonRunning(socketPath: string, spawnCwd?: string): Promi
...
-			// The daemon writes its own rotating log (see daemon-mode); nothing here
-			// needs its stdout/stderr, so leave them detached.
+			// A pipe would tie the daemon's stderr to this short-lived CLI
+			// (EPIPE once it exits); crash details come from the daemon log,
+			// which the supervisor writes to before rethrowing startup errors.
 			stdio: "ignore",
 		},
 	);
+	let childExit: { code: number | null; signal: NodeJS.Signals | null } | undefined;
+	child.once("exit", (code, signal) => {
```
alexzhang13, line 391: Should we also track spawn() failures or other errors (I'm not sure what else there is)?

alexzhang13 (changes requested):
Left a few comments (also with help from codex), main thing is I'm not entirely sure where some of these numbers come from, but they should be minor.

packages/coding-agent/src/cli/daemon-launch.ts
alexzhang13: Fixed

sethkarten (approved):

alexzhang13 (approved):

PR #657, opened 2026-08-05
fix(coding-agent): preserve untouched bytes when fuzzy edit matching is used

### LINKED ISSUE #654: fuzzy edit fallback silently rewrites the entire file and the reported diff hides it
Was digging around in the edit tool and noticed the fuzzy matching fallback doesn't just fix up the matched region, it writes the entire normalized file back to disk.

### Cause
When the exact match fails, `fuzzyFindText` in `packages/coding-agent/src/core/tools/edit-diff.ts` returns the whole fuzzy-normalized content as `contentForReplacement`, and `edit.ts` persists that. So a single edit where the model typed a smart quote instead of a straight one ends up stripping trailing whitespace on every line, ASCII-fying quotes and dashes everywhere, and applying NFKC folds like `½` → `1/2` and `ﬁ` → `fi` across the whole file, including inside string literals that had those characters on purpose.

### Repro
1. Make a file with `naïve — café²` on line 1 and `console.log('x')` further down.
2. Run an edit with `oldText: console.log(‘x’)` (curly quotes, so exact match fails) and `newText: console.log('y')`.
3. Line 1 is now `naïve - café2`.

What makes it worse: the diff shown for the edit is computed against the already-normalized content, so none of this shows up in it. The tool reports one changed line while the file changed everywhere. Multi-edit has the same problem: if any one edit needs fuzzy matching, all of them get applied to the normalized content.

There's a comment in `fuzzyFindText` saying the normalization is acceptable "since we're fixing minor formatting differences anyway", which seems fair for the matched region itself, but I don't think the file-wide rewrite (or the diff hiding it) is intended

Will poast a pr for this :3

0xqtpie:
## Issue

Fuzzy edit matching found its match in a fuzzy-normalized copy of the whole file and then wrote that normalized copy back to disk. One smart quote in `oldText` was enough to strip trailing whitespace on every line, ASCII-fy quotes and dashes, and apply NFKC folds (`½` → `1/2`, `ﬁ` → `fi`) file-wide. The rendered diff was computed from the normalized base, so none of it showed up. In multi-edit mode, one fuzzy edit dragged every other edit into normalized space too.

Fixes #654.

## Fix

Fuzzy matches are still located in normalized space, but the replacement is now spliced into the original content, leaving every byte outside the matched span untouched.

- Match boundaries are mapped back to original-content offsets and verified with prefix equality. Unmappable boundaries (e.g. a match starting inside an NFKC expansion) are rejected with the normal not-found error.
- Mapping doesn't trim, so matches after indentation or interior spaces work.
- A match ending at the last visible character of a line keeps that line's trailing whitespace; including the `\n` in `oldText` consumes it, same as before.
- Multi-edit matches everything against the original content; exact and fuzzy edits mix cleanly.
- `baseContent` from `applyEditsToNormalizedContent` is always the original LF-normalized content, so diff and preview match what's written.
- Mapping is fast-path first: a 1:1 column check when NFKC doesn't change lengths, then a single-pass grapheme-cluster accumulation (handles decomposed accents); both run at any line length. The exhaustive per-column scan is quadratic, so it only runs for lines up to 2048 chars; beyond that, if neither pass can verify a boundary, the fuzzy match returns the normal retryable not-found instead of stalling. Exact matching is unaffected.

## Validation

- `npm run check`
- `npx tsx ../../node_modules/vitest/dist/cli.js --run test/suite/regressions/654-fuzzy-edit-byte-preservation.test.ts`
  - 13 regressions: byte preservation (single + multi-edit), NFKC-expansion boundary rejection, trailing-whitespace semantics with/without newline and at newline-leading match starts, indented and mid-line starts, diff-matches-write, exact-match priority, and a multi-KB NFD line locking the single-pass path
- Existing edit-path suites all pass: `test/tools.test.ts` (original fuzzy tests unchanged), `test/edit-summary.test.ts`, `test/edit-tool-legacy-input.test.ts`, `test/edit-tool-no-full-redraw.test.ts`

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## [Unreleased]
 
+- Fixed fuzzy edit matching rewriting unrelated whitespace and Unicode across the whole file and omitting those changes from the reported diff ([#654](https://github.com/PrimeIntellect-ai/prime-agent/issues/654)).
+
 ## [0.7.0] - 2026-08-05
 
 ### Breaking Changes
--- a/packages/coding-agent/src/core/tools/edit-diff.ts
+++ b/packages/coding-agent/src/core/tools/edit-diff.ts
@@ -120,6 +120,7 @@ function mapNormalizedBoundaryToOriginal(
 	normalizedLines: string[],
 	originalLineStarts: number[],
 	normalizedOffset: number,
+	kind: "start" | "end",
 ): number | undefined {
 	const boundary = getTextBoundary(normalizedLines, normalizedOffset);
 	if (boundary === undefined) return undefined;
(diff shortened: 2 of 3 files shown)


0xqtpie:
Fixed bot review comments in 3b02ec62: start boundaries at a normalized line end now map past the trailing whitespace to the actual newline, end boundaries keep stopping before it, and both cases are covered by new regressions.

kevinjosethomas:

PR #664, opened 2026-08-06
fix(coding-agent): hide console windows for background child processes

podlinovvv:
## Problem

On Windows, `spawn`/`spawnSync`/`execFileSync` calls without `windowsHide: true` give every console child process its own visible console window. When Windows Terminal is the default terminal, each of those windows opens as a new WT window that flashes on screen and closes immediately.

Users see bursts of terminal windows flashing during normal agent use: at startup (node --version checks, tools --version checks), while typing in chat (powershell.exe process-identity checks in `session-lease.ts`), and in the footer (git branch), shell/tool exec, clipboard and daemon subprocess spawns.

## Fix

Add `windowsHide: true` to all background spawn/spawnSync/execFile/execFileSync/execSync call sites in `packages/coding-agent/src` (27 files, +39/-7). The flag is a no-op on POSIX, so this is safe cross-platform.

Interactive windows that the user is meant to see are intentionally untouched: `extension-editor.ts`, `login-dialog.ts`, update/relaunch paths that inherit the parent console (`stdio: "inherit"`).

## Reproduction

Windows 10/11 with Windows Terminal set as the default terminal application:
1. Run `prime-agent` from a terminal.
2. Start typing a message.
3. A burst of Windows Terminal windows opens and closes within a second.

Confirmed fixed after applying all call-site changes; reverting only the `session-lease.ts` change makes the windows come back, so the fix is needed across all background spawns, not just one.

---

> [!NOTE]
> **Low Risk**
> Windows-only UX change via a standard Node spawn option; no auth, data, or protocol changes, and interactive console paths were left untouched.
> 
> **Overview**
> Fixes **flashing Windows Terminal windows** during normal agent use by setting **`windowsHide: true`** on background `spawn` / `spawnSync` / `execFile` / `execSync` call sites across `packages/coding-agent/src` (daemon launch, workers, git/footer checks, bash/exec, kernel bootstrap, clipboard, package manager, etc.).
> 
> On Windows, console children without this flag get a visible console; with WT as the default terminal that shows up as brief new windows at startup, while typing (e.g. `powershell.exe` in session lease checks), and during tool/git subprocess work. The flag is a **no-op on POSIX**, so behavior elsewhere is unchanged.
> 
> **Intentionally not changed:** flows where the user is meant to see a console (e.g. extension editor, login dialog, inherited `stdio: "inherit"` update/relaunch).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b7e07abaae49258e5f9d288919e5cbb59f6bc84c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Hide console windows for all background child processes in the coding agent on Windows
> Adds `windowsHide: true` to every `spawn`, `spawnSync`, `execSync`, `execFile`, and `execFileSync` call across the coding agent package so that background processes do not create visible console windows on Windows. Affected sites include daemon launch, session workers, kernel management, package manager commands, git utilities, clipboard helpers, and the RPC client.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized b7e07ab.</sup>
> 
>

### DIFF
--- a/packages/coding-agent/src/cli/daemon-command.ts
+++ b/packages/coding-agent/src/cli/daemon-command.ts
@@ -692,6 +692,7 @@ async function runStart(parsed: ParsedDaemonClientCommand): Promise<void> {
 		...sessionArgs.daemonArgs.filter((arg) => arg !== "--background" && arg !== "-d"),
 	];
 	const child = spawn(process.execPath, daemonArgs, {
+		windowsHide: true,
 		cwd: sessionArgs.config?.cwd ?? process.cwd(),
 		detached: true,
 		env: process.env,
--- a/packages/coding-agent/src/cli/daemon-launch.ts
+++ b/packages/coding-agent/src/cli/daemon-launch.ts
@@ -372,6 +372,7 @@ async function ensureDaemonRunning(socketPath: string, spawnCwd?: string): Promi
 		process.execPath,
 		[...process.execArgv, entrypoint, "--mode", "daemon", "--daemon-socket", socketPath],
 		{
+			windowsHide: true,
 			cwd: spawnCwd ?? process.cwd(),
 			detached: true,
 			env,
--- a/packages/coding-agent/src/cli/daemon-update-restart.ts
+++ b/packages/coding-agent/src/cli/daemon-update-restart.ts
@@ -548,6 +548,7 @@ export async function launchDaemonUpdateRestartCoordinator(
 		...(originActiveSessionId ? [DAEMON_UPDATE_RESTART_ORIGIN_FLAG, originActiveSessionId] : []),
 	]);
 	const child = spawn(launch.command, launch.args, {
+		windowsHide: true,
 		cwd: options.cwd ?? process.cwd(),
 		detached: true,
 		env: coordinatorEnvironment(agentDir),
--- a/packages/coding-agent/src/cli/owned-session-worker.ts
+++ b/packages/coding-agent/src/cli/owned-session-worker.ts
@@ -344,6 +344,7 @@ export async function runOwnedSessionWorkerFrontend(
 			? ["inherit", "inherit", "inherit", "ipc"]
 			: [bridgeStdin ? "pipe" : "inherit", "pipe", "pipe", "ipc"];
 		const child = spawn(launch.command, launch.args, {
+			windowsHide: true,
 			cwd: process.cwd(),
 			detached: process.platform !== "win32",
 			env: {
--- a/packages/coding-agent/src/config.ts
+++ b/packages/coding-agent/src/config.ts
@@ -200,6 +200,8 @@ function readCommandOutput(
 	const result = spawnSync(command, args, {
 		encoding: "utf-8",
 		stdio: ["ignore", "pipe", "pipe"],
+		// Without windowsHide, console child processes get a visible window on Windows.
+		windowsHide: true,
 		shell: shouldUseWindowsShell(command),
 	});
 	if (result.status === 0) return result.stdout.trim() || undefined;
--- a/packages/coding-agent/src/core/autonomous.ts
+++ b/packages/coding-agent/src/core/autonomous.ts
@@ -492,6 +492,7 @@ function runChildProcess(
 	options.signal?.throwIfAborted();
 	return new Promise((resolve) => {
 		const child = spawn(command, args, {
+			windowsHide: true,
 			cwd: options.cwd,
 			detached: process.platform !== "win32",
 			shell: options.shell === true,
--- a/packages/coding-agent/src/core/exec.ts
+++ b/packages/coding-agent/src/core/exec.ts
@@ -59,6 +59,7 @@ export async function execCommand(
 ): Promise<ExecResult> {
 	return new Promise((resolve) => {
 		const proc = spawn(command, args, {
+			windowsHide: true,
 			cwd,
 			shell: false,
 			stdio: ["ignore", "pipe", "pipe"],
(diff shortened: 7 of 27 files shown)


noahsabaj:
Independent confirmation of this, plus a second symptom it fixes that may not be obvious.

I hit the same flood while working on native Windows support, and initially concluded the
opposite — that `windowsHide` was making things worse — because I measured it from a parent
that **has** a console. In that case the child inherits the console, no new one is created,
and adding the flag only appears to allocate an extra (hidden) console host. That measurement
is misleading. The case that matters is a **console-less** parent, which is what a daemon
always is, having been spawned `detached`.

Measured on Windows 11 (10.0.26200), Node v24.12.0, counting *visible* console windows via
`EnumWindows` + `IsWindowVisible` filtered to `ConsoleWindowClass` — not `conhost.exe`
process count, which is what misled me:

| parent | child spawn | visible console windows |
|---|---|---|
| `detached: true` (no console) | plain | 0 → **3** for 3 children |
| `detached: true` (no console) | `windowsHide: true` | 0 → **0** |
| has a console | plain | 0 → 0 (inherits) |
| has a console | `windowsHide: true` | 0 → 0 (hidden console host) |

And the windows **persist**: they outlive the child as orphaned console hosts, so they
accumulate rather than flashing. One `npm run test:process` run on my branch produced **61**
visible windows that had to be closed by hand.

**The second symptom.** Suppressing the console also fixed daemon worker startup for me.
`getProcessStartId` shells out to `powershell.exe`, and the daemon greeting called it. Paying
a console allocation inside the handshake stalled it past the supervisor's connect timeout, so
every `create` failed after 30 s with `Timed out connecting to daemon worker socket` — while
the worker log said it was listening. With the console suppressed, `--print` went from timing
out at 49 s to answering in 8 s, and the process-smoke suite went from 5 failures to 0. So this
PR is likely also a fix for daemon `create` timeouts on Windows, not only a cosmetic one.

**One suggestion.** Caching that start-id lookup is worth doing on top of this (#748 is heading
at the same cost from another direction): the value cannot change while a process is alive, so
a per-process cache removes the PowerShell startup from the handshake entirely rather than just
making it quieter.

**Agreement on the exclusions.** Keeping interactive launches visible is the right call — the
console editor, the CLI relaunch, and anything inheriting the frontend's console must not get
the flag, or the program the user just asked for disappears.

Repro, if useful:

```js
import { spawn, execFileSync } from "node:child_process";
// Count VISIBLE console windows, not conhost processes — that distinction is the whole trap.
const CHILD = (hide) => `
  const { execFileSync } = require("node:child_process");
  for (let i = 0; i < 3; i++) {
    try { execFileSync("powershell.exe", ["-NoLogo","-NoProfile","-NonInteractive","-Command","$null"],
      { stdio: ["ignore","pipe","ignore"]${hide ? ", windowsHide: true" : ""} }); } catch {}
  }
  setTimeout(() => {}, 500);
`;
for (const hide of [false, true]) {
  spawn(process.execPath, ["-e", CHILD(hide)], { detached: true, stdio: "ignore" });
  // ...sample visible ConsoleWindowClass windows while it runs
}
```

SandroHub013:
Tested `b7e07ab` cherry-picked onto current `main` (`a18809e0`) on Windows 11.

- A harness through the real `execCommand()` path captures `windowsHide=undefined` on `main` and `windowsHide=true` with this PR, so the core fix is effective.
- At least one user-triggered background spawn remains: `packages/tui/src/autocomplete.ts:165`. `walkDirectoryWithFd()` still spawns `fd` without `windowsHide`, and a harness through `CombinedAutocompleteProvider` still captures `windowsHide=undefined` after this PR. That path is relevant to the typing/autocomplete symptom in #735.

Please include that TUI spawn in the sweep and add an entry under the appropriate `[Unreleased]` changelog. A regression assertion at the `CombinedAutocompleteProvider` seam would lock down the remaining path.

`biome` and `tsgo --noEmit` passed locally. The root check then reached the existing Windows installer harness failure, which reproduces unchanged on `main` in this environment.

noahsabaj:
Confirming the autocomplete finding from a separate Windows 11 machine. `walkDirectoryWithFd()` in `packages/tui/src/autocomplete.ts` spawns `fd` once per keystroke, so completing a path opens one console window per character. That is the "terminal per keypress" half of #735, and it survives this PR because the spawn lives outside `packages/coding-agent/src`.

I checked the rest of the tree against this diff. One more site is missing: `interactive-mode.ts:8668`, the `spawnSync("gh", ["auth", "status"])` a few lines above the `gh gist create` you already fixed. Everything else I could find inside `packages/coding-agent/src` is covered here.

One caveat on the regression assertion, since the TUI seam behaves differently from the ones under vitest. `packages/tui` runs `node --test` with tsx and no module mocker, and `autocomplete.ts` binds `spawn` through an ESM named import. An ESM binding to a builtin does not observe writes to the CommonJS `child_process` exports, so a harness that patches `cp.spawn` reads `windowsHide=undefined` there whether or not the fix is applied. I asserted on the source of the spawn call instead; removing the flag fails the test. Under vitest `vi.mock` intercepts the module properly and a real seam assertion works, so the `packages/coding-agent` paths can be tested the usual way.

I have the two sites and that test as one commit on my fork if it saves the retyping: noahsabaj/prime-agent@f5092e3a. Reached the same conclusion independently — glad to see it land here.

kevinjosethomas:

PR #675, opened 2026-08-06
fix(coding-agent): preserve live workers during recovery

twaldin:
## Summary

- Keep identity-verified live workers alive when adoption or recovery probes time out.
- Continue recovery in the background, including when process identity observation is temporarily unavailable.
- Preserve destructive recovery for confirmed process death or PID reuse, without replaying uncertain operations.

## Why

A restarted supervisor can adopt a worker that is still alive but too busy to answer `hello`, `worker_subscribe`, or `list` within the control-plane deadlines. After three retries, recovery previously treated the matching live process as safe to `SIGKILL`, relaunched it, and marked every hosted session's in-flight operation interrupted. In one observed incident, this interrupted 18 concurrent RLM sessions at the same timestamp.

The supervisor cannot distinguish a load-slow worker from a wedged worker from probe timeouts alone. Keeping the verified process alive avoids irreversible work loss and retains the existing no-replay rule for uncertain side effects. Recovery continues on the existing deferred-retry path and will relaunch if the process later dies or its PID identity changes.

## Verification

- `npx tsx ../../node_modules/vitest/dist/cli.js --run test/daemon-supervisor-monitor.test.ts` (45 tests)
- `npm run check`
- Full package build

Adversarial review identified that the first bounded adoption probe can still delay supervisor readiness when `worker_subscribe` is load-slow. This PR moves recovery to the background after that first probe fails. Removing the first synchronous probe would change the existing startup contract by making healthy adopted sessions temporarily absent from the first `list` response, so that broader availability change is left out of this safety fix.

---
Generated by AI using OpenAI Codex `gpt-5.6-sol`.

---

> [!NOTE]
> **Medium Risk**
> Touches core daemon supervisor worker recovery and process-identity safety; wrong logic could leave wedged workers unrecovered or still kill live sessions, but changes narrow destructive recovery and add tests.
> 
> **Overview**
> Fixes daemon supervisor recovery so **identity-verified live session workers are not SIGKILL’d and relaunched** when adoption or reconnect probes time out on a busy but healthy process.
> 
> **Adoption:** If the initial adopt probe fails but the PID is still alive and `processStartId` still matches (or identity is temporarily unreadable), the worker is marked `recovering` and **`recoverWorker` runs in the background** instead of immediately entering destructive recovery.
> 
> **Recovery loop:** On connect/subscribe failures against a verified live process, recovery **keeps probing** rather than treating the last retry as grounds to kill and relaunch. **`recoverUncertainWorkerOperations` is no longer invoked with kill permission** for that path (`false` instead of conditional `SIGKILL`). If probes still fail after the bounded delays, recovery **defers** via `deferWorkerRecovery` instead of failing the worker; relaunch remains for confirmed death or PID reuse.
> 
> Changelog entry and **daemon-supervisor-monitor** tests cover failed adoption on live workers and repeated probe timeouts / temporary missing process identity.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 03721182018ea62f71298ff5b7525b472eb55c44. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Fix `DaemonSupervisor` to preserve verified live workers during recovery
> - `adoptOrRecoverWorker` now checks if a worker process is alive and its `processStartId` matches before recovery; if it does, it marks the descriptor as `"recovering"` and schedules `recoverWorker` instead of immediately replacing the worker.
> - `handleWorkerClose` tracks a `keepRetryingLiveWorker` flag during the retry loop; if the worker is still alive (identity confirmed or temporarily unavailable), it defers recovery via `deferWorkerRecovery` rather than launching a replacement.
> - `recoverUncertainWorkerOperations` is now always called with `false` (never kill the live process) instead of a computed value.
> - Behavioral Change: live workers that fail adoption or exhaust retry attempts are no longer replaced — they remain in `"recovering"` state until a subsequent probe succeeds.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 0372118.</sup>
> 
>

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## [Unreleased]
 
+- Fixed supervisor recovery replacing live, load-slow session workers and interrupting their in-flight work.
+
 ## [0.7.0] - 2026-08-05
 
 ### Breaking Changes
--- a/packages/coding-agent/src/modes/daemon/daemon-supervisor.ts
+++ b/packages/coding-agent/src/modes/daemon/daemon-supervisor.ts
@@ -2409,6 +2409,21 @@ export class DaemonSupervisor {
 				return;
 			}
 			this.log(`Could not adopt worker ${worker.descriptor.workerId}: ${String(error)}`);
+			const processAlive = isProcessAlive(worker.descriptor.pid);
+			const observedProcessStartId = processAlive ? getProcessStartId(worker.descriptor.pid) : undefined;
+			if (
+				worker.descriptor.processStartId !== undefined &&
+				processAlive &&
+				(observedProcessStartId === undefined || observedProcessStartId === worker.descriptor.processStartId)
+			) {
+				worker.descriptor.lifecycle = "recovering";
+				worker.descriptor.lastError = error instanceof Error ? error.message : String(error);
+				this.persistWorker(worker);
+				void this.recoverWorker(worker).catch((recoveryError) =>
+					this.log(`Could not recover worker ${worker.descriptor.workerId}: ${String(recoveryError)}`),
+				);
+				return;
+			}
 			await this.recoverWorker(worker);
 		}
 	}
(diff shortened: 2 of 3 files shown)


kevinjosethomas:

Issue #639, opened 2026-08-05
RLM excludes OpenAI Codex models because discovery sends Prime Agent version as client_version


maks-ivanov:
## Reproduction

With authenticated OpenAI Codex credentials:

```bash
prime-agent --provider openai-codex --model gpt-5.6-sol
```

Ask the agent to spawn an RLM child with `openai-codex/gpt-5.6-luna`.

Actual:

```text
Requested subagent model "openai-codex/gpt-5.6-luna" is unavailable, unauthenticated, or expired
```

A child without an explicit model succeeds by inheriting `openai-codex/gpt-5.6-sol`.

## Cause

In `openAICodexModelsUrl()`, Prime Agent sends its own `VERSION` as the Codex `client_version`. On Prime Agent 0.7.0:

```text
/codex/models?client_version=0.7.0  -> HTTP 200, {"models":[]}
```

The empty catalog causes `getExecutableModels()` to exclude all `openai-codex` models from `rlm.find_models()` and explicit child selection.

The same authenticated request with the installed Codex CLI version (`0.144.6`) returns nine models, including `gpt-5.6-luna`, `gpt-5.6-sol`, and `gpt-5.6-terra`.

## Validation

Tested in separate sessions:

- Unpatched 0.7.0: explicit Luna selection is rejected.
- Temporary patch using a Codex-compatible client version: the child is admitted as `openai-codex/gpt-5.6-luna`.

Prime Agent should use a Codex-compatible discovery version rather than the Prime Agent package version for this request.

etafund:
Independent reproduction on a second machine, plus a detail that I think makes the
proposed fix under-specified: **the catalog gate is graduated, not binary.**

### Environment
prime-agent 0.7.0 (global npm install), `openai-codex` OAuth authenticated and
unexpired, paid ChatGPT tier. Codex CLI on the same box: 0.146.0.

### Measured: `client_version` -> returned slugs

Same account, same valid token, same endpoint, all HTTP 200. Only `client_version`
was varied:

| `client_version` | n | slugs |
|---|---|---|
| **0.7.0** (what prime-agent sends) | **0** | *(empty)* |
| 0.10.0 – 0.97.0 | 0 | *(empty)* |
| 0.98.0 | 3 | `gpt-5.4`, `gpt-5.4-mini`, `codex-auto-review` |
| 0.105.0 | 4 | + `gpt-5.3-codex-spark` |
[...]

PR #717, opened 2026-08-06
fix: keep codex models executable when discovery returns an empty catalog

### LINKED ISSUE #696: openai-codex: empty-but-successful model discovery silently disables the provider and is reported as an auth failure
## Summary

When OpenAI Codex model discovery succeeds but returns an **empty** catalog
(`HTTP 200 {"models":[]}`), `getExecutableModels()` silently removes every
`openai-codex` model from the executable set. The resulting user-facing error
blames authentication, even though the credentials are valid and the model works.

This is separable from #639. #639 covers *why* the catalog comes back empty (the
`client_version` sent is the prime-agent package version). This issue is about the
**handling** of an empty/partial catalog, which stays latent after that fix: the
same silent-disable path triggers for any account whose entitlement gate returns a
subset or nothing.

## Actual behavior

```
Requested subagent model "openai-codex/gpt-5.6-sol" is unavailable, unauthenticated, or expired
```

Meanwhile, with the exact same credentials:

- `POST /backend-api/codex/responses` with `{"model": "gpt-5.6-sol", ...}` returns
  HTTP 200 and a normal completed response.
- The model is selectable as the interactive/lead model (the picker uses the
  unfiltered `getAvailable()` path).
- An RLM child spawned with **no** explicit model inherits the codex model and runs
  fine.

So the model is simultaneously "unavailable" and demonstrably working, depending on
which code path you go through.

## Expected behavior

1. An empty-but-successful discovery response should not be treated as equivalent to
   a hard discovery failure. A healthy `200` with zero models is far more likely a
   client-version or entitlement gate than a signal that an authenticated provider
   should be disabled.
2. Whatever the policy, the failure should be **visible** — a warning naming the
   provider and the empty catalog, rather than models quietly vanishing from
   `rlm.find_models()`.
3. The error should name the layer that actually rejected the model. "not present in
   the provider's discovery catalog" points at the real cause; "unavailable,
   unauthenticated, or expired" points at auth and sends you to re-login, re-check
   token expiry, and inspect credentials — none of which are the problem.

## Root cause

`dist/core/model-registry.js`, `getExecutableModels()`:

```js
const modelIds = readOpenAICodexModelIds(await response.json());
this.openAICodexModelsCache = { authFingerprint, modelIds, refreshedAt: Date.now() };
return availableModels.filter((model) => model.provider !== "openai-codex" || modelIds.has(model.id));
```

`readOpenAICodexModelIds({"models": []})` returns an empty `Set` — it only throws on a
*malformed* payload, not an empty one. The filter above then drops all
`openai-codex` models. That is the identical outcome to the catch branch:

```js
return availableModels.filter((model) => model.provider !== "openai-codex");
```

`dist/core/agent-session.js` then surfaces it as an auth-flavored error:

```js
const model = (await this._authenticatedRlmModels()).find(...);
if (!model) {
    throw new Error(`Requested subagent model "${reference}" is unavailable, unauthenticated, or expired`);
}
```

Note the auth preflight immediately below this throw (`getApiKeyAndHeaders`) has its
own distinct message, so auth failures are *already* separately reported — which
means the "unauthenticated, or expired" wording on this branch is describing a case
it doesn't actually cover.

## Reproduction

Any authenticated `openai-codex` account whose discovery call returns an empty
catalog reproduces it; on 0.7.0 that is every account, per #639.

```bash
# 1. authenticate the openai-codex provider
# 2. observe discovery returns an empty catalog:
curl -s -H "Authorization: Bearer <codex access token>" \
     -H "chatgpt-account-id: <account id>" \
     -H "originator: pi" \
     "https://chatgpt.com/backend-api/codex/models?client_version=0.7.0"
# -> {"models":[]}

# 3. in a session, ask for an RLM child on an openai-codex model:
#    rlm(..., model="openai-codex/gpt-5.6-sol")
# -> "unavailable, unauthenticated, or expired"
```

## Suggested fix

- Treat empty-but-successful discovery distinctly from discovery failure; at minimum
  do not silently disable an authenticated provider on it.
- Emit a warning that names the provider, the empty catalog, and the
  `client_version` that was sent.
- Split the `_resolveRlmSubagentModel` error into causes that are actually
  distinguishable at that point: not in catalog / not authenticated / expired.

## Environment

- prime-agent 0.7.0, global npm install
- `openai-codex` provider, OAuth, unexpired, paid ChatGPT tier
- Other providers (anthropic, kimi-coding) unaffected
- The running daemon executes the bundled build (`dist/bundle/cli.js`), where
  `openAICodexModelsUrl` and `ModelRegistry` live in the same bundle chunk — worth
  confirming any fix is verified against the bundle, not only `dist/core/`.

ayushnangia:
Fixes #696.

## Problem

`getExecutableModels()` treats a successful discovery response with zero models the same as a real catalog: every `openai-codex` model is filtered out and the empty set is cached for 5 minutes. As #696 documents, the same credentials serve completions over `/codex/responses`, the picker still offers the model via `getAvailable()`, and the resulting subagent error blames authentication.

An empty 200 is an entitlement or client-version gate (#639), not proof that an authenticated provider has no models — and #639's fix leaves this handling latent for any account whose gate returns a subset or nothing.

## Change

`packages/coding-agent/src/core/model-registry.ts`: empty catalogs fail open — codex models stay executable and per-request errors surface the real cause. Applies to the fresh result, the fresh-cache read, and the stale-cache fallback via one `filterByCatalog` helper. Non-empty catalogs filter exactly as before; the empty result is still cached, so there is no refetch-per-call.

## Tests

`test/suite/regressions/696-empty-codex-catalog.test.ts`:

- empty catalog keeps codex models executable (fails on main)
- cached empty catalog stays fail-open, single fetch (fails on main)
- subset catalog still filters to the returned slugs (control, passes on main)

`npm run check` clean; test run from the package root per AGENTS.md.

---

> [!NOTE]
> **Low Risk**
> Narrow change to Codex executable-model filtering with fail-open behavior on empty discovery; subset filtering and auth failure paths are unchanged.
> 
> **Overview**
> Fixes **#696** by changing how `getExecutableModels()` applies OpenAI Codex model discovery.
> 
> Previously, a successful discovery response with **zero** models was treated like a real catalog: every `openai-codex` entry was removed from the executable list and that empty set was cached for five minutes, while completions could still work and the UI could still show those models via `getAvailable()`.
> 
> **`filterByCatalog`** now **fails open** when the discovered ID set is empty—Codex models stay executable and any entitlement or version issue surfaces on the actual request. When discovery returns one or more slugs, filtering is unchanged. The helper is used for fresh discovery, cache hits, and stale-cache fallback.
> 
> Regression tests cover empty catalog, cached empty catalog (single fetch), and subset filtering.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e1a557912a41926ce384560f49c4b7b37d486d3a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Fix `ModelRegistry.getExecutableModels` to keep codex models available when discovery returns an empty catalog
> When the OpenAI Codex catalog discovery returns an empty set, `getExecutableModels` previously filtered out all codex models. The fix introduces a `filterByCatalog` helper in [model-registry.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/717/files#diff-2fd8c4a2672b4ce27e73d494479df612c8b30ebac3dda5a160e2d861c612f453) that skips catalog filtering when the discovered ID set is empty, treating an empty catalog as a failed entitlement check and failing open. Non-empty catalogs still filter as before. A regression test suite in [696-empty-codex-catalog.test.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/717/files#diff-97a3a2438d1f3677b3ca811b481f8b168b29bc44c4406deac880d1399b07af9e) covers the empty catalog, cached empty catalog, and non-empty catalog cases.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized e1a5579.</sup>
> 
>

### DIFF
--- a/packages/coding-agent/src/core/model-registry.ts
+++ b/packages/coding-agent/src/core/model-registry.ts
@@ -978,9 +978,19 @@ export class ModelRegistry {
 			return availableModels.filter((model) => model.provider !== "openai-codex");
 		}
 		const authFingerprint = createHash("sha256").update(auth.apiKey).digest("hex");
+		// An empty catalog from a successful discovery is an entitlement or
+		// client-version gate (#639), not proof that an authenticated provider
+		// has no models — filtering on it silently disabled every codex model
+		// while the models demonstrably worked, and the resulting error blamed
+		// authentication (#696). Fail open on empty and let per-request errors
+		// surface the real cause; non-empty catalogs filter as before.
+		const filterByCatalog = (ids: Set<string>): Model<Api>[] =>
+			ids.size === 0
+				? availableModels
+				: availableModels.filter((model) => model.provider !== "openai-codex" || ids.has(model.id));
 		const cached = this.openAICodexModelsCache;
 		if (cached?.authFingerprint === authFingerprint && Date.now() - cached.refreshedAt < 300_000) {
-			return availableModels.filter((model) => model.provider !== "openai-codex" || cached.modelIds.has(model.id));
+			return filterByCatalog(cached.modelIds);
 		}
 
 		const accountId = readOpenAICodexAccountId(auth.apiKey);
@@ -1002,12 +1012,10 @@ export class ModelRegistry {
 			}
 			const modelIds = readOpenAICodexModelIds(await response.json());
 			this.openAICodexModelsCache = { authFingerprint, modelIds, refreshedAt: Date.now() };
-			return availableModels.filter((model) => model.provider !== "openai-codex" || modelIds.has(model.id));
+			return filterByCatalog(modelIds);
 		} catch {
 			if (cached?.authFingerprint === authFingerprint && Date.now() - cached.refreshedAt < 300_000) {
-				return availableModels.filter(
-					(model) => model.provider !== "openai-codex" || cached.modelIds.has(model.id),
-				);
+				return filterByCatalog(cached.modelIds);
 			}
 			return availableModels.filter((model) => model.provider !== "openai-codex");
 		}
--- a/packages/coding-agent/test/suite/regressions/696-empty-codex-catalog.test.ts
+++ b/packages/coding-agent/test/suite/regressions/696-empty-codex-catalog.test.ts
@@ -1,0 +1,86 @@
+import { existsSync, mkdirSync, rmSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import { AuthStorage } from "../../../src/core/auth-storage.js";
+import { clearApiKeyCache, ModelRegistry } from "../../../src/core/model-registry.js";
+
+function fakeCodexJwt(accountId: string): string {
+	const payload = Buffer.from(
+		JSON.stringify({ "https://api.openai.com/auth": { chatgpt_account_id: accountId } }),
+		"utf8",
+	).toString("base64url");
+	return `${Buffer.from("{}").toString("base64url")}.${payload}.sig`;
+}
+
+function catalogResponse(slugs: string[]): Response {
+	return new Response(JSON.stringify({ models: slugs.map((slug) => ({ slug })) }), {
+		status: 200,
+		headers: { "content-type": "application/json" },
+	});
+}
+
+describe("issue #696 empty codex catalog must not disable the provider", () => {
+	let tempDir: string;
+	let authStorage: AuthStorage;
+
+	beforeEach(() => {
+		tempDir = join(tmpdir(), `pi-test-696-${Date.now()}-${Math.random().toString(36).slice(2)}`);
+		mkdirSync(tempDir, { recursive: true });
+		authStorage = AuthStorage.create(join(tempDir, "auth.json"));
+	});
+
+	afterEach(() => {
+		if (tempDir && existsSync(tempDir)) {
+			rmSync(tempDir, { recursive: true });
+		}
+		clearApiKeyCache();
+		vi.unstubAllGlobals();
+	});
+
+	it("keeps codex models executable when discovery succeeds with zero models", async () => {
+		const registry = ModelRegistry.inMemory(authStorage);
+		authStorage.setRuntimeApiKey("openai-codex", fakeCodexJwt("acct_696"));
+		const available = registry.getAvailable().filter((model) => model.provider === "openai-codex");
+		expect(available.length).toBeGreaterThan(0);
+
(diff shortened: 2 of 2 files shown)


fredluz:
Tested this locally against the current branch, including a real Codex/Luna subagent dispatch. The fix works as intended, and I did not observe any other breaking changes in the affected flow.

ayushnangia:
Thanks for the independent verification — a real Codex/Luna subagent dispatch is exactly the flow this guards. For reference, #731 is the companion fix (correct `client_version` so discovery returns the real catalog); this PR covers the fail-open when discovery legitimately returns empty.

amphetamarina:
Verified locally. I applied this PR's empty-catalog fail-open fix to my installed Prime Agent runtime and restarted the daemon.

Environment:
- Prime Agent: `0.7.0`
- OS: Ubuntu 24.04.4 LTS
- Kernel: `7.0.0-28-generic`
- Architecture: `x86_64`

Before the fix:
- `await rlm.find_models("luna")` returned an empty list.
- Spawning `openai-codex/gpt-5.6-luna` failed as unavailable, unauthenticated, or expired.

After the fix and restart:
- Model discovery returned `openai-codex/gpt-5.6-luna`.
- A Luna RLM subagent was admitted, completed, and sent its expected reply to the parent.
- A direct registry check kept 13 OpenAI Codex models executable, including Luna and Sol.

This verifies the fix against the bundled runtime path used by the daemon, not only the unbundled module.

ayushnangia:
Thanks for the before/after verification — that's the second independent confirmation on this PR (with the Codex/Luna dispatch test above), both on live runtimes. For anyone landing here from the RLM-discovery symptom: #731 is the companion that fixes discovery itself (correct `client_version`), and the pair together resolves #740/#639/#702.

sethkarten:
Closed as superseded by #1153, an independent implementation written from `upstream/main` without importing outside commits, source code, or test files. #1153 addresses the same empty-catalog behavior with its own implementation and regression coverage.

ayushnangia:
Completely reasonable — a clean review surface built from `upstream/main` is the right call for the fork's provenance. Glad the diagnosis and regression shape were useful as a report; the fail-open empty-catalog behavior was the part that mattered. I'll keep future findings coming as detailed reports with repros.

PR #721, opened 2026-08-06
fix: wait for scheduled post-compaction continuation before reporting idle

### LINKED ISSUE #674: Headless execution terminates prematurely on compaction
On first attempt to integrate prime-agent into my benchmark, I found it terminated prematurely. The termination was triggered by auto compaction. 

## Simple Repro

It reproduces consistently. I was able to trigger the termination with just this inside my project (modify the prompt as appropriate):

`.prime/agent/settings.json`
```json
{ "compaction": { "enabled": true, "reserveTokens": 265000, "keepRecentTokens": 4000 } }
```

```bash
prime-agent --print --mode json --model gpt-5.6-luna --provider openai-codex "lets build a new level around the theme of fire. read the existing level code"
```

In this simple case, the termination occurs even if the auto compaction was skipped. These are the last events that appear:

```jsonl
{"type":"compaction_start","reason":"threshold"}
{"type":"message_start","message":{"role":"custom","customType":"compaction_outcome","content":"Auto-compaction skipped: Session is too short to compact — try again once it grows","display":true,"details":{"reason":"threshold","outcome":"skipped"},"timestamp":1785980979988}}
{"type":"message_end","message":{"role":"custom","customType":"compaction_outcome","content":"Auto-compaction skipped: Session is too short to compact — try again once it grows","display":true,"details":{"reason":"threshold","outcome":"skipped"},"timestamp":1785980979988}}
{"type":"compaction_end","reason":"threshold","aborted":false,"willRetry":false,"errorMessage":"Auto-compaction skipped: Session is too short to compact — try again once it grows","errorSeverity":"warning"}
```

**The rest of this issue was written by AI.**

---

# Clanker Report 🤖: upstream pi fixed this one; prime-agent has diverged from that code

Follow-up. I went looking upstream and found the same bug already reported and fixed on pi, but prime-agent can't just pick it up.

## pi had this and fixed it

[earendil-works/pi#7020](https://github.com/earendil-works/pi/issues/7020) — "Sometimes Pi doesn't continue after compaction":

> When the LLM then returns `stopReason: "length"` with a small amount of output, Pi compacts successfully but sets `willRetry: false` and returns control.

Fixed by [pi#7540](https://github.com/earendil-works/pi/pull/7540), merged 2026-08-03. The change is in `_checkCompaction`:

```js
const recoverableLength = sameModel && isRecoverableLength(assistantMessage, this.model?.maxTokens ?? 0);
if (sameModel && (isContextOverflow(assistantMessage, contextWindow) || recoverableLength)) {
```

A length stop that ended below the model's own `maxTokens` now counts as recoverable, so the session compacts and retries instead of handing control back. Related open meta-issue: [pi#5886](https://github.com/earendil-works/pi/issues/5886), on `agent_end` not implying the session is settled.

## prime-agent doesn't have it, and the patch won't drop in

`isRecoverableLength` appears zero times in prime-agent 0.7.0's bundled core, and `dist/core/agent-session.js` has no `"length"` stop handling at all.

The file is also 9,151 lines against pi 0.80.10's 2,659, and the divergence is concentrated in exactly this subsystem. Present in prime-agent, absent from pi:

```
_schedulePostCompactionContinue    _continueAfterThresholdCompaction
hasPendingSessionWork              _scheduleAutoRefineAfterCompaction
unfinishedActionCount
```

So prime-agent already has post-compaction continuation machinery that pi never had — this is a re-implementation to fix, not a merge.

## Which lines up with the aborted turn I reported

The continuation is deferred rather than run inline:

```js
_schedulePostCompactionContinue() {
  if (this._postCompactionContinuationScheduled) return;
  this._postCompactionContinuationScheduled = true;
  this._scheduledPostCompactionContinuationMessages = [...this._postCompactionContinuationMessages];
  this._postCompactionContinuationTimer = setTimeout(() => {
    this._postCompactionContinuationTimer = undefined;
    void this._runScheduledPostCompactionContinue();
  }, 100);
}
```

A 100ms timer is exactly what print-mode teardown would race, which fits the zero-token `stopReason: "aborted"` turn 0.25s after compaction in my original report. So the continuation is scheduled and then killed, rather than never scheduled.

## What I'd ask for

1. Handle a `length` stop the way pi#7540 does — resume rather than treat it as the end of the run.
2. Don't let print mode tear down a scheduled continuation; the 100ms timer needs to be work the completion wait knows about.
3. Failing both, make it observable. Right now a cut-short run and a finished run are identical from outside: exit 0, empty stderr, no marker in the final message. That is what makes this dangerous unattended — I sealed and shipped a truncated run without knowing, and only caught it because an unrelated check tripped.

## Environment

prime-agent 0.7.0, node v22.19.0, Ubuntu 22.04 on WSL2. Comparisons above are against pi 0.80.10 (`@earendil-works/pi-coding-agent`).

ayushnangia:
Fixes #674.

## Problem

The post-compaction continuation is deferred through a 100ms `setTimeout`. `waitForIdle()` checks queued actions, the input pump, streaming state, and unfinished actions — none of which see a pending continuation timer. `waitForHeadlessCompletion()` resolves through that gap, print mode tears the session down, and the scheduled continuation is killed. From outside the run looks finished: exit 0, empty stderr, no marker — exactly the truncation reported in #674 (compaction_end at T, zero-token `aborted` turn 0.25s later).

## Change

`packages/coding-agent/src/core/agent-session.ts`:

- `waitForIdle()` waits while `_postCompactionContinuationScheduled` is set. If the continuation starts a turn, the next loop iteration observes the streaming state as before.
- Waiters are woken on every continuation exit path: the scheduled run (wrapped in try/finally), reschedule (waiter re-arms against the new timer), cancel, and dispose — so the wait cannot outlive the continuation.

One existing fake-timers test awaited `waitForIdle()` with a continuation scheduled; it now advances the 100ms timer first, since the wait is the point of this change.

## Tests

`test/suite/regressions/674-headless-compaction-continuation.test.ts`:

- `waitForIdle` resolves only after the scheduled continuation ran (fails on main: resolves with the timer still pending)
- cancellation releases the wait (control against deadlock)

`agent-session-compaction.test.ts` suite: 34/34. `npm run check` clean.

---

> [!NOTE]
> **Medium Risk**
> Changes core session idle/completion timing (affects headless exit) and model availability logic for OpenAI Codex; behavior is intentional but touches run lifecycle and provider selection.
> 
> **Overview**
> Fixes **headless/print runs ending early** (#674) by treating the deferred post-compaction continuation (100ms timer) as session work in `waitForIdle()`. When `_postCompactionContinuationScheduled` is set, the idle loop waits on waiter callbacks that are released when the continuation runs (try/finally), reschedules, or is cancelled—so completion no longer tears the session down before `agent.continue` fires.
> 
> Separately fixes **OpenAI Codex models disappearing** (#696): `getExecutableModels()` used to filter Codex entries against discovery catalog IDs, so a **successful but empty** catalog removed every Codex model and surfaced misleading auth errors. Empty catalogs now **fail open** (keep configured Codex models); non-empty catalogs still filter to the returned subset.
> 
> Regression tests cover continuation wait/cancel behavior and empty vs subset Codex catalogs; an existing compaction test advances fake timers before `waitForIdle` under the new semantics.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 414eb580f4671e146fea9fd61effb89138413134. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Fix `waitForIdle` to wait for scheduled post-compaction continuation before reporting idle
> - `AgentSession.waitForIdle` previously returned before a scheduled post-compaction continuation timer had fired, causing premature session teardown. It now blocks until the continuation either runs or is cancelled.
> - A new `_postCompactionContinuationWaiters` set tracks resolvers; `_cancelPostCompactionContinue` and `_runScheduledPostCompactionContinue` both call `_notifyPostCompactionContinuationWaiters` to unblock waiters on every exit path.
> - Fixes a separate bug where an empty OpenAI Codex model catalog returned by a successful discovery request would disable all codex models; empty catalogs are now treated as unfiltered.
> - Regression tests added for both fixes in [674-headless-compaction-continuation.test.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/721/files#diff-38cb2313e648eeecc82412001b5cc56a51efe16c52137bf4be892e9a607b78b5) and [696-empty-codex-catalog.test.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/721/files#diff-97a3a2438d1f3677b3ca811b481f8b168b29bc44c4406deac880d1399b07af9e).
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 414eb58.</sup>
> 
>

### DIFF
--- a/packages/coding-agent/src/core/agent-session.ts
+++ b/packages/coding-agent/src/core/agent-session.ts
@@ -1253,6 +1253,7 @@ export class AgentSession {
 	private _turnIntervalAutoRefinePending = false;
 	private _postCompactionContinuationScheduled = false;
 	private _postCompactionContinuationTimer: ReturnType<typeof setTimeout> | undefined;
+	private _postCompactionContinuationWaiters = new Set<() => void>();
 	private _postCompactionContinuationMessages: AgentMessage[] = [];
 	private _scheduledPostCompactionContinuationMessages: AgentMessage[] = [];
 	private _queuedAutonomousThresholdContinuations = new WeakMap<AssistantMessage, AgentMessage>();
@@ -6426,6 +6427,16 @@ export class AgentSession {
 			await this.agent.waitForIdle();
 			const agentEventQueue = this._agentEventQueue;
 			await agentEventQueue;
+			// A scheduled post-compaction continuation is pending session work
+			// that lives only in a timer — none of the checks below see it, so
+			// headless/print completion would tear the session down before the
+			// continuation fires and silently truncate the run (#674). Wait for
+			// the continuation to run or be cancelled; if it starts a turn, the
+			// next loop iteration observes the streaming state as usual.
+			if (this._postCompactionContinuationScheduled) {
+				await new Promise<void>((resolve) => this._postCompactionContinuationWaiters.add(resolve));
+				continue;
+			}
 			if (
 				pump === this._sessionInputPump &&
 				agentEventQueue === this._agentEventQueue &&
@@ -7180,6 +7191,15 @@ export class AgentSession {
 		}
 		this._postCompactionContinuationScheduled = false;
 		this._scheduledPostCompactionContinuationMessages = [];
+		this._notifyPostCompactionContinuationWaiters();
+	}
+
+	private _notifyPostCompactionContinuationWaiters(): void {
+		const waiters = [...this._postCompactionContinuationWaiters];
+		this._postCompactionContinuationWaiters.clear();
+		for (const waiter of waiters) {
+			waiter();
+		}
 	}
 
 	private _discardPendingAutoRefine(options: { cancelPostCompactionContinue?: boolean } = {}): void {
@@ -7291,6 +7311,18 @@ export class AgentSession {
 	}
 
 	private async _runScheduledPostCompactionContinue(): Promise<void> {
+		try {
+			await this._runScheduledPostCompactionContinueInner();
+		} finally {
+			// Wake waitForIdle waiters on every exit path. If the run
+			// rescheduled itself, the flag is set again and the waiter loop
+			// re-arms; if it started a turn, the streaming state keeps
+			// waitForIdle looping as usual.
+			this._notifyPostCompactionContinuationWaiters();
+		}
+	}
+
+	private async _runScheduledPostCompactionContinueInner(): Promise<void> {
 		await this._waitForRefineIdle();
 		if (!this._postCompactionContinuationScheduled) {
 			return;
--- a/packages/coding-agent/src/core/model-registry.ts
+++ b/packages/coding-agent/src/core/model-registry.ts
@@ -978,9 +978,19 @@ export class ModelRegistry {
 			return availableModels.filter((model) => model.provider !== "openai-codex");
 		}
 		const authFingerprint = createHash("sha256").update(auth.apiKey).digest("hex");
+		// An empty catalog from a successful discovery is an entitlement or
+		// client-version gate (#639), not proof that an authenticated provider
+		// has no models — filtering on it silently disabled every codex model
+		// while the models demonstrably worked, and the resulting error blamed
+		// authentication (#696). Fail open on empty and let per-request errors
+		// surface the real cause; non-empty catalogs filter as before.
+		const filterByCatalog = (ids: Set<string>): Model<Api>[] =>
+			ids.size === 0
+				? availableModels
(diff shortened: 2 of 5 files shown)


kevinjosethomas:

PR #726, opened 2026-08-06
fix(coding-agent): render belowEditor widgets below the prompt in fullscreen

### LINKED ISSUE #697: ctx.ui.setWidget placement: 'belowEditor' renders above the prompt input, not below
### Summary
An extension using `ctx.ui.setWidget(key, string[], { placement: "belowEditor" })` renders the widget above the prompt input box instead of below it. The placement option is being ignored (or inverted) in the RPC extension UI bridge.

### Environment
- prime-agent version: 0.7.0 (npm/Node CLI install)
- Depends on: @earendil-works/pi-agent-core, pi-ai, pi-tui (private release tarballs bundled by prime-agent)

### Repro
Minimal extension:

```ts
export default function (pi) {
  pi.on("session_start", async (_event, ctx) => {
    if (!ctx.hasUI) return;
    ctx.ui.setWidget("status-bar", ["line one", "line two"], { placement: "belowEditor" });
  });
}
```

Load it, start a session. Expected: the widget renders below the prompt input, at the bottom of the screen. Actual: it renders above the prompt input, between the conversation transcript and the input box.

### Screenshot
See attached - custom two-line status widget rendered above the prompt input, with the built-in session/model status line below the input as expected.

### Notes
- Confirmed this isn't a bug in the extension itself: `placement: "belowEditor"` is passed explicitly and the RPC bridge's setWidget implementation forwards the string[] content but appears to not honor (or mishandle) the placement value.
- Suspect this lives in the RPC extension-UI bridge layer (`dist/modes/rpc/rpc-extension-ui-context.js` in the installed bundle) given the daemon/worker architecture that routes extension UI calls over RPC to the terminal-owning client.

andreolf:
## Problem

In fullscreen mode, extension widgets registered with `placement: 'belowEditor'` render **above** the prompt editor instead of below it (#697). In non-fullscreen mode the placement is correct.

## Root cause

The placement value plumbs through correctly end to end, and non-fullscreen layout is right — `mainContainer` orders `widgetContainerBelow` after the editor. The bug is fullscreen-only, in `applyFullscreen()` (`packages/coding-agent/src/modes/interactive/interactive-mode.ts`):

```ts
this.ui.enterFullscreen({
  scroll: [ ..., this.widgetContainerBelow ],  // widget in the SCROLL region
  dock: this.promptDock,                        // editor lives in the dock
});
```

`getPromptDockComponents()` returns `[editorContainer, subagentSummaryLine, footerSlot]`, and `renderFullscreen()` (`packages/tui/src/tui.ts`) renders the entire `scroll` region above the `dock`. So `widgetContainerBelow` ends up above the editor — exactly the reported symptom.

## Fix

Move `widgetContainerBelow` out of the fullscreen `scroll` region and into the prompt dock, between `subagentSummaryLine` and `footerSlot`, so both layout paths agree.

This is safe for the non-fullscreen path:
- `promptDock` is only ever rendered as the fullscreen dock; it is never attached to the UI in non-fullscreen mode.
- `Container.addChild` (`packages/tui/src/tui.ts`) appends without reparenting, so adding `widgetContainerBelow` to `getPromptDockComponents()` does not remove it from `mainContainer`.
- `renderFullscreen()` renders only `scroll` + `dock` (never `mainContainer`), so there is no double render in fullscreen.

## Testing

- Reproduced with the in-repo `packages/coding-agent/examples/extensions/widget-placement.ts` example.
- `npm run check` passes (biome, `tsgo --noEmit`, installer render, browser smoke).

Fixes #697

---

> [!NOTE]
> **Low Risk**
> Small interactive-mode layout change only; non-fullscreen `mainContainer` ordering is unchanged and the dock is fullscreen-only.
> 
> **Overview**
> Fixes **fullscreen** layout so extension widgets with `placement: 'belowEditor'` sit **under** the prompt editor, matching non-fullscreen behavior ([#697](https://github.com/PrimeIntellect-ai/prime-agent/issues/697)).
> 
> Previously `widgetContainerBelow` was in the fullscreen **scroll** stack above the dock, so below-editor widgets appeared above the editor. The container is now part of **`getPromptDockComponents()`** (after `subagentSummaryLine`, before `footerSlot`) and is no longer listed in `applyFullscreen()`’s scroll region. The unreleased changelog records the fix.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 622d313789c5b5d6f8b9dda084df7f2d57305ee6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Fix `belowEditor` widgets rendering above the prompt editor in fullscreen mode
> Moves `widgetContainerBelow` from the scrollable region into the prompt dock in [interactive-mode.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/726/files#diff-b96137f89d422953665b334785273ae6cdbcb53ff37d8828db43f85038658316). This ensures extension `belowEditor` widgets appear below the prompt editor rather than above it when in fullscreen.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 622d313.</sup>
> 
>

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## [Unreleased]
 
+- Fixed extension `belowEditor` widgets rendering above the prompt editor in fullscreen mode ([#697](https://github.com/PrimeIntellect-ai/prime-agent/issues/697)).
+
 ## [0.7.0] - 2026-08-05
 
 ### Breaking Changes
--- a/packages/coding-agent/src/modes/interactive/interactive-mode.ts
+++ b/packages/coding-agent/src/modes/interactive/interactive-mode.ts
@@ -6899,7 +6899,7 @@ export class InteractiveMode {
 	}
 
 	private getPromptDockComponents(): Component[] {
-		return [this.editorContainer, this.subagentSummaryLine, this.footerSlot];
+		return [this.editorContainer, this.subagentSummaryLine, this.widgetContainerBelow, this.footerSlot];
 	}
 
 	/** Enter or leave fullscreen rendering without touching the persisted setting. */
@@ -6912,7 +6912,6 @@ export class InteractiveMode {
 					this.mainViewContainer,
 					this.widgetContainerAbove,
 					...this.getPromptContextContainers(),
-					this.widgetContainerBelow,
 				],
 				dock: this.promptDock,
 				mouse: this.settingsManager.getFullscreenMouse(),


andreolf:
Good catch from the bot — the finding is valid. Confirming the mechanism and proposing a fix.

`composeFrame` in `packages/tui/src/fullscreen.ts` clips an overflowing dock by keeping the bottom rows:

```ts
const dockHeight = clippedFullscreenDockHeight(dockLines.length, height);
if (dockLines.length > dockHeight) {
    // bottom of the dock (editor + footer) wins over widgets above it
    dockLines = dockLines.slice(dockLines.length - dockHeight);
}
```

With this PR the dock order is `[editor, subagentSummary, widgetBelow, footer]`, so the editor is at the **top** of the dock and is the first thing dropped when the dock overflows. A `belowEditor` widget tall enough to overflow the fullscreen dock (small terminal or a large custom widget) can therefore push the editor off-screen while the widget and footer remain — exactly as described. Note the existing clip comment assumes "editor + footer wins over widgets above it," which held when widgets lived in the scroll region; moving the widget into the dock between the editor and footer breaks that assumption.

Scope for reviewers: the original bug (#697) affected **every** `belowEditor` widget unconditionally in fullscreen; this regression is narrower (only widgets tall enough to overflow the dock), but it shouldn't ship as-is.

Proposed fix: preserve the editor rows so the **widget** yields, not the editor — i.e. when the dock would overflow, cap/scroll `widgetContainerBelow` so the editor (and footer) stay visible, keeping the "editor + footer wins" guarantee intact. Concretely, reserve `editor + subagentSummary + footer` height first and give `widgetContainerBelow` the remainder in fullscreen.

I'm happy to add that guard to this PR, or land the current fix (which resolves the common case) and do the overflow guard as a follow-up — whichever you prefer for scope. Since it touches shared fullscreen clip behavior, I'd want to verify it in a real terminal at a few sizes before pushing. Let me know.

kevinjosethomas:

PR #727, opened 2026-08-06
fix(coding-agent): reclaim stale session leases on Windows

### LINKED ISSUE #667: Windows: a stale session lease blocks session recovery
## Environment

- prime-agent 0.7.0

## Observation

## Reproduction

1. Start a session.
3. Restart prime-agent.
4. Try to resume the session.

```
```

## Cause

## Impact

## Suggested fix

andreolf:
## Problem

On Windows, a stale session lease permanently blocks session recovery (#667). Once a lease directory is left behind by a dead owner, acquiring the session keeps failing and never reclaims it.

## Root cause

`acquireSessionLease()` (`packages/coding-agent/src/core/session-lease.ts`) claims a lease by renaming a candidate directory onto the lease directory:

```ts
try {
  renameSync(candidateDirectory, directory);
  ...
} catch (error) {
  const code = (error as NodeJS.ErrnoException).code;
  if (code !== "EEXIST" && code !== "ENOTEMPTY") {
    throw error;                 // Windows lands here
  }
  ...
  reclaimStaleLease(directory);  // never reached on Windows
}
```

`renameSync` onto an existing non-empty directory returns `EEXIST`/`ENOTEMPTY` on POSIX, but **`EPERM`/`EACCES` on Windows**. So on Windows the error is rethrown, `reclaimStaleLease()` never runs, and the stale lease persists across every retry.

## Fix

Classify `EPERM`/`EACCES` as a rename-target-exists condition on `win32`, via a small `isRenameTargetExistsError(code, platform)` helper. This mirrors the platform-specific handling already present in this file (`getProcessStartId` / `getWindowsProcessStartId`) and keeps POSIX behavior unchanged — `EPERM`/`EACCES` are only accepted on `win32`, so genuine permission failures on Linux/macOS still surface.

## Testing

- Added `test/suite/regressions/667-windows-stale-lease.test.ts` covering the code/platform matrix: POSIX codes accepted everywhere, `EPERM`/`EACCES` accepted only on `win32`, and unrelated codes (`ENOENT`, `EBADF`, `undefined`) still rethrown.
- New test plus the existing `test/session-lease.test.ts` pass.
- `npm run check` passes (biome, `tsgo --noEmit`, installer render, browser smoke).

## Note

Verified by code inspection on macOS; the `win32` runtime path itself was not exercised on a Windows host. The behavioral change is gated to `platform === "win32"`, so it cannot regress POSIX. A confirmation on a Windows box would be welcome.

Fixes #667

---

> [!NOTE]
> **Low Risk**
> Small, platform-gated change to lease error handling on win32 only; POSIX behavior preserved and covered by unit tests.
> 
> **Overview**
> Fixes **session recovery on Windows** when a stale lease directory is left behind: `renameSync` onto an existing lease path now treats **`EPERM`/`EACCES` on `win32`** like POSIX **`EEXIST`/`ENOTEMPTY`**, so `acquireSessionLease` can run **`reclaimStaleLease`** instead of rethrowing and blocking forever ([#667](https://github.com/PrimeIntellect-ai/prime-agent/issues/667)).
> 
> Adds exported **`isRenameTargetExistsError(code, platform)`** in `session-lease.ts` and wires it into the lease-acquire rename catch path. **Linux/macOS** still rethrow **`EPERM`/`EACCES`**; only Windows gets the extra codes. Regression tests cover the code/platform matrix; changelog updated under Unreleased.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit aeb1d3111c18a7db101655acf1dbf34bd9cbb3da. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Fix stale session lease reclamation on Windows in coding-agent
> - On Windows, `renameSync` raises `EPERM`/`EACCES` instead of `EEXIST`/`ENOTEMPTY` when the rename target already exists, blocking session recovery. A new `isRenameTargetExistsError` utility in [`session-lease.ts`](https://github.com/PrimeIntellect-ai/prime-agent/pull/727/files#diff-74865e03a1e2d7f4025a576ca8f4c09dbc1ec7c9e2776b5d93f2691ee29e6cda) now treats these Windows-specific codes as target-exists errors.
> - `readLeaseOwner` now returns a discriminated union (`"owner"` | `"absent"` | `"unreadable"`) so callers can distinguish a missing `owner.json` from one that exists but cannot be read.
> - When `owner.json` is unreadable, `acquireSessionLease` throws `SessionAlreadyActiveError` instead of reclaiming the lease, avoiding takeover of a possibly-active session.
> - Regression tests added for `isRenameTargetExistsError` and the unreadable-owner case.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized c88a0e2.</sup>
> 
>

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## [Unreleased]
 
+- Fixed session recovery on Windows being permanently blocked by a stale lease, because `renameSync` reports `EPERM`/`EACCES` for an existing target there instead of `EEXIST`/`ENOTEMPTY` ([#667](https://github.com/PrimeIntellect-ai/prime-agent/issues/667)).
+
 ## [0.7.0] - 2026-08-05
 
 ### Breaking Changes
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [Unreleased]
 
 - Fixed session recovery on Windows being permanently blocked by a stale lease, because `renameSync` reports `EPERM`/`EACCES` for an existing target there instead of `EEXIST`/`ENOTEMPTY` ([#667](https://github.com/PrimeIntellect-ai/prime-agent/issues/667)).
+- Fixed a live session lease being reclaimed when its `owner.json` was momentarily unreadable (e.g. a briefly held handle from antivirus or the search indexer on Windows); an unreadable owner is now treated as active instead of stale ([#667](https://github.com/PrimeIntellect-ai/prime-agent/issues/667)).
 
 ## [0.7.0] - 2026-08-05
 
--- a/packages/coding-agent/src/core/session-lease.ts
+++ b/packages/coding-agent/src/core/session-lease.ts
@@ -49,8 +49,8 @@ export class SessionLease {
 		this.released = true;
 		try {
 			withLeaseGuard(this.directory, () => {
-				const owner = readLeaseOwner(this.directory);
-				if (owner?.token === this.token) {
+				const read = readLeaseOwner(this.directory);
+				if (read.status === "owner" && read.owner.token === this.token) {
 					rmSync(this.directory, { recursive: true, force: true });
 				}
 			});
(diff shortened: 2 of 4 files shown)


BILLKISHORE:
The `EPERM`/`EACCES` mapping looks right, but widening the filter exposes a pre-existing hole downstream that is much easier to hit on Windows.

`readLeaseOwner` (`session-lease.ts:86-102`) catches every error and returns `undefined`, including a transient read failure on `owner.json`. The caller at `:271` treats `undefined` as "no owner", so it skips the `isLeaseOwnerAlive` guard entirely and falls through to `reclaimStaleLease`:

```ts
const existingOwner = readLeaseOwner(directory);
if (existingOwner && isLeaseOwnerAlive(existingOwner)) {
    throw new SessionAlreadyActiveError(canonicalPath, existingOwner.activeSessionId);
}
reclaimStaleLease(directory);
```

So the sequence is: a live session holds the lease, something briefly locks `owner.json`, a second process fails the rename, the error now passes the widened filter, the owner reads back as `undefined`, and a live lease gets reclaimed. Two processes then believe they own the session.

This path is reachable on main today through `EEXIST`, but an unreadable `owner.json` is rare on POSIX. On Windows, transient locking from antivirus or the search indexer is routine, so this PR makes a narrow hole considerably wider on the platform where it is most likely to fire.

The fix that fits the file's own convention is to distinguish "no owner" from "owner unreadable" and treat unreadable as alive. `isProcessAlive` at `:104-111` already does exactly this, returning `true` on `EPERM` rather than assuming the process is gone.

I have not reproduced this: it needs a Windows host and a held handle on `owner.json` while a live session owns the lease. The code path is from reading the source, so it is worth confirming before acting on it.

andreolf:
Great catch, thank you — you're right, and I've pushed a fix.

`readLeaseOwner` collapsed "owner.json absent" and "owner.json unreadable" into a single `undefined`, so a transient read failure fell through to `reclaimStaleLease` and a live lease could be reclaimed — and widening the rename filter to `EPERM`/`EACCES` made that path far easier to hit on Windows, exactly as you described.

Fixed as you suggested — distinguish the two and treat unreadable as live, following the file's own `isProcessAlive`/`EPERM` convention. `readLeaseOwner` now returns `{ status: "owner" | "absent" | "unreadable" }`: a missing file is `absent` (reclaimable), any other read error is `unreadable` and treated as an active owner (`SessionAlreadyActiveError`), never reclaimed. Updated all three call sites.

Added a cross-platform regression test that doesn't need a Windows host: it makes `owner.json` a directory so `readFileSync` throws `EISDIR` (present-but-unreadable), and asserts the lease is not reclaimed. `npm run check` and the lease tests pass.

kevinjosethomas:

PR #670, opened 2026-08-06
fix(coding-agent): tolerate unsupported directory fsync

GGomesBRA:
Fixes #666

## Summary

Make command journal compaction tolerate platforms that do not support `fsync` on directory handles.

## Root cause

`CommandRecoveryJournal.compact()` atomically renames the temporary journal and then calls `fsyncSync()` on the containing directory. On Windows, Node maps this to `FlushFileBuffers()` on a directory handle and returns `EPERM`. The exception escapes `acknowledge()`, so every daemon command acknowledgment logs an error and can leave the daemon unable to create or authenticate workers.

## Change

- Keep the temporary file `fsync` and atomic `renameSync`.
- Treat the post-rename directory `fsync` as best effort, matching the existing `cron-jobs` persistence path.
- Document the fix in the coding-agent changelog.

## Validation

- On Windows, the existing command journal test failed before the change: 5 passed, 1 failed at the directory `fsync`.
- The same test passes after the change: 6 passed.
- `npm run check` passes: Biome, TypeScript, installer render, and browser smoke checks.

> [!NOTE]
> ### Fix `CommandRecoveryJournal` to tolerate platforms without directory fsync support
> On some platforms, fsyncing a directory after an atomic rename is unsupported and throws. The directory open/fsync/close in [command-recovery-journal.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/670/files#diff-ad9ad93f7564a34f4c838495026b460ad85292338ab337c5cdca8f48161ecd3a) is now wrapped in a try/catch that swallows errors from missing directory fsync support. The atomic rename still occurs, so journal writes and compaction proceed normally on all platforms.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized e728af6.</sup>
> 
>

---

> [!NOTE]
> **Low Risk**
> Small, localized durability tweak on an already-best-effort directory fsync; temp file fsync and atomic rename are unchanged.
> 
> **Overview**
> **Command recovery journal compaction** no longer fails when the OS rejects `fsync` on the journal directory (notably Windows `EPERM`). After the temp-file fsync and atomic `renameSync`, directory fsync is **best-effort** inside a try/catch—the same pattern as cron job persistence—so `acknowledge()` and compaction complete instead of surfacing errors that could break daemon command acknowledgment and worker auth.
> 
> The unreleased changelog notes the fix for [#666](https://github.com/PrimeIntellect-ai/prime-agent/issues/666).
> 
[...]

PR #695, opened 2026-08-06
fix(coding-agent): bootstrap IPython kernel venv with the Windows interpreter path

MarcusNeufeldt:
﻿## Summary

Fixes IPython kernel bootstrap on Windows. The kernel bootstrapper hardcoded the POSIX venv interpreter path (`venv/bin/python`) in both `bootstrapVenv` and the resolution path in `ensureKernelPythonUncached`. On Windows a uv-created venv places the interpreter at `venv/Scripts/python.exe`, so `uv pip install --python <venv>/bin/python ...` failed with exit code 2 and the IPython tool could never start.

## Change

- Added a small `kernelVenvPython(venv)` helper that selects `Scripts\python.exe` on `win32` and `bin\python` elsewhere (mirroring the existing `process.platform === "win32"` pattern already used in this file).
- Used it in both places that need the venv interpreter path.
- Added a win32-layout test to `kernel-bootstrap.test.ts` (the fake uv now supports a Windows-style venv shape).
- Updated `docs/rlm-runtime.md` and a changelog entry.

The POSIX path is unchanged, so existing Linux/CI behavior is preserved (the helper returns `bin/python` when `process.platform` is not `win32`; existing tests assert that path and stay green).

## Testing

- `npm run check` pre-commit hook: biome + `tsgo --noEmit` pass.
- Manually verified the Windows path on a real Windows host: a patched install rebuilt the managed venv with the `Scripts\python.exe` layout, installed `ipykernel` + `prime-agent-runtime` + default packages, and a launched IPython kernel executed `print(6*7)` correctly.

> [!NOTE]
> ### Fix IPython kernel venv bootstrap to use `Scripts/python.exe` on Windows
> The venv interpreter path was hardcoded to the POSIX `bin/python` layout, causing bootstrap failures on Windows. A new `kernelVenvPython` helper in [`bootstrap.ts`](https://github.com/PrimeIntellect-ai/prime-agent/pull/695/files#diff-e87b9a04ab9b4ac02e8dd33af2f2f58320d8b7f6be6853acf4a3e05a34623cae) resolves to `venv/Scripts/python.exe` on `win32` and `venv/bin/python` otherwise. This path is now used consistently across pip installs, skill sync, and the `ensureKernelPython` readiness check.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 93eb5a2.</sup>
> 
>

---

> [!NOTE]
> **Low Risk**
[...]

PR #744, opened 2026-08-06
feat(coding-agent): add native Windows support

### LINKED ISSUE #660: Windows: kernel bootstrap uses venv bin/python, so the IPython kernel never starts and each retry wipes the venv

### LINKED ISSUE #665: Windows: no supported install path (installer and docs cover macOS/Linux only)
## Summary

README and `docs/quickstart.md` document installation only for macOS and Linux (`curl -fsSL .../install.sh | sh`). The runtime itself already treats Windows as a platform: `docs/windows.md` documents the Git Bash requirement, the coding agent has `win32` branches (`findExecutable`, `ensureUv`), and #660 / #663 come from a real Windows install. But there is no documented or supported way to install Prime Agent on Windows.

## Current behavior on Windows

- Under Git Bash, `install.sh` works when Node >= 20.6 is already installed: the release tarball and `SHA256SUMS` download, `sha256sum` is available for verification, and the `npm install -g` step succeeds. This is the path the reporter of #660 used ("installed via the official `install.sh` -> `npm install -g` path" on Windows 11).
- If Node is missing, the installer cannot recover: `detect_node_install_method` falls through to `standalone`, and the standalone flow only maps `Darwin`/`Linux` — `uname -s` reports `MINGW64_NT-...` under Git Bash, so it ends in "Unsupported operating system for automatic Node.js install".
- From native PowerShell/cmd there is no path at all; the installer is POSIX sh.

## Why a Windows install path is viable

The install payload is platform-independent: `prime-agent-<version>.tgz` is an npm package installed with `npm install -g`, and a per-release `SHA256SUMS` is already published next to it. Nothing in the artifact is macOS/Linux-specific. The release workflow already renders `install.sh` (base URL substitution in `build-binaries.yml`) and uploads it to the public bucket, so publishing a Windows installer alongside it is mechanical.

## Proposal

In increasing effort:

1. Document the supported manual path: with Node >= 20.6 and Git for Windows installed, the existing `curl ... | sh` line works from Git Bash. Add a Windows note to README / quickstart pointing at `docs/windows.md` for the runtime shell requirement.
2. Publish an `install.ps1` rendered and uploaded by the same release workflow: check Node/npm >= 20.6, resolve the channel version, download the tarball and `SHA256SUMS`, verify with `Get-FileHash`, run `npm install -g`, optionally bootstrap the kernel runtime (`uv` ships Windows builds). That gives an `irm ... | iex` install without a bash prerequisite at install time; the runtime requirement from `docs/windows.md` still applies and the installer can check for it.

Either way this probably should not be advertised in the README until #663 lands, since a fresh Windows install currently comes up with a non-working kernel (#660).

Related: #660 / #663 cover a runtime bug hit after installing on Windows; this issue is only about the install path.

## Environment checked

Windows 11 Pro, Git Bash (`uname -s` = `MINGW64_NT-10.0-26300`), Node 24, repo at v0.7.0 (`c98941a2`).

I can submit a PR for either option.

### LINKED ISSUE #666: Windows: fsync on a directory handle fails with EPERM
## Environment

- prime-agent 0.7.0

## Observation

On Windows, `CommandRecoveryJournal.compact()` fails with `EPERM`. The daemon writes this error to its log on every command acknowledgment.

## Reproduction

1. Open a directory with `fs.openSync()`.
2. Call `fs.fsyncSync()` on the directory descriptor.
3. Observe the error `EPERM: operation not permitted, fsync`.

The daemon log shows this stack trace:

```
Supervisor command ack_result failed: Error: EPERM: operation not permitted, fsync
    at fsyncSync (node:fs:1346:11)
    at CommandRecoveryJournal.compact (.../dist/bundle/chunk-L3VO7F2S.js:56586:7)
```

## Cause

The code in `CommandRecoveryJournal.compact()` calls `fsyncSync()` on a directory handle. Windows does not permit `FlushFileBuffers()` on a directory. This POSIX idiom works on Linux and macOS.

## Impact

Every daemon command acknowledgment throws an error. The command journal never compacts. Crash recovery does not work completely. In one session, the catalog process exited with code 1.

## Related code

The `writeJobsState()` function in `chunk-ALQBG3TN.js` wraps the same operation in `try/catch`. The journal site does not have this guard.

## Suggested fix

Wrap the directory `fsync` call in `try/catch`. Or skip the directory `fsync` on Windows.

### LINKED ISSUE #667: Windows: a stale session lease blocks session recovery

### LINKED ISSUE #668: Windows: detached child processes open visible console windows
## Environment

- prime-agent 0.7.0

## Observation

On Windows, each detached child process opens a new visible console window. The daemon and its workers run without a console. A child of a process that has no console gets a new console window. This behavior occurs unless the child is created with `windowsHide: true`.

## Locations

These spawn sites use `detached: true` without `windowsHide: true`:

- `dist/cli/daemon-launch.js:315`
- `dist/modes/daemon/daemon-supervisor.js:1730` and `:4133`
- `dist/core/kernel/bootstrap.js` in `run()` (line 322)
- `dist/core/tools/bash.js:36`

## Impact

Any crash or retry loop opens many console windows in quick succession. The user sees a rapid flicker on the desktop. In our test, a recovery loop opened a new window for each worker restart.

## Also observed

The supervisor runs `cli.js --version` as a child of the daemon. This child process stayed alive for more than 90 seconds. It opens another console window and does not exit.

## Suggested fix

Set `windowsHide: true` on every child process spawn on Windows.

WDDGDA:
## Summary

- add a PowerShell 5.1 installer with strict SemVer parsing, SHA-256 verification, x64/Git-for-Windows checks, and release-channel integration
- fix Windows kernel venv resolution, directory-fsync handling, stale lease recovery, hidden child processes, and named-pipe test fixtures
- add fail-closed Windows Job Object isolation so daemon Bash trees are assigned before commands run and are terminated after worker crashes
- replace repeated PowerShell PID queries with native `GetProcessTimes`, and add Windows CI coverage for installer, kernel, daemon, and recovery paths

## Process safety

- Windows daemon isolation supports x64 Node.js and Git for Windows only
- worker and replacement supervisor processes remain outside the Job; only gated Bash trees are assigned
- native/assignment failures reject command execution instead of degrading to an unisolated mode
- no persistent disk or network I/O is added by Job isolation

## Validation

- clean latest-main worktree: fresh `npm ci`, root build, checks, and Windows-targeted tests (89 passed, 7 platform skips)
- final shell-contract update: TypeScript and Biome checks plus Windows-targeted tests (90 passed, 7 platform skips)
- packaged install smoke: global package loaded exact `koffi@2.16.2`; hard worker kill removed the Bash leaf and a replacement worker became ready
- public diff scanned for credentials, private keys, personal paths, email addresses, and non-test localhost URLs

Builds on the reports and proposed fixes in #663, #664, and #670.

Closes #660
Closes #665
Closes #666
Closes #667
Closes #668

---

> [!NOTE]
> **High Risk**
> Changes authentication-adjacent daemon lifecycle, session leases, and child-process termination on Windows using memory-unsafe FFI; failures are intended to fail closed but mis-assignment or Win32 errors could block Bash or leave orphans.
> 
> **Overview**
> Adds **native Windows x64** support: a published **PowerShell installer**, release/CI wiring, and runtime fixes so daemon workers, the Python kernel, and Bash behave correctly on Windows.
> 
> **Install & docs:** New `install.ps1` (channel, SHA-256, x64 Node 22.8+, Git for Windows) is rendered and uploaded beside the shell installers; README/quickstart document `irm … | iex`. Root `check:installer` runs `check-windows-installer.mjs`. **`koffi@2.16.2`** is a direct dependency for Win32 FFI.
> 
> **Daemon process isolation:** Daemon workers create a **kill-on-close Job Object** and gate Bash through Git’s direct `usr\bin\bash.exe` so the process is assigned to the job before descendants start; worker crash tears down that Bash tree while supervisors stay outside the job. Session lease identity on Windows uses **`GetProcessTimes`** via Koffi instead of spawning PowerShell per PID; lease/supervisor lock renames tolerate Windows `EPERM` when the destination already exists.
> 
> **Cross-platform spawn hygiene:** `windowsHide: true` on spawns; `signalProcessTree` / `taskkill` for tree kills; kernel bootstrap uses `Scripts\python.exe` and PowerShell `uv` install on Windows; command-recovery journal skips directory fsync when unsupported.
> 
> **CI:** New `windows-latest` job (installer check, Windows-targeted vitest, kernel bootstrap + import smoke) gates `build-check-test`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0ef7b8e56dc5f2de92eb230c53ee409c33c8ee53. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Add native Windows support with PowerShell installer, Job Object isolation, and hidden console windows
> - Adds [install.ps1](https://github.com/PrimeIntellect-ai/prime-agent/pull/744/files#diff-50d51a233d8ff4886061fc5cf1a03642de1f64b3625903e4f876e8cd2a3f5806), a PowerShell installer for Windows x64 that requires Node.js ≥22.8.0 and Git Bash, downloads the release tarball, verifies SHA-256, and installs via npm.
> - Introduces [windows-job-object.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/744/files#diff-336c1a599b762f2a14274cdfbf05023b45f26c7166c40afbeb5a13e8dc3f9b0d) using `koffi` to bind Win32 Job Object APIs, enabling reliable process-tree termination for daemon workers on Windows.
> - Adds `windowsHide: true` to all `spawn`/`spawnSync` calls across the codebase so no console windows appear on Windows.
> - Fixes `venvPython` to use `Scripts/python.exe` on Windows and updates `ensureUv` to invoke the PowerShell-based uv installer instead of `sh`.
> - Fixes session lease acquisition to treat `EPERM` on rename (when destination exists) as contention rather than throwing, matching Windows filesystem semantics.
> - Adds new Git-for-Windows shell helpers (`gitForWindowsRoot`, `getDirectWindowsBashPath`, `getWindowsGitBashLauncherPath`) and probes `LOCALAPPDATA\Programs\Git` for bash detection.
> - Risk: `koffi` is added as a runtime dependency; Windows-specific paths (Job Objects, named pipes, junction symlinks) are active only on `win32`.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized b9581b3.</sup>
> 
>

### DIFF
--- a/.github/workflows/build-binaries.yml
+++ b/.github/workflows/build-binaries.yml
@@ -237,11 +237,14 @@ jobs:
           const baseUrl = process.env.INSTALL_BASE_URL;
           if (!baseUrl) throw new Error("INSTALL_BASE_URL is required");
           const installer = fs.readFileSync("install.sh", "utf8");
-          const renderInstaller = (channel) => installer
+          const windowsInstaller = fs.readFileSync("install.ps1", "utf8");
+          const renderInstaller = (source, channel) => source
             .replaceAll("__PRIME_AGENT_DOWNLOAD_BASE_URL__", baseUrl)
             .replaceAll("__PRIME_AGENT_DEFAULT_RELEASE_CHANNEL__", channel);
-          fs.writeFileSync("/tmp/prime-agent-install.sh", renderInstaller("stable"));
-          fs.writeFileSync("/tmp/prime-agent-install-beta.sh", renderInstaller("beta"));
+          fs.writeFileSync("/tmp/prime-agent-install.sh", renderInstaller(installer, "stable"));
+          fs.writeFileSync("/tmp/prime-agent-install-beta.sh", renderInstaller(installer, "beta"));
+          fs.writeFileSync("/tmp/prime-agent-install.ps1", renderInstaller(windowsInstaller, "stable"));
+          fs.writeFileSync("/tmp/prime-agent-install-beta.ps1", renderInstaller(windowsInstaller, "beta"));
           NODE
 
       - name: Extract production release notes
@@ -296,6 +299,16 @@ jobs:
           aws s3 cp /tmp/prime-agent-install-beta.sh "s3://${R2_BUCKET}/install-beta.sh" \
             --endpoint-url "$R2_ENDPOINT_URL" \
             --content-type text/x-shellscript \
+            --cache-control no-cache
+
+          aws s3 cp /tmp/prime-agent-install.ps1 "s3://${R2_BUCKET}/install.ps1" \
+            --endpoint-url "$R2_ENDPOINT_URL" \
+            --content-type text/plain \
+            --cache-control no-cache
+
+          aws s3 cp /tmp/prime-agent-install-beta.ps1 "s3://${R2_BUCKET}/install-beta.ps1" \
+            --endpoint-url "$R2_ENDPOINT_URL" \
+            --content-type text/plain \
             --cache-control no-cache
 
       - name: Create production GitHub release
@@ -379,6 +392,16 @@ jobs:
             --content-type text/x-shellscript \
             --cache-control no-cache
 
+          aws s3 cp /tmp/prime-agent-install.ps1 "s3://${R2_BUCKET}/install.ps1" \
+            --endpoint-url "$R2_ENDPOINT_URL" \
+            --content-type text/plain \
+            --cache-control no-cache
+
+          aws s3 cp /tmp/prime-agent-install-beta.ps1 "s3://${R2_BUCKET}/install-beta.ps1" \
+            --endpoint-url "$R2_ENDPOINT_URL" \
+            --content-type text/plain \
+            --cache-control no-cache
+
           printf 'Automated beta build from `%s` (`%s`).\n' "$DEFAULT_BRANCH" "$BUILD_REF" > /tmp/beta-release-notes.md
 
           if gh api "repos/${GITHUB_REPOSITORY}/git/ref/tags/beta" >/dev/null 2>&1; then
@@ -409,4 +432,4 @@ jobs:
           fi
 
           gh release upload beta "$BETA_DIR"/* --clobber
-          echo "Beta installer: ${R2_PUBLIC_BASE_URL%/}/install-beta.sh"
+          echo "Beta installers: ${R2_PUBLIC_BASE_URL%/}/install-beta.sh and ${R2_PUBLIC_BASE_URL%/}/install-beta.ps1"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -115,16 +115,67 @@ jobs:
         working-directory: ${{ matrix.package }}
         run: ${{ matrix.command }}
 
+  windows:
+    name: Windows native smoke
+    runs-on: windows-latest
+    timeout-minutes: 15
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v7
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v7.0.0
+        with:
+          node-version: 22
+          cache: npm
+
+      - name: Install dependencies
+        run: npm ci
+
+      - name: Install uv
+        run: python -m pip install uv
+
+      - name: Check PowerShell installer
+        run: node scripts/check-windows-installer.mjs
+
+      - name: Test Windows runtime paths
+        working-directory: packages/coding-agent
+        shell: pwsh
+        run: |
+          npx tsx ../../node_modules/vitest/dist/cli.js --run test/command-recovery-journal.test.ts
+          if ($LASTEXITCODE -ne 0) { exit $LASTEXITCODE }
+          npx tsx ../../node_modules/vitest/dist/cli.js --run test/session-lease.test.ts
+          if ($LASTEXITCODE -ne 0) { exit $LASTEXITCODE }
+          npx tsx ../../node_modules/vitest/dist/cli.js --run test/daemon-launch.test.ts
+          if ($LASTEXITCODE -ne 0) { exit $LASTEXITCODE }
+          npx tsx ../../node_modules/vitest/dist/cli.js --run test/daemon-catalog-process.test.ts
+          if ($LASTEXITCODE -ne 0) { exit $LASTEXITCODE }
+          npx tsx ../../node_modules/vitest/dist/cli.js --run test/windows-job-object.test.ts
+          if ($LASTEXITCODE -ne 0) { exit $LASTEXITCODE }
+          npx tsx ../../node_modules/vitest/dist/cli.js --run test/daemon-supervisor-process.test.ts
+          if ($LASTEXITCODE -ne 0) { exit $LASTEXITCODE }
+          npx tsx ../../node_modules/vitest/dist/cli.js --run test/suite/agent-session-autonomous.test.ts test/tools.test.ts
+          if ($LASTEXITCODE -ne 0) { exit $LASTEXITCODE }
+          npx tsx ../../node_modules/vitest/dist/cli.js --run test/bash-close-hang-windows.test.ts
+          if ($LASTEXITCODE -ne 0) { exit $LASTEXITCODE }
+          $env:PRIME_AGENT_KERNEL_VENV = Join-Path $env:RUNNER_TEMP "prime-agent-kernel-venv"
+          npx tsx src/core/kernel/bootstrap-cli.ts
+          if ($LASTEXITCODE -ne 0) { exit $LASTEXITCODE }
+          & "$env:PRIME_AGENT_KERNEL_VENV\Scripts\python.exe" -c "import ipykernel, rlm, dill"
+          if ($LASTEXITCODE -ne 0) { exit $LASTEXITCODE }
+
   build-check-test:
     name: build-check-test
     if: always()
-    needs: [build-check, test]
+    needs: [build-check, test, windows]
     runs-on: ubuntu-latest
     steps:
       - name: Verify CI results
         env:
           BUILD_CHECK_RESULT: ${{ needs.build-check.result }}
           TEST_RESULT: ${{ needs.test.result }}
+          WINDOWS_RESULT: ${{ needs.windows.result }}
         run: |
           test "$BUILD_CHECK_RESULT" = success
           test "$TEST_RESULT" = success
+          test "$WINDOWS_RESULT" = success
--- a/README.md
+++ b/README.md
@@ -51,7 +51,13 @@ Install the latest stable release on macOS or Linux:
 curl -fsSL https://app.primeintellect.ai/prime-agent/install.sh | sh
 ```
 
-The installer downloads a versioned release, verifies its SHA-256 checksum, installs the `prime-agent` command, and can prepare the IPython runtime used by the agent.
+On Windows x64, install x64 Node.js 22.8.0 or newer and Git for Windows, then run in PowerShell:
+
+```powershell
+irm https://app.primeintellect.ai/prime-agent/install.ps1 | iex
+```
+
+The installers download a versioned release, verify its SHA-256 checksum, install the `prime-agent` command, and can prepare the IPython runtime used by the agent.
 
 Start Prime Agent from the repository or directory you want it to work in:
 
--- a/install.ps1
+++ b/install.ps1
@@ -1,0 +1,154 @@
+[CmdletBinding()]
+param(
+	[ValidateSet("stable", "beta")]
+	[string]$Channel = "__PRIME_AGENT_DEFAULT_RELEASE_CHANNEL__"
+)
+
+Set-StrictMode -Version Latest
+$ErrorActionPreference = "Stop"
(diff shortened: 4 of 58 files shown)


FlamesONE:
Tested this branch on real Windows hardware, since CI has no Windows runner for the suite.

Environment: Windows 11 Pro 26200 (x64), Node 24.15.0, npm 11.x, Python 3.11 via uv 0.11.6, Git Bash 2.51.0. Branch `pr-744` at 2aa1a10, compared against `main` at c5991bc, both freshly built with `npm run build`.

## Results

- `npm run check`: passes.
- `test/windows-job-object.test.ts` and `test/bash-close-hang-windows.test.ts`: pass (5/5). Note both need the workspace packages built first; with a stale `packages/tui/dist` the Job Object test fails with `Cannot find package '@earendil-works/pi-tui'` from the worker fixture, which reads as a Job Object failure but is not one.
- Full `packages/coding-agent` suite:

  | branch | failing tests | failing files |
  | --- | --- | --- |
  | main | 137 | 44 |
  | this PR | 93 | 35 |

  44 tests fixed, no regressions: every test failing on this branch also fails on `main`.

Manual end-to-end check: the kernel starts from the built `dist`, resolves to `kernel-venv\Scripts\python.exe`, executes cells, and runs `subprocess` calls. The kernel process has `MainWindowHandle=0`, so the console window flashing from #668 is gone. `doctor` reports the daemon on `\\.\pipe\prime-agent-daemon`, and `model list` works.

## Two things still blocking a green Windows suite

**1. `test/kernel-bootstrap.test.ts` still hardcodes the POSIX path (17 failures).**

The production path moved to `venvPython()`, but the test did not:

```
test/kernel-bootstrap.test.ts:182
await expect(ensureKernelPython()).resolves.toBe(join(venv, "bin", "python"));

Expected: "...\kernel-venv\bin\python"
Received: "...\kernel-venv\Scripts\python.exe"
```

These 17 also fail on `main`, but for the original reason, so this is not a regression — just the last piece of the same change. Exporting `venvPython()` (or a test-local equivalent) and using it in the assertions should clear all of them.

**2. `bootstrapVenv` deletes a venv whose `python.exe` is still mapped (about 25 failures).**

Every test that needs a live kernel and hits a bootstrap-version mismatch fails with:

```
Error: Failed to set up the Python kernel runtime.
EPERM: operation not permitted, unlink 'C:\Users\user\.prime\agent\kernel-venv\Scripts\python.exe'
```

Windows refuses to unlink a running image, so the `rm(venv, { recursive: true, force: true })` before a rebuild cannot succeed while any kernel from that venv is alive. Affected files include `kernel-attach-image-skill` (9), `kernel-agent-message-skill` (6), `kernel-rlm-heartbeat-skill` (3), `kernel-goal-skill` (3), `kernel-agent-observe-skill` (2). This one is independent of this PR, so I opened #763 against main with a rename-then-delete plus tests. Nothing here needs to change for it.

The remaining failures (`config`, `agents-view-*`, `package-command-paths`, `ipython-provisioner`, several `test/suite/regressions/*`) fail identically on `main` and look out of scope here: path-separator assertions, `chmod` mode expectations such as `expected 438 to be 384`, and socket listen errors.

## One suggestion

`initializeWindowsDaemonWorkerJob()` throws on any non-x64 arch:

```ts
if (process.arch !== "x64") {
    throw new Error(`Windows daemon process isolation requires x64, received ${process.arch}`);
}
```

`JOBOBJECT_EXTENDED_LIMIT_INFORMATION` is 144 bytes on every 64-bit Windows target, including arm64 — each member is either a fixed-width type or pointer-sized, so the x64 layout is also the arm64 layout. koffi ships a `win32_arm64` build, and `main.ts` calls this without a `try`, so on Windows on ARM the daemon worker would die at startup rather than fall back. Widening the guard to reject only 32-bit arches would cover arm64 for free. I could not verify this on real arm64 hardware.

Separately, I checked whether `GetLastError()` through koffi returns stale values across calls, since the error codes here would be misleading if it did. It does not — `OpenProcess` on an invalid pid reports 87, and on pid 4 reports 5, both correct.

lennix1337:
While testing the latest PR commit on Windows 11, I found one remaining visible-console issue in clipboard image handling: each `powershell.exe` clipboard probe could open another CMD/console window. This is fixed in follow-up commit [`0ef7b8e`](https://github.com/PrimeIntellect-ai/prime-agent/commit/0ef7b8e56dc5f2de92eb230c53ee409c33c8ee53): `windowsHide: true` is now passed to the clipboard subprocess, with a regression assertion. Without this, repeated clipboard operations could leave multiple console windows open.

khanra17:
Windows follow-up from a native Windows 11 x64 check:

- #763 already covers the venv rebuild EPERM case from #660; this should be integrated/rebased rather than duplicated.
- I opened #917 for a separate recovery gap: `reapOrphanProcesses` polls on Windows but does not call `signalProcessTree`, so autonomous processes tracked outside the Job Object can survive a worker crash and make recovery fail.
- Two smaller portability details do not need separate PRs: the installer can derive Git Bash from `git.exe` when it is not in a standard directory, and `.cmd`/`.bat` command wrappers can be discovered for bootstrap commands. Neither changes the standard Git-for-Windows path.

SandroHub013:
I reproduced a Windows-specific gap not covered by the current native-support summary. Herdr sets `HERDR_SOCKET_PATH` to `C:\Users\...\AppData\Roaming\herdr\herdr.sock`, but its server endpoint is the named pipe `\\.\pipe\C:\Users\...\herdr.sock`. Passing the env value directly to `net.createConnection()` fails with `ENOTSOCK`; prefixing `\\.\pipe\` connects successfully.

I added a local adapter in `herdr-agent-state.ts` that preserves Unix paths and already-normalized pipe paths, plus Windows endpoint and session-path tests. This should be included in the umbrella Windows fix and covered by a native Windows Herdr smoke test.

grimmjoww:
Validated this branch on real Windows 11 hardware (x64, Node 24.18.0, npm 11.12.1, Git Bash at the standard path) and fixed one remaining gap on top of it: #917. On Windows, `reapOrphanProcesses` polls journaled orphans but never signals them, so orphans outside the worker's Job Object survive a worker crash and recovery fails after three attempts.

Stacked PR on your fork: WDDGDA/prime-agent#1. One commit: the fix, a win32 regression test (observed failing with the exact #917 signature before the fix, passing after), and a changelog bullet.

Numbers from this machine: targeted Windows files 12/12; `daemon-supervisor-process` 8/8 with 6 platform skips; full `coding-agent` suite 87 failed / 3918 passed, mostly kernel/python provisioning and POSIX literals baked into test data — none from this change. Also ran `install.ps1` end to end locally against a packed branch release: checksum-verified npm-global install, kernel venv provisioned, `prime-agent --version` printing from a fresh PowerShell.

@khanra17's two smaller points check out: #763 still reproduces by construction on this branch (`bootstrap.ts` does `rm(venv, { recursive: true })` over a mapped executable), and the Git-Bash-from-`git.exe` idea can't be exercised on a standard-layout machine like mine, so I left it unpatched rather than guessing. @lennix1337's clipboard fix is in this head and its tests pass.

BILLKISHORE:
The `destinationExists` gate on the lease rename is the right call, and it is stricter than the mapping in #727, which accepts `EPERM` without checking whether the target is actually there.

The path downstream of it is still open in both, though. `readLeaseOwner` (`session-lease.ts:86-102`) catches every error and returns `undefined`, and this branch leaves it unchanged. At `:271` an `undefined` owner skips the `isLeaseOwnerAlive` guard and falls straight through to `reclaimStaleLease`. So when the lease directory exists but `owner.json` is briefly unreadable, which on Windows is a routine antivirus or indexer interaction, a live lease can still be reclaimed and two processes end up believing they own the session.

`isProcessAlive` at `:104-111` already takes the safe side of exactly this question, returning `true` on `EPERM` rather than assuming the process is gone. Applying the same convention to `readLeaseOwner`, so that "unreadable" is distinct from "absent", would close it.

Separately, on the duplication @khanra17 raised for #763: it is wider than that one PR. This branch also reimplements the venv interpreter path (#663, #695), the directory fsync guard (#670), the lease `EPERM` mapping (#727), and the `windowsHide` spawn set (#664). However that gets resolved, it is four open PRs plus this one editing the same lines.

Replacing the `powershell.exe` call in `getWindowsProcessStartId` with a native query is a good change on its own, since that spawn ran on every lease check.

I have not reproduced the lease case. It needs a held handle on `owner.json` while a live session owns the lease.

kevinjosethomas:

PR #763, opened 2026-08-06
fix(coding-agent): rebuild the kernel venv while a kernel is running on Windows

FlamesONE:
## Problem

On Windows a kernel venv rebuild fails whenever a kernel started from that venv is still running:

```
Error: Failed to set up the Python kernel runtime.
EPERM: operation not permitted, unlink 'C:\Users\<user>\.prime\agent\kernel-venv\Scripts\python.exe'
```

`ensureKernelPythonUncached` removes the old venv with `rm(venv, { recursive: true, force: true })` before rebuilding. Windows keeps a running executable mapped and refuses to unlink it, so the whole bootstrap aborts and the agent is left without a kernel. Every session sharing that venv hits it, and the daemon keeps kernels alive across sessions, so it is easy to reach.

Measured on Windows 11 Pro 26200 with a live uv venv (`python.exe` running):

| operation | result |
| --- | --- |
| `unlink Scripts/python.exe` | EPERM |
| `rm -rf venv` | EPERM |
| `rename venv -> venv.stale` | OK |

## Fix

Rename the venv out of the way instead of deleting it in place. The rename frees the path for the new venv even while the old interpreter is mapped; the renamed directory is deleted right away when nothing holds it, and otherwise by the next rebuild, so leftovers cannot accumulate. Non-Windows behaviour is unchanged apart from going through the rename.

## Tests

New `test/kernel-venv-discard.test.ts`, four cases: idle venv removed, missing venv is a no-op, venv path freed while its interpreter runs, and leftovers from an earlier rebuild cleaned up. The running-interpreter case copies the current `node` binary into the fake venv and runs it, which reproduces the same mapped-image lock a live kernel holds. It fails with the original `EPERM` against the previous `rm` implementation, and it is meaningful on Unix too, where it asserts the path is freed and nothing is left behind.

`npm run check` passes. Verified on real Windows hardware (Windows 11 Pro 26200, Node 24.15.0) and the tests are platform-independent, so they also run on the Linux CI.

This is orthogonal to the Windows work in #744; it touches only the rebuild path in `bootstrap.ts` and does not depend on it.

---

> [!NOTE]
> **Medium Risk**
> Changes the kernel bootstrap rebuild path used for every agent session; behavior is well-tested but a failed rename fallback still uses recursive rm, which can still fail on Windows with a live interpreter.
> 
> **Overview**
> On Windows, rebuilding the Python **kernel venv** used to call `rm` on the whole directory. If a kernel from that venv was still running (common with the daemon), unlinking `python.exe` fails with **EPERM** and bootstrap aborts.
> 
> The rebuild path now uses **`discardVenvDir`**: rename the old venv to a `*.stale-<pid>-<time>` directory (which succeeds while the interpreter is mapped), then delete that staging dir when possible. **`removeStaleVenvDirs`** clears leftover stale folders from prior rebuilds so they do not accumulate. **`ensureKernelPythonUncached`** swaps `rm` for `discardVenvDir` when rebuilding.
> 
> Adds **`kernel-venv-discard.test.ts`** (idle removal, missing venv, path freed with a running stand-in interpreter, stale cleanup).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ab946de0a2b13840cff89f442bac4a22084c3b94. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Fix kernel venv rebuild on Windows when a kernel is still running
> On Windows, deleting a venv directory fails when the Python interpreter is still mapped by a running kernel. This fix introduces `discardVenvDir` in [bootstrap.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/763/files#diff-e87b9a04ab9b4ac02e8dd33af2f2f58320d8b7f6be6853acf4a3e05a34623cae), which renames the venv to a PID/timestamp-suffixed `.stale-` directory to free the original path, then attempts a deferred removal. A `removeStaleVenvDirs` helper cleans up any leftover staged directories from prior rebuild attempts before each discard. The kernel venv rebuild now calls `discardVenvDir` instead of a direct `rm -rf`.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized ab946de.</sup>
> 
>

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## [Unreleased]
 
+- Fixed the kernel venv rebuild failing on Windows when a kernel from the old venv was still running.
+
 ## [0.7.0] - 2026-08-05
 
 ### Breaking Changes
--- a/packages/coding-agent/src/core/kernel/bootstrap.ts
+++ b/packages/coding-agent/src/core/kernel/bootstrap.ts
@@ -1,7 +1,7 @@
 import { spawn } from "node:child_process";
 import { createHash } from "node:crypto";
 import { constants, existsSync, readdirSync, readFileSync } from "node:fs";
-import { access, mkdir, readdir, readFile, rm, stat, writeFile } from "node:fs/promises";
+import { access, mkdir, readdir, readFile, rename, rm, stat, writeFile } from "node:fs/promises";
 import os from "node:os";
 import path from "node:path";
 import { stderr, stdin } from "node:process";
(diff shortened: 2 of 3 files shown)


khanra17:
Confirmed this is the existing fix for the Windows venv rebuild failure reported in #660. On Windows 11, deleting a venv containing a mapped `Scripts\\python.exe` returns `EPERM`, while renaming the directory frees the original path for a replacement venv. No duplicate PR is needed.

When rebasing this onto the current Windows support work, stale renamed directories should remain best-effort cleanup while the old kernel is still running and be retried later rather than blocking the new kernel bootstrap.

kevinjosethomas:

Issue #764, opened 2026-08-06
kernel crash


mrairdon-midmark:
This tool is phenomenal.

I've gotten this a few times:

<img width="636" height="173" alt="Image" src="https://github.com/user-attachments/assets/c7140aea-638a-4afe-9e01-ea10b63dca02" />

The only way to get python responding again seems to be running `prime-agent shutdown` then reopening and resuming a session.

Any place I can hunt for logs on what's happening there?

kevinjosethomas:
Hi! Logs should be in ``~/.prime/agent/logs`` if you can take a look for any recurring errors

mrairdon-midmark:
I think both of these are on my end but dropping them here in case you're interested for ideas to increase the robustness of the kernel:

```
I found two distinct recent crashes:

1. Aug 5, 9:11 PM — OOM kill
   — An IPython command recursively called read_text() on every repository file, including a 5.7 GB .git/objects pack.
   — The Python kernel reached ~24.2 GB RSS, exhausted 8 GB swap, and Linux killed it.
   — Confirmed in journalctl: Out of memory: Killed process ... (python).

2. Aug 6, 1:01 PM — accidental process-group kill
   — The agent tried stopping a long-running test with:
       os.killpg(os.getpgid(237205), signal.SIGTERM)
   — That test inherited the agent worker’s process group, so this also terminated the worker and IPython kernel.
   — daemon.sock.ee92e1d8.log confirms SIGTERM, exit 143, two sessions closed, then automatic recovery.

The other shutdown entries are mostly normal idle eviction or clean exit-0 restarts, not crashes.

Prevention: exclude .git/build/binary files from content scans, and never killpg() unless the subprocess was launched with
start_new_session=True and its PGID differs from the agent’s own process group.
```

Feel free to close. Thanks for the quick response!

bilelrais:
I hit the same symptom on Prime Agent 0.7.1 and was able to isolate both the external trigger and the Prime-side failure mode. The external SIGTERM explains why the kernel died, but Prime then made the session permanently unrecoverable and an active goal/heartbeat turned that into a costly continuation loop.

### Incident evidence (sanitized)

- Linux, Prime Agent `0.7.1`, `openai-codex/gpt-5.6-sol`, daemon-backed root session.
- A parent agent launched 15 concurrent headless Prime Agent relay workers. The worker logs show 15 independent worker daemons starting within ~2 seconds, each with one active session. This caused severe memory pressure.
[...]

PR #770, opened 2026-08-06
fix(prime-agent-runtime): kill subprocess tree on TimeoutExpired to prevent Windows kernel hangs

podlinovvv:
## Problem

The IPython kernel can hang forever on Windows when an agent runs a command that spawns a long-lived grandchild process. The common trigger is `subprocess.run(["powershell", "Start-Process <gui-app>", ...], capture_output=True, timeout=N)`: `Start-Process` launches a GUI application that **inherits the stdout/stderr pipes** of the powershell process.

Observed in a real session on Windows 11: the kernel sat for 9 minutes in `subprocess.py:559`, and the only recovery was killing the session.

## Root cause

When the timeout fires, `subprocess.run` kills only the direct child, then -- on Windows -- calls `process.communicate()` a second time **without a timeout** to collect partial output:

```python
except TimeoutExpired as exc:
    process.kill()
    if _mswindows:
        exc.stdout, exc.stderr = process.communicate()  # waits for pipe EOF forever
    else:
        process.wait()
    raise
```

That second `communicate()` waits for EOF on the stdout/stderr pipes. The grandchild (the GUI app launched via `Start-Process`) is still alive and holds them open, so EOF never arrives and the kernel wedges permanently. `py-spy dump` confirmed the stack (`subprocess.py:559`) and a surviving grandchild whose powershell parent was already dead (`Get-CimInstance`).

On POSIX this exact hang cannot happen: the stdlib path is `process.kill()` + `process.wait()` with no second `communicate()`. The timeout still leaves grandchild processes behind, but the kernel itself does not wedge.

## Fix

Install a small wrapper around `subprocess.run` in the kernel runtime (`prime-agent-runtime/src/rlm/__init__.py`). On `TimeoutExpired` it kills the **entire process tree** before the final `wait()`:

- Windows: `taskkill /PID <pid> /T /F`
- POSIX: `os.killpg` -- only when the child's process group differs from the kernel's own group; otherwise falls back to `os.kill(pid, SIGKILL)` so a plain `subprocess.run` (which inherits the kernel's group) can never kill the kernel itself

The direct child is also `kill()`ed explicitly so the wait cannot hang even if `taskkill` is unavailable. `TimeoutExpired` is re-raised with standard semantics (partial output dropped, matching the POSIX stdlib behavior). Calls without `timeout=` are untouched, and the wrapper can be disabled with `PRIME_AGENT_DISABLE_SUBPROCESS_KILL_TREE=1`.

## Verification

- New `test/test_subprocess_kill_tree.py` (9 cases): patch installed and idempotent, env-flag opt-out, normal `run`/`check=True`/`input=` semantics preserved, and the hang scenario -- a python parent spawning a sleeping python grandchild that records its pid -- where `timeout=2` raises `TimeoutExpired` in ~2s **and the grandchild is confirmed dead on Windows**. The POSIX branch is covered with mocked `getpgid`/`killpg`/`kill` (same-group vs different-group).
- Full runtime suite: 74 tests pass (2 POSIX-only skips on Windows).
- Live reproduction on Windows 11: the original hang now raises `TimeoutExpired` in ~3s with no surviving grandchildren, instead of hanging forever.
- `npm run check` passes (Biome, TypeScript, installer render, browser smoke).

## Note

The bug is Windows-specific; on Linux/macOS behavior is unchanged except that a timed-out run now also cleans up grandchild processes when their process group differs from the kernel's.

Related: #748 (daemon-side livelock that can leave the host unresponsive to the kernel), #764 (kernel crash report with a related killpg footgun).

> [!NOTE]
> ### Kill subprocess tree on `TimeoutExpired` to prevent Windows kernel hangs
> - Adds `_subprocess_kill_tree` in [rlm/__init__.py](https://github.com/PrimeIntellect-ai/prime-agent/pull/770/files#diff-8f05bb028ac34c9ff421f45f913f35be2a99010ef924da344d7a427869c7ead3) to terminate a process and all descendants: uses `taskkill /T /F` on Windows and `os.killpg`/`os.kill(SIGKILL)` on POSIX.
> - Monkeypatches `subprocess.run` via `_install_subprocess_run_kill_tree` so that on `TimeoutExpired`, the full process tree is killed before re-raising, preventing orphaned grandchildren from holding pipes open.
> - The patch is applied at import time unless `PRIME_AGENT_DISABLE_SUBPROCESS_KILL_TREE=1` is set, and is idempotent.
> - Adds tests covering install behavior, normal `subprocess.run` semantics, tree-kill on timeout (including grandchild verification on Windows), and POSIX branch logic.
> - Risk: importing `rlm` globally replaces `subprocess.run` for the entire process; the original is preserved and non-timeout paths delegate unchanged.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 1cecd46.</sup>
> 
>

### DIFF
--- a/prime-agent-runtime/src/rlm/__init__.py
+++ b/prime-agent-runtime/src/rlm/__init__.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import asyncio
+import os
 import sys
 import types
 from dataclasses import dataclass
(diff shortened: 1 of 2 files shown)


kevinjosethomas:

PR #731, opened 2026-08-06
fix: codex model discovery sends prime-agent version as client_version

ayushnangia:
Fixes #702.

## Problem

`openAICodexModelsUrl()` (`packages/coding-agent/src/core/model-registry.ts:386`) passed prime-agent's own package `VERSION` as the `client_version` query parameter. The ChatGPT backend gates the Codex model catalog by Codex CLI version, so `0.7.0` reads as an ancient client and returns `{"models": []}`. The empty catalog filters every `openai-codex` model out of `getAvailable()`, `findRlmModels`, and subagent spawns, producing the misleading "unavailable, unauthenticated, or expired" error with fresh auth — exactly as diagnosed in the report.

## Change

Pin a verified Codex CLI version (`OPENAI_CODEX_CLIENT_VERSION = "0.146.1"`, confirmed by the reporter to restore the full 9-model catalog with the same bearer token) instead of our own version, with a comment explaining why the package version must not be used. `VERSION` import dropped from model-registry (now unused there).

Complementary to #717, which fail-opens when discovery legitimately returns empty; this PR makes discovery return the real catalog in the first place. The report's optional suggestion (distinct error for discovery-empty vs auth failure) is left out to keep the change minimal.

## Tests

`test/suite/regressions/702-codex-client-version.test.ts`: stubs fetch, drives `findRlmModels` through the harness, and asserts the `/codex/models` request's `client_version` is not the package `VERSION` and matches a Codex CLI version shape. Fails on main (`expected '0.7.0' not to be '0.7.0'`).

Neighbor suites (`4649-subagent-model-selection`, `model-registry`, `696-empty-codex-catalog`): 85/85. Note: two 4649 tests fail on unmodified main when `DEEPSEEK_API_KEY`/`OPENROUTER_API_KEY`/`HF_TOKEN` are set in the environment (provider availability leaks into discovery); all green with those unset, before and after this change. `npm run check` clean.

---

> [!NOTE]
> **Low Risk**
> Narrow change to Codex catalog discovery query params plus a regression test; no auth or data-path changes.
> 
> **Overview**
> Fixes **#702** by stopping `openAICodexModelsUrl` from sending prime-agent’s package `VERSION` as the `client_version` query param on `/codex/models` requests. The ChatGPT backend treats that as an outdated Codex CLI client and returns an empty catalog, which made every `openai-codex` model disappear from discovery, `getAvailable()`, and subagent flows despite valid auth.
> 
[...]

PR #724, opened 2026-08-06
fix: rank configured providers first in model selector search

ayushnangia:
Fixes #653.

## Problem

The unfiltered model list sorts signed-in providers first (`sortModels`), and the selector's hint says "Signed-in providers first." — but the search path sorts by raw fuzzy score before configured-provider status (`filterModels`). A query whose text matches an unconfigured provider's model more tightly surfaces junk the user cannot run above the provider they are logged into, which is the screenshot in #653.

## Change

`packages/coding-agent/src/modes/interactive/components/model-selector.ts`: flip the search tiebreak to configured provider, then match score, then recency — making search consistent with browse and with the component's own hint. One-line reorder.

The second observation in the issue (signed-in state not visible per row) is a display addition, not touched here.

## Tests

`test/suite/regressions/653-search-configured-provider-priority.test.ts`: an unconfigured provider's `k2` scores strictly better than the configured `kimi-k2-thinking` for the query `k2` (score probe: -24.9 vs -4.4), so score-first ordering puts the unrunnable model on top — fails on main, passes with the fix.

Neighbor suites (`model-selector-actions`, `selector-search-reset`, `3217-scoped-model-order`): 14/14. `npm run check` clean.

---

> [!NOTE]
> **Low Risk**
> Small comparator reorder in interactive UI only; behavior change is intentional and covered by a regression test.
> 
> **Overview**
> Fixes **#653** by aligning model search ranking with the unfiltered list and the “Signed-in providers first” hint.
> 
> In `filterModels`, search results are now sorted by **configured/signed-in provider** before fuzzy match score, then recency. Previously, a tighter text match on an unconfigured provider could appear above a runnable model from a provider the user is logged into.
> 
> Adds regression test `653-search-configured-provider-priority.test.ts` where query `k2` would rank `kimi/k2` above the configured `kimi-k2-thinking` under score-first ordering.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f26bec8e755afbf0af1d2cf79a0ac9eac2426793. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Rank configured providers first in model selector search results
[...]

PR #773, opened 2026-08-06
fix: npm 10+ compatibility across the update and install surfaces

### LINKED ISSUE #738: `/update --extensions` always fails with "Update exited with code 1"; update-available notice is a false positive on npm >= 11
**Repo:** https://github.com/PrimeIntellect-ai/prime-agent
**Type:** Bug
**Affected version:** v0.7.0 (current latest stable per `latest.json`)
**Severity:** Medium — feature advertised by the UI is broken; startup notice nags with false positives

---

## Summary

Two related defects in the package-update flow:

1. **The startup notice and the `/update` slash command still advertise `/update --extensions`, but the CLI `update` command now rejects that flag.** Running `/update --extensions` in the TUI spawns `prime-agent update --extensions`, which exits 1 with `Error: Package updates moved to the package command. Use "prime-agent package update [source]".` The TUI then reports `Update exited with code 1`. The legacy `--extensions`/`--extension`/positional-source handling was removed from the CLI but left behind in the slash command (`argumentHint: "[source|--self|--extensions]"`), the arg parser (`updateArgsIncludeSelf`), and the startup notice.

2. **`getLatestNpmVersion()` breaks on npm ≥ 11: `npm view <pkg> version --json` now emits a JSON *array*, not a bare string.** The parsed value (`["2.87.2"]`) is compared with `!==` against the installed string (`"2.87.2"`), so the check is *always* true. Result: every startup shows "Package updates available: …" even when packages are already at the latest version — which then funnels the user into defect #1.

## Environment

- prime-agent v0.7.0 (installed at `/opt/homebrew/lib/node_modules/prime-agent`)
- macOS (darwin arm64)
- Node v26.6.0, npm 12.0.2
- Configured packages (`~/.prime/agent/settings.json`): `npm:bigpowers` (2.87.2 installed = latest), `npm:pi-hashline-edit-pro` (1.1.0 installed = latest), plus a git package

### Defect 1 — `/update --extensions` always fails

1. Configure at least one npm capability package: `prime-agent package install npm:bigpowers` (or any package).
2. Start the interactive TUI and run:

   ```
   /update --extensions
   ```

   Result in the TUI:

   ```
   Error: Update exited with code 1
   ```

   The underlying command, reproduced from a shell:

   ```
   $ prime-agent update --extensions
   Error: Package updates moved to the package command.
   Use "prime-agent package update [source]".
   $ echo $?
   1
   ```

   Expected: either the packages get updated (legacy behavior) or the command is not advertised anywhere.

### Defect 2 — false "Package updates available" notice

1. Ensure all npm packages are already at latest (verified: bigpowers 2.87.2, pi-hashline-edit-pro 1.1.0 = `dist-tags.latest`).
2. Start prime-agent. The agents view still shows:

   ```
   Package updates available: bigpowers, pi-hashline-edit-pro. Run /update --extensions
   ```

   Expected: no notice when everything is up to date.

### Defect 1 (broken legacy path)

- `dist/cli/public-command.js:96-102` — the CLI `update` command explicitly fails for any legacy package target:
  ```js
  const hasLegacyPackageTarget = rest.some((arg) => arg === "--extensions" || arg === "--extension" || (!arg.startsWith("-") && !isSelfUpdateSource(arg)));
  if (hasLegacyPackageTarget) {
      return fail("Package updates moved to the package command.", `Use "${APP_NAME} package update [source]".`);
  }
  ```
- `dist/modes/interactive/interactive-mode.js:7115-7205` — `handleUpdateCommand()` spawns the CLI child with the user's raw args. For `--extensions`, `updateArgsIncludeSelf()` (line 412-443) explicitly returns `false`, so `updateChildArgs = ["--extensions"]` is passed through unchanged (line 7125), the child exits 1, and the TUI prints `Update exited with code 1` (line 7205). There is no branch that routes `--extensions` to the package manager.
- `dist/core/slash-commands.js:77-81` — `/update` still advertises `argumentHint: "[source|--self|--extensions]"`.
- `dist/modes/shared/startup-notices.js:86-87` — the notice tells the user to run the now-broken `/update --extensions`.

### Defect 2 (npm ≥ 11 `--json` output shape)

- `dist/core/package-manager.js:1165-1174`:
  ```js
  async getLatestNpmVersion(packageName) {
      const stdout = await this.runCommandCapture(npmCommand.command, [...npmCommand.args, "view", packageName, "version", "--json"], ...);
      const raw = stdout.trim();
      if (!raw) throw new Error("Empty response from npm view");
      return JSON.parse(raw);
  }
  ```
- `npm view bigpowers version --json` on npm 12.0.2 prints `["2.87.2"]` (an array), not `"2.87.2"`. `JSON.parse` therefore returns `["2.87.2"]`.
- `npmHasAvailableUpdate()` (line 1136) and `shouldUpdateNpmSource()` (line 856) compare `latestVersion !== installedVersion`, i.e. `["2.87.2"] !== "2.87.2"` → always `true`.
- Note: plain `npm view bigpowers version` (without `--json`) still prints the bare string `2.87.2`, so the fix can be minimal.

## Suggested fix

1. **Route or remove the legacy TUI path (defect 1):**
   - Either make `/update --extensions` (and positional package sources) delegate to the package manager — e.g. spawn `prime-agent package update` — or reject the flag in the TUI with the same "moved to the package command" guidance instead of a bare exit-code error.
   - Update `dist/modes/shared/startup-notices.js` to recommend the working command (e.g. `Run "prime-agent package update"` or a working slash form) instead of `/update --extensions`.
   - Update the `/update` `argumentHint` in `dist/core/slash-commands.js` to match what the CLI actually supports (`[--self]`/`[--force]`), and drop the dead `--extensions`/`--extension` handling in `updateArgsIncludeSelf`.
   - Consider surfacing the child's stderr/error message in the TUI (`Update exited with code 1` gives no actionable info; the real error was printed only because the child inherits stdio).

2. **Normalize the npm version lookup (defect 2):** in `getLatestNpmVersion()`, accept both the npm ≤ 10 string output and the npm ≥ 11 array output, e.g. parse and take `Array.isArray(parsed) ? parsed[0] : parsed`, or drop `--json` and trim quotes. This also stops the perpetual "updates available" notice and avoids needless `npm install` runs on every `/update`.

## Workaround (for affected users)

```
prime-agent package update        # updates capability packages (still works)
prime-agent update                # updates prime-agent itself (still works)
```

The extension packages in this report were already at latest; the "Package updates available" notice is a false positive caused by defect 2.

### LINKED ISSUE #741: install fails on npm 12+ due to allow-remote=none
As of npm 12, `allow-remote` defaults to `none`. [0] [1] Not sure of the proper solution here, but at least warn the user that you will be doing `npm install --allow-remote` and maybe link to some documentation explaining the risks?

[0]: https://docs.npmjs.com/cli/v12/using-npm/config#allow-remote
[1]: https://github.blog/changelog/2026-06-09-upcoming-breaking-changes-for-npm-v12/

### LINKED ISSUE #749: Installer gives unsupported `npm bin -g` PATH instructions on npm 10+
## Summary

The installer currently tells users to run `npm bin -g` when `prime-agent` was installed but is not on `PATH`. npm 10 no longer supports the `bin` command, so following the printed recovery instructions fails with `Unknown command: "bin"`.

Current installer guidance:
https://github.com/PrimeIntellect-ai/prime-agent/blob/main/install.sh#L135-L140

## Reproduction

Environment:

- macOS
- prime-agent 0.7.0
- npm 10.9.8
- Node v22.22.3
- npm global prefix: `~/.hermes/node`
- installed executable: `~/.hermes/node/bin/prime-agent`

After installation, run the suggested command:

```console
$ npm bin -g
Unknown command: "bin"

To see a list of supported npm commands, run:
  npm help
```

## Expected behavior

The installer should print a recovery command that works with supported npm versions, or preferably detect and print the exact executable directory itself.

On macOS/Linux, the global executable directory can be derived from:

```sh
printf "%s/bin\n" "$(npm prefix -g)"
```

For this installation that resolves to `~/.hermes/node/bin`. The implementation should account for platform differences if Windows is supported by this path.

ayushnangia:
Fixes #738. Fixes #749. Fixes #741.

## Problem

Three faces of one drift: npm's CLI moved (10, 11) and prime-agent's own update-command migration left the TUI behind.

1. **`/update --extensions` always fails** (#738): the standalone `update` command deliberately hard-rejects package targets ("Package updates moved to the package command", `public-command.ts:129`) — but the TUI still spawns `update --extensions` for the flow it advertises in the slash hint and the startup notice. Every advertised path exits 1.
2. **Update notice is a permanent false positive on npm ≥ 11** (#738): `npm view <pkg> version --json` now emits a JSON *array*; `getLatestNpmVersion` parsed it verbatim, and `["2.87.2"] !== "2.87.2"` is always true — so every startup nags, funneling users into defect 1.
3. **PATH recovery instructions use a removed command** (#749): `install.sh` tells users to run `npm bin -g`, which npm 10 removed.

## Change

- `interactive-mode.ts`: non-self updates now spawn `package update` with legacy flags translated (`translatePackageUpdateArgs`: `--extensions` → bare, `--extension <src>` → positional, daemon-socket plumbing dropped). Self-updates unchanged. The advertised `/update --extensions` flow now runs the command the CLI migration intended.
- `package-manager.ts`: `normalizeNpmViewVersion` handles the npm 10 bare-string and npm ≥ 11 array forms, takes the newest entry, and fails loudly on malformed output.
- `install.sh`: PATH guidance prints `echo "$(npm prefix -g)/bin"` instead of `npm bin -g`.

## Tests

`test/suite/regressions/738-npm-update-compat.test.ts` (9 tests, fail on main — the helpers don't exist): all translation shapes, agreement with `updateArgsIncludeSelf` about what routes to the package path, both npm output forms, newest-entry selection, loud rejection of empty/malformed output, and the exact false-positive equality from the report.

Neighbor suites (`interactive-mode-effort-command`, `public-command`): 47/47. `npm run check` clean, including the installer render gate.

## Scope notes

- #741 is now included (00d8d19): @d4not verified the mechanism on npm 12.0.2 — `EALLOWREMOTE` from *transitive* URL dependencies (R2-hosted runtime tarballs), `root` insufficient. The installer scopes `npm_config_allow_remote=all` to its own `npm install -g` invocations via env (no `~/.npmrc` changes), gated to npm ≥ 12 so older versions see no unknown-config warnings. Scope notes: the npm self-update step (`config.ts` `makeSelfUpdateCommandStep`) has the same exposure on npm 12; and the deeper fix — publishing without URL transitive deps — is a release-pipeline call, so both are left for maintainers rather than decided here.
- The slash `argumentHint` and startup-notice wording still advertise `--extensions`; they now work as advertised, so the wording is untouched.

> [!NOTE]
> ### Fix npm 10+ compatibility in package install and update flows
> - Replaces deprecated `npm bin -g` with `npm prefix -g` in [install.sh](https://github.com/PrimeIntellect-ai/prime-agent/pull/773/files#diff-043df5bdbf6639d7a77e1d44c5226fd7371e5259a1e4df3a0dd5d64c30dca44f) and adds `npm_config_allow_remote=all` env prefix for npm >= 12 to avoid `EALLOWREMOTE` errors from transitive URL dependencies.
> - Adds `normalizeNpmViewVersion` in [package-manager.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/773/files#diff-c3cd658663a2617a59357bae55e941f8d7885378fdd6e4cd0400dfa967fbe750) to handle the changed output format of `npm view <pkg> version --json` between npm 10 (string) and npm 11+ (array).
> - Rewrites the TUI non-self update path in [interactive-mode.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/773/files#diff-b96137f89d422953665b334785273ae6cdbcb53ff37d8828db43f85038658316) to translate legacy `/update` flags into the `package update` subcommand syntax, fixing package updates that broke with newer npm CLI argument handling.
> - Adds regression tests in [738-npm-update-compat.test.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/773/files#diff-17d39b36db7181d62081bdf83ac839f3dc5a3e13516c49bce712a7449f764005) covering flag translation, version normalization, and error handling.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 00d8d19.</sup>
> 
>

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -143,6 +143,24 @@ EOF
 	fi
 }
 
+# npm 12 defaults allow-remote=none, and the published prime-agent tarball
+# carries transitive URL dependencies (R2-hosted @earendil-works tarballs),
+# so a plain "npm install -g" exits with EALLOWREMOTE. "root" is not enough —
+# the blocked tarballs are transitive — so the install invocation is scoped
+# to allow-remote=all via env, without touching the user's ~/.npmrc. (#741,
+# verified by @d4not on npm 12.0.2.) Gated to npm >= 12 so older npm versions
+# see no unknown-config warnings.
+npm_allow_remote_env() {
+	case "$(npm --version 2>/dev/null)" in
+	1[2-9].* | [2-9][0-9].*)
+		printf 'npm_config_allow_remote=all'
+		;;
+	*)
+		printf ''
+		;;
+	esac
+}
+
 create_temp_dir() {
 	if command -v mktemp >/dev/null 2>&1; then
 		if tmp_dir=$(mktemp -d "${TMPDIR:-/tmp}/prime-agent-install.XXXXXX" 2>/dev/null); then
@@ -1603,7 +1621,7 @@ Finalizing npm install."
 			"Installing Prime Agent" \
 			"Installing Prime Agent" \
 			"$npm_install_details" \
-			env PRIME_AGENT_BOOTSTRAP_TOOLS_ON_INSTALL=1 PRIME_AGENT_BOOTSTRAP_KERNEL_ON_INSTALL=1 PRIME_AGENT_INSTALL_UV=1 npm install -g --no-fund --no-audit --loglevel=error --progress=false "$tarball_path"
+			env $(npm_allow_remote_env) PRIME_AGENT_BOOTSTRAP_TOOLS_ON_INSTALL=1 PRIME_AGENT_BOOTSTRAP_KERNEL_ON_INSTALL=1 PRIME_AGENT_INSTALL_UV=1 npm install -g --no-fund --no-audit --loglevel=error --progress=false "$tarball_path"
 	else
 		npm_install_details="Preparing global install.
 Linking command binaries.
@@ -1614,7 +1632,7 @@ Finalizing npm install."
 			"Installing Prime Agent" \
 			"Installing Prime Agent" \
 			"$npm_install_details" \
-			env PRIME_AGENT_BOOTSTRAP_TOOLS_ON_INSTALL=1 npm install -g --no-fund --no-audit --loglevel=error --progress=false "$tarball_path"
+			env $(npm_allow_remote_env) PRIME_AGENT_BOOTSTRAP_TOOLS_ON_INSTALL=1 npm install -g --no-fund --no-audit --loglevel=error --progress=false "$tarball_path"
 	fi
 }
 
--- a/install.sh
+++ b/install.sh
@@ -135,9 +135,10 @@ main() {
 The $prime_agent_cmd command was installed, but it is not on your PATH yet.
 Check npm's global bin directory with:
 
-  npm bin -g
+  echo "\$(npm prefix -g)/bin"
 
 Then add that directory to your shell PATH.
+(npm 10 removed the old "npm bin -g" command.)
 EOF
 	fi
 }
--- a/packages/coding-agent/src/core/package-manager.ts
+++ b/packages/coding-agent/src/core/package-manager.ts
@@ -36,6 +36,24 @@ import { isStdoutTakenOver } from "./output-guard.js";
 import type { PackageSource, SettingsManager } from "./settings-manager.js";
 
 const NETWORK_TIMEOUT_MS = 10000;
+
+/**
+ * Parse `npm view <pkg> version --json` output. npm 10 emits a bare JSON
+ * string; npm >= 11 emits a JSON array (e.g. `["2.87.2"]`). Comparing the
+ * un-normalized array against the installed version string made every
+ * startup report "Package updates available" as a false positive (#738).
+ * Exported for testing.
+ */
+export function normalizeNpmViewVersion(stdout: string): string {
+	const raw = stdout.trim();
+	if (!raw) throw new Error("Empty response from npm view");
+	const parsed: unknown = JSON.parse(raw);
+	const version = Array.isArray(parsed) ? parsed[parsed.length - 1] : parsed;
+	if (typeof version !== "string" || version.length === 0) {
+		throw new Error(`Unexpected npm view version output: ${raw}`);
+	}
+	return version;
+}
 const UPDATE_CHECK_CONCURRENCY = 4;
 const GIT_UPDATE_CONCURRENCY = 4;
 
--- a/packages/coding-agent/src/modes/interactive/interactive-mode.ts
+++ b/packages/coding-agent/src/modes/interactive/interactive-mode.ts
@@ -671,6 +671,37 @@ function getPayloadWorkingIndicatorOptions(
 	};
 }
 
+/**
+ * Translate legacy `/update` package arguments into the `package update`
+ * CLI surface. The standalone `update` command hard-rejects package targets
+ * ("Package updates moved to the package command"), but the TUI kept
+ * spawning `update --extensions`, so the advertised flow always exited 1
+ * (#738). `--extensions` means "all packages" (the default), `--extension
+ * <source>` and positional sources name one package, and daemon-socket
+ * plumbing belongs only to self-updates.
+ */
+export function translatePackageUpdateArgs(args: readonly string[]): string[] {
+	const translated: string[] = [];
+	for (let index = 0; index < args.length; index++) {
+		const arg = args[index];
+		if (arg === "--extensions" || arg === "--self") {
+			continue;
+		}
+		if (arg === "--extension" || arg === "--daemon-socket") {
+			const value = args[index + 1];
+			index++;
+			if (arg === "--extension" && value !== undefined) {
+				translated.push(value);
+			}
+			continue;
+		}
+		if (arg !== undefined) {
+			translated.push(arg);
+		}
+	}
+	return translated;
+}
+
 export function updateArgsIncludeSelf(args: readonly string[]): boolean {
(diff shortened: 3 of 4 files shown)


ayushnangia:
Noting this for whichever maintainer picks the npm-compat area: given the clean-room approach you used on #717 → #1153 and #731 → #1164, I assume these remaining fix PRs of mine (this one, #721, #722, #723, #724, #730) are also best treated as bug reports rather than import candidates — which is completely fine. Each contains a full diagnosis, a minimal repro, and regression tests that stand alone: the npm-12 EALLOWREMOTE semantics in this one were independently verified by @d4not (comment history). Happy to leave them open as living reports, or convert each into a proper issue if that fits your triage flow better — just say which you'd prefer. The root causes matter more than the diffs.

kevinjosethomas:

ayushnangia:
Makes sense — a local agent that runs on users' machines needs the tighter intake, and discussions-first is the right shape for it. The six reports this account left behind (#721, #722, #723, #724, #730, #773) each contain a full diagnosis and standalone repro/regression tests; if any of them correspond to work you'd want pursued, happy to file them as Discussions with the diagnostics carried over — or just let them stand as reference material while you triage the backlog. Thanks for the deliberate process.

PR #775, opened 2026-08-06
feat(ai): add OpenRouter OAuth provider

andrew-scott-fischer:
Adds a PKCE login flow for OpenRouter so `/login` can authenticate with a browser sign-in instead of requiring `OPENROUTER_API_KEY`.

- New `openrouterOAuthProvider` in `packages/ai/src/utils/oauth/openrouter.ts`, registered in `BUILT_IN_OAUTH_PROVIDERS`
- OpenRouter shows up as a subscription entry in `/login` alongside its existing API-key entry, the same way `anthropic` already does
- The `OPENROUTER_API_KEY` environment variable and `auth.json` API-key entries are unchanged

Implementation notes:

- OpenRouter's authorize endpoint has no `state` parameter, so CSRF protection comes from a random UUID callback path; only a request to that exact path is accepted
- The callback host is `localhost` rather than the `127.0.0.1` the other providers default to. OpenRouter's authorize page does not proceed with a `127.0.0.1` callback_url
- The callback server binds an ephemeral port, and `callbacks.signal` is forwarded to both the server and the token exchange
- The exchange returns an API key rather than an access/refresh pair and there is no refresh endpoint, so `refreshToken()` throws to route the user back to `/login`
- The exchange response carries no expiry, so `expires` is stored far in the future. A key that expires or is revoked server-side therefore surfaces as a provider 401 rather than the "run /login" auth-failed message

Verified against the live service: browser sign-in, an inference call, and credential reuse across a restart.

> [!NOTE]
> ### Add OpenRouter OAuth provider with PKCE browser login flow
> - Adds `openrouterOAuthProvider` to the built-in OAuth providers list, enabling OpenRouter login via browser sign-in instead of a manual API key.
> - Implements a full PKCE-based login flow in [openrouter.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/775/files#diff-48645384163946682f69df5de80242f482ad7162e08a8a122f9a7411b75e441a): starts a transient loopback callback server, opens a browser to the OpenRouter authorize URL, and exchanges the authorization code for an API key.
> - Supports manual code entry as a fallback when browser callback is unavailable; accepts a full redirect URL, query string, or bare code.
> - Returned credentials use `Number.MAX_SAFE_INTEGER` as the expiry (keys do not expire), and `refreshOpenRouterToken` always throws since OpenRouter keys cannot be refreshed.
> - Behavioral Change: any code attempting to refresh OpenRouter credentials will receive an explicit error rather than a silent failure.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 6bf267f.</sup>
> 
>

### DIFF
--- a/packages/ai/CHANGELOG.md
+++ b/packages/ai/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## [Unreleased]
 
+- Added an OpenRouter OAuth provider so OpenRouter can be authenticated with a browser sign-in instead of an API key ([#775](https://github.com/PrimeIntellect-ai/prime-agent/pull/775) by [@andrew-scott-fischer](https://github.com/andrew-scott-fischer)).
+
 ## [0.7.1] - 2026-08-07
 
 ## [0.7.0] - 2026-08-05
--- a/packages/ai/src/utils/oauth/index.ts
+++ b/packages/ai/src/utils/oauth/index.ts
@@ -5,6 +5,7 @@
  * for OAuth-based providers:
  * - Anthropic (Claude Pro/Max)
  * - GitHub Copilot
+ * - OpenRouter
  */
 
 // Anthropic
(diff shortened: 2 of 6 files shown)


kevinjosethomas:

PR #732, opened 2026-08-06
Add native Windows installation and launchers

PierrunoYT:
## Summary

- add native PowerShell and Command Prompt launchers for source checkouts
- add checksum-verified stable and beta PowerShell installers
- publish Windows installers through the existing release workflow
- document the native Windows setup and make installer validation locate Git Bash on Windows

The application, daemon, and Python runtime run natively on Windows. The model-facing command shell continues to use Bash for command compatibility and resolves Git Bash automatically.

## Validation

- `npm run check`
- `prime-agent.cmd --version`
- parsed both PowerShell scripts with the PowerShell AST parser
- exercised installer download, SHA-256 verification, and npm invocation against a local fake release

---

> [!NOTE]
> **Medium Risk**
> Changes first-time install and Python kernel bootstrap on Windows (download, checksum, global npm, uv install); release workflow now publishes additional install artifacts, but scope is platform/install plumbing rather than auth or data handling.
> 
> **Overview**
> **Native Windows install and dev launch** — New `install.ps1` mirrors the shell installer: resolve stable/beta, download the release tarball, verify SHA-256, and `npm install -g` with kernel/tool/uv bootstrap env vars enabled. `prime-agent.cmd` / `prime-agent.ps1` run from a source checkout (tsx or `--dist`), including `--no-env` stripping of API keys.
> 
> **Release and docs** — `build-binaries.yml` token-replaces and uploads `install.ps1` / `install-beta.ps1` to R2 with the existing shell installers. README, `windows.md`, and `development.md` document `irm … | iex` and `prime-agent.cmd`.
> 
> **Runtime fixes on Windows** — Kernel bootstrap uses `venv\Scripts\python.exe` via `getKernelVenvPython`, and optional uv install runs through PowerShell (`irm …/uv/install.ps1 | iex`) via `createUvInstallLaunchSpec`. `formatCurrentCliCommand` emits PowerShell `&` invocation with Windows-safe quoting when `platform === "win32"`. Installer render checks pick Git Bash `sh.exe` on Windows hosts.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f0df3a7139299d403df1af852d7775949f5f81be. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Add native Windows PowerShell installer and source launchers for Prime Agent
[...]

PR #716, opened 2026-08-06
feat(coding-agent): add Windows PowerShell installer

Villoh:
## Summary

Adds an official PowerShell installer for Windows with the same verified stable/beta release flow and terminal experience as `install.sh`.

Windows users can install Prime Agent with:

```powershell
irm https://app.primeintellect.ai/prime-agent/install.ps1 | iex
```

## Change

- Add `install.ps1` with Node.js/npm and bash preflight checks.
- Offer Node.js LTS and Git for Windows installation through `winget` when required.
- Resolve stable, beta, or explicit versions and verify downloaded release tarballs with SHA-256.
- Port the responsive Prime Agent splash, progress animation, prompts, compact layout, and terminal restoration from `install.sh`.
- Publish stable and beta PowerShell installers through the existing R2 release workflow.
- Document Windows installation and add an unreleased changelog entry.

The existing shell installer and macOS/Linux release flow are unchanged.

## Regression coverage

The installer check runs the real PowerShell functions with controlled HTTP responses and fake Node/npm executables. It verifies:

- PowerShell 5.1 and PowerShell 7 compatibility.
- Stable and beta installer rendering.
- Exact visual frame parity with `install.sh`.
- Responsive and compact terminal layouts.
- SHA-256 verification and npm installation arguments.
- Bash discovery through existing Prime Agent settings.
- Plain output when terminal control sequences are unavailable.

## Validation

- `npm run check`
- `npm run check:installer`
- `Invoke-ScriptAnalyzer -Path ./install.ps1 -Severity Warning,Error`

---

> [!NOTE]
> **Medium Risk**
> Touches the production release publish path and runs privileged global installs (npm/winget) on user machines, though behavior is isolated to Windows and covered by installer regression tests.
> 
> **Overview**
> Adds a **Windows PowerShell installer** (`install.ps1`) that mirrors the verified **stable/beta** release flow from `install.sh`: channel resolution, tarball download, **SHA-256** checks, and **global npm install**, with optional **Node.js LTS**, **Git for Windows**, and IPython runtime setup via **winget**.
> 
> The **release workflow** now renders and uploads `install.ps1` / `install-beta.ps1` to R2 next to the shell installers. **CI** gains `scripts/check-windows-installer.mjs` (wired into `check:installer`) for render parity with `install.sh`, layout checks, and a mocked end-to-end install path. **README** and coding-agent docs document the PowerShell one-liners and Windows bash requirements.
> 
[...]

PR #687, opened 2026-08-06
fix(coding-agent): use portable worker socket paths

omnomburp:
Worker sockets were generated under $TMPDIR, which is long by default on macOS. Combined with the per-user directory, supervisor hash, worker ID, and socket filename, paths could exceed macOS’s 103-byte Unix socket limit. The worker then failed to bind, while the client only surfaced a 30-second daemon create
 timeout. Linux usually avoids this because $TMPDIR is /tmp, but custom long temporary paths could cause the same issue.

---

> [!NOTE]
> **Medium Risk**
> Changes where workers bind and how recovery picks socket paths; misclassification could affect shutdown/discovery, but scope is limited to daemon IPC paths with tests covering limits and classification.
> 
> **Overview**
> Daemon **worker** Unix sockets are no longer created under `$TMPDIR` (which can be very long on macOS). They now use a fixed short per-user directory at `/tmp/prime-agent-<uid>`, while the main supervisor socket stays on `$TMPDIR`.
> 
> `workerSocketPath` enforces a **100-byte** path limit and errors if generation would exceed it. Socket dir setup was generalized so both the legacy supervisor dir and the new worker dir get the same ownership/mode checks. On Unix recovery, existing workers are **not** pinned to an old long path (Windows still reuses the stored path).
> 
> `daemon ps` treats worker sockets in **both** the new worker dir and the legacy `$TMPDIR` dir as workers, with macOS `/private` normalization so discovery/shutdown still filters them correctly.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 49e8cc94258d55d24552e0220fd6f5d2b7f36e40. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Fix worker socket paths in the coding agent daemon to be portable across platforms
> - Introduces `defaultDaemonWorkerSocketDir` in [daemon-socket.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/687/files#diff-59c4ea71700afcf7d09d14c5d30e2c73fd0c71b28284e8758cd8fbd4b2337dbc), placing Unix worker sockets in a per-user `/tmp/prime-agent-<uid>` directory instead of the shared daemon socket directory.
[...]

Issue #669, opened 2026-08-06
macOS worker socket path can exceed the Unix socket limit


omaclaren:
`prime-agent` times out while creating a session on macOS:

```text
Timed out after 30000ms waiting for the Prime Agent daemon response to "create"
```

The daemon log shows the worker failing with:

```text
Error: ENOENT: no such file or directory, lstat '.../worker-<id>.sock'
```

The generated worker socket path is 109 bytes. macOS allows 103 usable bytes for a Unix socket path. My numeric UID has 10 digits, which pushes the path over the limit.

Environment:

- Prime Agent 0.7.0
- Node 22.22.3
- macOS 14.8.7 arm64

Using a shorter temporary directory works:

```bash
TMPDIR=/tmp prime-agent
```

Prime Agent should probably use a shorter socket path or detect when the generated path exceeds the platform limit.

kevinjosethomas:
Hi, thanks for taking the time to report this! Since open sourcing Prime Agent, we’ve received far more Issues than we can reliably triage and maintain. Rather than leave a large backlog that we cannot meaningfully work through, we’re closing the current Issue queue and moving to a discussion-first process.

We have established new contribution guidelines to help us continue iterating on Prime Agent and better manage contributions from the community. Going forward, GitHub Issues will be our maintainer work queue rather than the initial place for public reports and requests. Bug reports, feature requests, questions, and other feedback should begin in [GitHub Discussions](https://github.com/PrimeIntellect-ai/prime-agent/discussions). Please read the full process documented in our [contribution guidelines](https://github.com/PrimeIntellect-ai/prime-agent/blob/main/CONTRIBUTING.md). Security vulnerabilities should continue to be reported privately according to our [security policy](https://github.com/PrimeIntellect-ai/prime-agent/blob/main/SECURITY.md).

While we’re closing this backlog, we’re still reviewing it at a high level to identify recurring bugs, useful ideas, and important problems that we should address ourselves. Thanks again for the time you put into this!

PR #781, opened 2026-08-06
ci: run the full suite on macOS, and matrix all three platforms together

noahsabaj:
## Why

macOS is a supported platform with no CI at all, and Linux and Windows coverage were two
hand-maintained jobs that had already drifted apart: Windows never ran the `ai`, process-smoke,
or kernel suites, and nothing reported the gap.

Adding macOS surfaced a real bug immediately — the one reported in #669 — which is included
here so the new legs go green rather than red.

## What changes

**One matrix over `(os × suite)`.** A suite added to the list now runs on every platform by
construction rather than by remembering to copy it into a second job. Per-suite settings come
from `include`; only the steps that genuinely differ — system packages and uv — branch on
`runner.os`.

- **macOS** takes uv from Homebrew. `pip install --user` puts its scripts under
  `~/Library/Python/<version>/bin`, not `~/.local/bin`, so the Linux step could not be reused.
- **Windows** asks `sysconfig` for the user scripts directory, which carries the interpreter
  version (`%APPDATA%\Python\Python313\Scripts`). The fixed `%APPDATA%\Python\Scripts` it used
  before never held uv, so every Windows job needing the Python kernel failed at bootstrap.
- **Windows legs run `continue-on-error`.** They report their real state without gating the
  merge, because the native Windows port is still landing in other PRs (#663, #664, #670, #695,
  #716, #722, #727, #732, #744, #748). Delete that one line once it has, and Windows gates like
  the other two.

The nightly process-stress workflow gets the same three-platform matrix, since it exercises the
process-teardown paths that differ most across platforms.

## The macOS fix (`fix(macos): keep daemon socket paths inside sun_path`)

`sockaddr_un.sun_path` is 104 bytes on macOS and the BSDs, and a stock macOS `$TMPDIR` already
spends 49 of them. Measured against the longest name the daemon creates:

| platform | uid | bytes | limit | result |
|---|---|---|---|---|
| linux | 1000 | 59 | 108 | unchanged |
| linux | 4294967294 (max) | 65 | 108 | unchanged |
| macOS | 501 (default) | 102 | 104 | unchanged — two bytes spare |
| macOS | 501234 | 105 | 104 | **overruns** |
| macOS | #669's reported path | 142 | 104 | **overruns** |

The overrun does not announce itself as "name too long": the name is truncated, so the server
binds one path while clients dial another, and it surfaces 30 seconds later as a daemon
`create` timeout.

The fallback engages **only** when the longest socket name would not otherwise fit, so Linux
and the default macOS layout keep byte-identical paths — on Linux it is unreachable at any
possible uid. The fallback directory name includes a hash of the preferred one, so processes
sharing a `$TMPDIR` agree on it while separate `$TMPDIR`s stay separate, which keeps an
isolated temporary directory isolating.

**Overlaps with #722 and #687**, which fix the same report. This variant is here so the macOS
legs added alongside it go green; happy to drop it in favour of either if one lands first. The
difference worth keeping is the hash — a flat `/tmp/prime-agent-<uid>` collapses two isolated
`TMPDIR`s into one directory.

## Verification

- Linux `Build and check` and the Linux suites pass unchanged.
- macOS legs pass with the `sun_path` fix; without it, `coding-agent 1/3` fails on worker socket
  binding.
- Windows legs report their real state; they are advisory until the port lands.

> Note: full three-platform CI verification was still pending at the time of writing because of
> the GitHub Actions incident on 2026-08-06. Re-run before merging.

## Note for reviewers

Individual job names change (`Build and check` → `Build and check (ubuntu-latest)`, etc.), so any
branch-protection rules pinned to the old names need updating. The aggregate `build-check-test`
job keeps its name and remains the single required check.

---

> [!NOTE]
> ### Run CI and nightly stress tests across Ubuntu, macOS, and Windows in a matrix
> - Converts build, check, and test jobs in [ci.yml](.github/workflows/ci.yml) and [nightly-process-stress.yml](.github/workflows/nightly-process-stress.yml) to OS matrix jobs across `ubuntu-latest`, `macos-latest`, and `windows-latest`; job timeouts are increased from 15 to 30 minutes.
> - Adds per-OS installation steps for system dependencies (`apt`, Homebrew, Chocolatey) and `uv`; `uv` is installed via Homebrew on macOS and `pip --user` on Linux/Windows with the appropriate bin directory added to `PATH`.
> - Fixes `defaultDaemonSocketDir` in [daemon-socket.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/781/files#diff-59c4ea71700afcf7d09d14c5d30e2c73fd0c71b28284e8758cd8fbd4b2337dbc) to fall back to a deterministic SHA-256-hashed `/tmp` subdirectory when the `tmpdir`-based socket path would exceed the platform's UNIX domain socket path limit (104 bytes on macOS, 108 on Linux).
> - Guards permission-bit and symlink assertions in [telemetry.test.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/781/files#diff-c66ea84d0d1cb3278023b8a31ff450dbe9085c13ab1de7616ddd468f7b938d5f) behind runtime capability probes so tests pass on Windows where these features are not meaningful.
> - Behavioral Change: Windows CI failures are reported but do not gate merges (`continue-on-error: true`).
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 577285b.</sup>
> 
>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,9 +15,15 @@ permissions:
 
 jobs:
   build-check:
-    name: Build and check
-    runs-on: ubuntu-latest
-    timeout-minutes: 15
+    name: Build and check (${{ matrix.os }})
+    runs-on: ${{ matrix.os }}
+    # Reported but not enforced; see the note on the test job.
+    continue-on-error: ${{ matrix.os == 'windows-latest' }}
+    timeout-minutes: 30
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
     steps:
       - name: Checkout
         uses: actions/checkout@v7
@@ -28,11 +34,22 @@ jobs:
           node-version: 22
           cache: npm
 
-      - name: Install system dependencies
+      # fd and ripgrep back the file-search tools. The image libraries build
+      # `canvas` from source, which only Linux lacks a prebuild for.
+      - name: Install system dependencies (Linux)
+        if: runner.os == 'Linux'
         run: |
           sudo apt-get update
           sudo apt-get install -y libcairo2-dev libpango1.0-dev libjpeg-dev libgif-dev librsvg2-dev fd-find ripgrep
           sudo ln -s $(which fdfind) /usr/local/bin/fd
+
+      - name: Install system dependencies (macOS)
+        if: runner.os == 'macOS'
+        run: brew install fd ripgrep
+
+      - name: Install system dependencies (Windows)
+        if: runner.os == 'Windows'
+        run: choco install fd ripgrep --no-progress -y
 
       - name: Install dependencies
         run: npm ci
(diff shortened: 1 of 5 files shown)


SandroHub013:
The matrix is the right place to catch the Windows daemon gaps being reported in #841 and #917. Please include a smoke case where a default named-pipe daemon has no worker descriptor and is still found by `status`/`doctor`, plus a process-tree force-stop case. Herdr's Windows endpoint also needs a small integration smoke test: `HERDR_SOCKET_PATH` is path-like (`C:\Users\...\herdr.sock`) but the Node endpoint is `\\.\pipe\` plus that value.

noahsabaj:
Thanks — taking the three in turn.

**Herdr.** The premise does not hold in this tree. `herdr-agent-state.ts` calls `createConnection(socketPath)` with the value of `HERDR_SOCKET_PATH` verbatim; nothing prepends `\\.\pipe\`. The Windows integration smoke test also already exists: `test/herdr-agent-state.test.ts` starts the fake herdr server on a named pipe when `process.platform === "win32"` and points the env var at it. Twelve tests, green on Windows 11 here. If you have a build where the endpoint is composed rather than passed through, please point me at the line — I would like to see it.

**The daemon cases.** Both are worth having, and I do not think they belong in this PR. This one changes CI wiring only: it adds macOS, folds the three platforms into one `(os × suite)` matrix, and fixes the uv bootstrap. A no-worker-descriptor `status`/`doctor` case and a process-tree force-stop case are new daemon coverage, and reviewing them here would hold up the matrix that has to exist before either can run on Windows at all.

The right home is the port PRs that own those paths — #841 is session-lease recovery and #917 is orphan-tree teardown. Once the matrix lands, any suite added there runs on all three platforms by construction, which is the point of doing it this way round.

noahsabaj:
Rebased onto current `main` and added one commit, because the analytics work that landed in the meantime brought a test that cannot pass on the Windows legs this PR adds.

`test/telemetry.test.ts` makes two POSIX-only assumptions:

- `expect(statSync(path).mode & 0o777).toBe(0o600)` — Windows does not implement permission bits, so a file written with mode `0o600` reads back as `0o666`. The assertion is now guarded rather than the case skipped, since the identity and persistence assertions around it are worth running everywhere.
- `symlinkSync` throws `EPERM` without SeCreateSymbolicLinkPrivilege, which only Developer Mode or an elevated shell grants. The case now probes the privilege once and skips when it is absent, matching how `packages/tui/test/autocomplete.test.ts` already handles this.

Both fail deterministically on Windows 11 today and pass with the commit. This is the same reason the macOS `sun_path` fix is in this PR: a matrix that adds a platform has to make that platform's legs report honestly.

Still no CI run on this PR — the checks sit at `action_required` pending maintainer approval, so `build-check-test` has never executed. Nothing here can be verified upstream until someone approves the workflow.

kevinjosethomas:

PR #782, opened 2026-08-06
fix(ai): restore Kimi K3 reasoning levels

paralin:
Kimi K3 accepts the standard reasoning effort levels, but its generated model metadata marked every level except max as unsupported. Requests for high effort were therefore clamped to max before reaching the provider.

Remove the unsupported-level overrides while retaining explicit xhigh and max mappings. Regenerate the provider catalogs and update the regression coverage to require every supported level.

> [!NOTE]
> ### Restore Kimi K3 reasoning levels and add OpenAI Codex provider-native compaction
> - Fixes `KIMI_K3_THINKING_LEVEL_MAP` in [generate-models.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/782/files#diff-89e0723f62b802c23d445c0de0813e3c9a937463cdc57f1585eedd07c9a4635c) so all reasoning levels (`off` through `max`) are surfaced; previously only `xhigh` was mapped, restricting model selection to a single level.
> - Adds provider-native compaction via a `POST /compact` endpoint for the `openai-codex-responses` provider, returning replacement history and a compaction item that can be replayed in subsequent sessions.
> - Introduces `ClaudeCodeRuntime` and `createClaudeCodeFamilyMcpServer` to allow Claude Code subagents to run as RLM children alongside native subagents, with family `list`/`send`/`inbox`/`wait` coordination tools.
> - Adds a durable family mailbox to `AgentDaemon` with stable message IDs, reply correlation, idempotent retry-on-resend, and waiter-based delivery with timeout and cancellation semantics.
> - Extends `SessionManager` and compaction to prefer provider-compatible native history boundaries and fall back to full replay when none are available; model role-based provider fallback is now supported with ordered candidate lists.
> - Behavioral Change: default compaction now opts into provider-native compaction (`native: true`); sessions with a prior native compaction boundary will replay provider history instead of re-summarizing when the active provider matches.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized cc54f1b.</sup>
> 
>


paralin:
Oops, accidentally included the compaction series. Will open a new PR.

PR #786, opened 2026-08-06
fix(daemon): namespace socket by agent dir without lengthening paths

### LINKED ISSUE #768: Daemon socket identity ignores PRIME_AGENT_CODING_AGENT_DIR — sessions silently run under another install's daemon
## Summary

The daemon socket path is derived from uid only (`$TMPDIR/prime-agent-<uid>/daemon.sock`, `packages/coding-agent/src/modes/daemon/daemon-socket.ts` → `defaultDaemonSocketDir()`). `PRIME_AGENT_CODING_AGENT_DIR` (`src/config.ts` → `ENV_AGENT_DIR`) changes which state directory an instance uses, but it does **not** participate in daemon identity. Two instances configured with different agent dirs therefore share one daemon socket, and whichever daemon starts first serves both: the second instance's sessions silently execute under the first instance's code and state.

This is nasty to debug because it is almost invisible: the daemon protocol forwards the client's env to the worker, so session files still land under the client's agent dir. Everything looks right — except the code that actually ran is the other install's. We lost a day to this while A/B-testing a patched build against a vanilla install: our build's new session-entry writes never appeared, with zero errors anywhere, because a stale vanilla daemon (started earlier by a `--version` invocation) was executing every session.

**Tested at:** `main` @ `c5991bc8`, v0.7.0 base · macOS 26.0 (Darwin 25.5.0) · Node v22.23.1 · uid 501

## Reproduction (cross-instance hijack)

1. Install prime-agent normally. Run any command once so a daemon starts (e.g. `prime-agent -p "hi"`).
2. In another shell: `PRIME_AGENT_CODING_AGENT_DIR=~/.prime-b/agent prime-agent -p "hi"`.
3. Observe: no second daemon starts. The second run's session is served by the daemon from step 1 (its worker process, its loaded code), while its session JSONL is written under `~/.prime-b/agent/sessions/` via the forwarded env.

**Expected:** distinct agent dirs get distinct daemons (or at minimum a hard mismatch error), since the agent dir determines settings, harness state, sessions, and — for anyone testing a patched build — the code the daemon loaded.

Version-mismatch self-healing does not cover this: both instances can be the same version and still be different installs/states.

## Constraint that shapes the fix (related: #669)

Our first fix attempt suffixed the socket **directory** (`prime-agent-<uid>-<hash8>/`) and broke daemon workers on startup: `bind()` succeeded but created the socket at a truncated path, and the follow-up `lstat` of the full path failed with `ENOENT` inside `Server.onListening` (`getDaemonSocketIdentity`), so create timed out.

Cause: macOS `sun_path` is 104 bytes (103 usable). Worker socket paths in the shared dir are already at the edge. With the standard macOS `$TMPDIR` (49 chars, `/var/folders/xx/<30 chars>/T/`):

| uid width | `$TMPDIR/prime-agent-<uid>/worker-<12hex>-<12hex>.sock` | fits? |
|-----------|----------------------------------------------------------|-------|
| 3 (e.g. 501) | 102 | 1 char headroom |
| 4 (e.g. 1000) | 103 | exactly at the limit |
| 5 (directory-service uids on managed Macs) | 104 | broken today, stock install |

So this is not only a multi-instance concern: any macOS user whose uid is 5+ digits (common with AD/LDAP-managed machines) should currently hit the same silent-truncation worker crash on an unmodified install. A slightly longer `$TMPDIR` does the same. That path-length failure is already tracked in #669; calling it out here because it constrains how Part 1 can be fixed without making workers worse.

## Suggested fix (reference)

Namespace the daemon socket **filename**, not the directory, so worker path lengths are unchanged:

```ts
// daemon-socket.ts
export function defaultDaemonSocketPath(): string {
  if (process.platform === "win32") {
    return `\\\\.\\pipe\\prime-agent-daemon-${agentDirSocketSuffix()}`;
  }
  return join(defaultDaemonSocketDir(), `daemon-${agentDirSocketSuffix()}.sock`);
}

function agentDirSocketSuffix(): string {
  return createHash("sha256").update(resolve(getAgentDir())).digest("hex").slice(0, 8);
}
```

`defaultDaemonSocketDir()` stays `prime-agent-<uid>`. Worker socket names are already unique per daemon id, so they need no change.

Happy to share the full diff (including test updates) if useful; an alternative acceptable fix would be a handshake that compares resolved agent dirs and refuses with a clear error.

MumuTW:
## Summary
- Include `PRIME_AGENT_CODING_AGENT_DIR` in daemon socket identity so distinct agent dirs get distinct daemons (instead of silently sharing whichever uid-keyed daemon started first).
- Put the agent-dir hash in the socket **filename** (`daemon-<hash8>.sock`), not the directory, so worker socket paths stay within the macOS `sun_path` limit.
- Cover distinct agent dirs (Unix + Windows pipe) in `daemon-socket` tests.

Fixes #768. Related path-length constraint: #669.

## Test plan
- [x] `npm --prefix packages/coding-agent test -- test/daemon-socket.test.ts` (8/8)
- [ ] Manually: start a vanilla daemon (`prime-agent -p "hi"`), then `PRIME_AGENT_CODING_AGENT_DIR=~/.prime-b/agent prime-agent -p "hi"` and confirm a second daemon/socket under `$TMPDIR/prime-agent-<uid>/daemon-<hash>.sock`
- [ ] Confirm `prime-agent doctor` / `ps` still lists both daemons via socket-dir scan

Made with [Cursor](https://cursor.com)

> [!NOTE]
> ### Namespace daemon socket path by agent directory using an 8-char hash suffix
> - `defaultDaemonSocketPath` in [`daemon-socket.ts`](https://github.com/PrimeIntellect-ai/prime-agent/pull/786/files#diff-59c4ea71700afcf7d09d14c5d30e2c73fd0c71b28284e8758cd8fbd4b2337dbc) now appends an 8-hex-character suffix derived from the SHA-256 hash of the resolved `PRIME_AGENT_CODING_AGENT_DIR` path, producing distinct socket names per agent directory.
> - On Unix, only the filename changes (e.g. `daemon-<hash>.sock`); the per-user directory is unchanged, keeping paths within `sun_path` limits.
> - On Windows, the named pipe name gains the same suffix (e.g. `\\.\pipe\prime-agent-daemon-<hash>`).
> - Behavioral Change: any existing code or tooling that assumes a fixed socket path (`daemon.sock` or `prime-agent-daemon`) will no longer find the socket at that location.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized aa353ab.</sup>
> 
>

### DIFF
--- a/packages/coding-agent/src/modes/daemon/daemon-socket.ts
+++ b/packages/coding-agent/src/modes/daemon/daemon-socket.ts
@@ -1,8 +1,10 @@
+import { createHash } from "node:crypto";
 import { chmodSync, existsSync, lstatSync, mkdirSync, unlinkSync } from "node:fs";
 import { createConnection } from "node:net";
 import { tmpdir } from "node:os";
-import { dirname, join } from "node:path";
+import { dirname, join, resolve } from "node:path";
 import lockfile from "proper-lockfile";
+import { getAgentDir } from "../../config.js";
 
 const DAEMON_SOCKET_MODE = 0o600;
 const DAEMON_SOCKET_DIR_MODE = 0o700;
@@ -35,9 +37,9 @@ export interface DaemonSocketIdentity {
 
 export function defaultDaemonSocketPath(): string {
 	if (process.platform === "win32") {
-		return "\\\\.\\pipe\\prime-agent-daemon";
+		return `\\\\.\\pipe\\prime-agent-daemon-${agentDirSocketSuffix()}`;
 	}
-	return join(defaultDaemonSocketDir(), "daemon.sock");
+	return join(defaultDaemonSocketDir(), `daemon-${agentDirSocketSuffix()}.sock`);
 }
 
 export async function acquireDaemonSocketPathLease(socketPath: string): Promise<DaemonSocketPathLease | undefined> {
@@ -214,6 +216,21 @@ function assertSocketLease(socketPath: string, lease: DaemonSocketPathLease): vo
 export function defaultDaemonSocketDir(): string {
 	const suffix = typeof process.getuid === "function" ? String(process.getuid()) : "user";
 	return join(tmpdir(), `prime-agent-${suffix}`);
+}
+
+/**
+ * Daemon identity must include the agent state dir: two installs sharing one
+ * uid-keyed socket (e.g. a fork pointed at PRIME_AGENT_CODING_AGENT_DIR and a
+ * vanilla install) would otherwise attach to whichever daemon started first and
+ * silently run sessions under the other install's code and state.
+ *
+ * The suffix goes into the socket FILENAME, not the directory: worker socket
+ * paths in the shared directory already sit at ~103 chars, right at the macOS
+ * sun_path limit (104) — a longer directory name makes bind() silently
+ * truncate the path and every later stat of the full path fail with ENOENT.
+ */
+function agentDirSocketSuffix(): string {
+	return createHash("sha256").update(resolve(getAgentDir())).digest("hex").slice(0, 8);
 }
 
 function ensureDefaultDaemonSocketDir(socketPath: string): void {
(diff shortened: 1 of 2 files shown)


kevinjosethomas:

PR #795, opened 2026-08-06
fix(agent): fail closed on provider quota exhaustion

kcoopermiller:
This PR prevents Prime Agent from continuing background inference after a provider reports quota or credit exhaustion.

Quota exhaustion was previously handled like a transient rate limit. This allowed retries to occur at both the provider SDK and agent layers while detached subagents and daemon-scheduled work could remain active. With multiple agents running concurrently, those retries could substantially amplify model usage.

The new behavior treats quota exhaustion as a terminal, family-wide condition. Prime Agent stops the current session and its subagents, blocks automatic background work, and requires an explicit user prompt before attempting inference again.

---

> [!NOTE]
> **High Risk**
> Changes core inference, retry, and multi-session orchestration on billing-sensitive failures; misclassification could block legitimate work or still allow unwanted retries.
> 
> **Overview**
> **Provider stream failures** now include a dedicated `quota` kind: HTTP 402, quota/credit language in errors, and nested SDK messages are classified separately from transient 429 throttling (per-minute metrics, “resource exhausted” hints). `classifyStreamFailure` takes optional detail text so quota wording can override a generic `rate_limit_error` type.
> 
> **Agent sessions** trip a **provider quota circuit** on the root when any family member hits quota: abort in-flight work, cancel queued provider-backed actions (turns, `/compact`, inference `/refine`), clear agent queues, stop RLM children, and persist trip/reset markers plus a user-visible notice. Subagents share state via `rlmParentSession`; daemon/cron prompts set `automatic: true` so they cannot clear the latch—only an explicit user prompt does. Quota failures are excluded from agent-level auto-retry; inference-free `/refine rollback` still works.
> 
> **Retry defaults**: provider SDK `maxRetries` defaults to **0** (agent-level retry owns policy), documented in settings. Custom messages persist with rollback on append failure so half-written quota notices do not corrupt the branch.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1164892c5245e8e36aeac05b156d4fd7322bc133. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Fail closed on provider quota exhaustion by implementing a quota circuit breaker in agent sessions
> - Introduces a provider quota circuit in `AgentSession` that trips when a quota/credit exhaustion error is detected, halting all subsequent provider-backed work (turns, compaction, refinement, subagent spawning) until an explicit user prompt resets it.
> - Adds a new `'quota'` stream failure kind in [stream-failure.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/795/files#diff-2587e64b20a4d7daabfd77ca9486cd89366592abf324c6a08c41e8a8822aa474) with helpers to distinguish true quota exhaustion from transient rate-limit throttling; HTTP 402 responses are classified as `'quota'`.
> - Quota circuit state is persisted to the session branch and propagated across the session family (parent → subagents) so all related sessions share the same safety latch.
> - Daemon-originated automatic/cron prompts are marked with `PromptOptions.automatic = true` so they cannot reset a tripped circuit; only an explicit user prompt clears it.
> - `provider.maxRetries` now defaults to `0` to avoid repeated costly retries on quota failures.
> - Risk: any session that hits a quota error will stop all autonomous background work immediately and stay stopped until the user sends a new prompt.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 1164892.</sup>
> 
>

### DIFF
--- a/packages/ai/src/utils/stream-failure.ts
+++ b/packages/ai/src/utils/stream-failure.ts
@@ -13,6 +13,7 @@ export type StreamFailureKind =
 	| "safety"
 	| "overloaded"
 	| "rate_limit"
+	| "quota"
 	| "server_error"
 	| "auth"
 	| "invalid_request"
@@ -44,6 +45,7 @@ const KIND_MESSAGES: Record<StreamFailureKind, string> = {
 	safety: "Response blocked by provider safety filters",
 	overloaded: "Provider overloaded",
 	rate_limit: "Provider rate limit exceeded",
+	quota: "Provider usage quota or credit exhausted",
 	server_error: "Provider server error",
 	auth: "Provider authentication failed",
 	invalid_request: "Provider rejected the request",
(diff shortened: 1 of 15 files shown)


nnunley:
I reproduced the failure mode this PR targets with the Anthropic OAuth provider on Prime Agent 0.7.1.

Two requests remained silent for 28m05s and 27m14s. The session log recorded zero-token `aborted` messages only after user recovery aborts. A direct Claude request then emitted `rate_limit_event: You've hit your session limit`, and authenticated usage showed five-hour utilization at 100%, weekly utilization at 84%, and extra usage disabled.

The proposed `maxRetries: 0` default and terminal quota classification would remove the hidden SDK retry amplification once an error reaches the agent. Anthropic also sends `anthropic-ratelimit-unified-{5h,7d}-utilization` and reset headers on successful OAuth responses. Retaining that last-known snapshot would let the runtime warn before the next request stalls and include evidence in the quota-circuit notice. I am filing that pre-exhaustion propagation separately because it complements rather than duplicates this PR.

PR #804, opened 2026-08-06
show current openrouter models

kristianernst:
- openrouter adds and drops models constantly, but our list came from a snapshot baked in at release time. new models like meta/muse-spark-1.2 never showed up, and models openrouter had removed stayed in the picker.
- the catalog is now fetched from openrouter when you open the model picker or run `model list`. the snapshot is still used offline, when the request fails, and as the fallback before the first successful fetch.
- reasoning info comes from the live api too. muse spark 1.2 has to think, so it no longer offers "off", and effort levels a model does not support are hidden instead of being offered and failing.
- your models.json overrides, custom models, and provider logins still win over anything fetched. the router aliases openrouter/auto and the ~*-latest entries are kept even though the filtered endpoint does not return them.
- cached for 5 minutes with a 3 second timeout, one shared request between callers, and a 30 second backoff after a failure. startup never waits on the network, so nothing feels slower.
- the generator still has its own copy of this parsing. wiring it to the shared parser and regenerating the snapshot is a small follow-up (measured as a one line diff), left out here to keep this reviewable.

tested against the live api: 275 openrouter models with muse spark 1.2 present, 272 from the snapshot with PI_OFFLINE=1. also verified in a clean container with no daemon or prior state.

> [!NOTE]
> ### Show live OpenRouter model catalog with fallback to static snapshot
> - Adds `parseOpenRouterModels` in [`packages/ai/src/openrouter-models.ts`](https://github.com/PrimeIntellect-ai/prime-agent/pull/804/files#diff-72b94a81b18b69003032ae3e45c50805864bb85275242ccd014789a860b797ce) to normalize raw OpenRouter catalog entries into typed `Model<'openai-completions'>` objects, filtering non-tool, `:batch`, and non-text-output routes.
> - Adds `getOpenRouterModels` in [`packages/coding-agent/src/core/openrouter-model-catalog.ts`](https://github.com/PrimeIntellect-ai/prime-agent/pull/804/files#diff-826a7bde17d30273d5f74ccb977e7c9735958ff4dc6cc6c05b71d559433ee627): a process-wide cached fetcher with TTL, timeout, deduplication, and last-known-good fallback on failures.
> - `ModelRegistry` merges live OpenRouter data with the curated static snapshot, preserving virtual aliases (`auto`, `openrouter/*`, `~*`) and static compat/featured flags absent from live results.
> - The `list-models` CLI now calls `refreshOpenRouterModels()` before listing so newly released models appear without a snapshot rebuild.
> - Behavioral Change: on each registry refresh, a live HTTP fetch to OpenRouter is attempted; cold failures return `undefined` and fall back silently to the static snapshot.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized e7fdc2f.</sup>
> 
>

### DIFF
--- a/packages/ai/CHANGELOG.md
+++ b/packages/ai/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## [Unreleased]
 
+- Added a shared strict parser for the OpenRouter model catalog, exposed from the root package export.
+
 ## [0.7.0] - 2026-08-05
 
 ## [0.6.1] - 2026-08-05
--- a/packages/ai/src/index.ts
+++ b/packages/ai/src/index.ts
@@ -5,6 +5,7 @@ export * from "./api-registry.js";
 export * from "./env-api-keys.js";
 export * from "./log.js";
 export * from "./models.js";
+export * from "./openrouter-models.js";
 export type { BedrockOptions, BedrockThinkingDisplay } from "./providers/amazon-bedrock.js";
 export type { AnthropicEffort, AnthropicOptions, AnthropicThinkingDisplay } from "./providers/anthropic.js";
 export type { AzureOpenAIResponsesOptions } from "./providers/azure-openai-responses.js";
--- a/packages/ai/src/openrouter-models.ts
+++ b/packages/ai/src/openrouter-models.ts
@@ -1,0 +1,111 @@
+import type { KnownProvider, Model } from "./types.js";
+
+const OPENROUTER_BASE_URL = "https://openrouter.ai/api/v1";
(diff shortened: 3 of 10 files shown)


kristianernst:
Superseded by https://github.com/PrimeIntellect-ai/prime-agent/pull/1310, rebased onto current main after #1258. That PR reuses `getOpenRouterReasoningCapabilities` instead of a parallel effort parser.

kristianernst:
closing it because some functionality has already been included by a later PR and replacing it with a leaner #1310

PR #805, opened 2026-08-07
feat(coding-agent): keep ACP daemon sessions resident

sethkarten:
## Why

ACP hardcoded a client-owned daemon session, so the worker — and with it the live IPython kernel — stopped when the ACP client disconnected. Any ACP consumer that wants state to survive a turn boundary therefore had to keep one long-lived client process alive for the whole rollout, and build keepalive, transport and teardown machinery around it.

`resident` was already on the wire (`daemon-protocol.ts`: `DaemonSessionLifecycle = "resident" | "client_owned"`), so ACP simply never asked for it.

## What changed

`clientOwned: appMode !== "acp"` at the daemon-client call site. ACP gets a resident worker; `rpc`, `print` and `json` keep client-owned lifetimes unchanged.

No protocol or schema change, and no new capability. An earlier draft of this branch bumped `DAEMON_PROTOCOL_VERSION` and negotiated an `acp_resident_sessions` capability; both were removed. The bump was actively harmful — startup matches version and schema exactly, so protocol-7 daemons would go stale *before* any fallback could engage, letting a busy old daemon block ACP startup. A silent fallback to client-owned was also rejected: it would quietly destroy the persistence this change exists to provide.

## Re-attach

A resident worker is only useful if a later client can get back to the session. Prime Agent already supports this: relaunching ACP with `--continue` resolves the resident session by canonical session file and attaches to it. No `session/load` and no new ACP method are required.

## Validation

- `npm run check`
- `test/acp-resident-lifecycle.test.ts` and `test/daemon-protocol.test.ts`: 14 passed, 1 skipped (the skipped one is `kernel-heavy`, run separately below)
- kernel-heavy re-attach test, run explicitly: passes in ~12s against a live kernel. It sets an `object()` marker, disconnects ACP stdio, relaunches with `--continue`, and asserts live namespace identity — `object()` cannot survive a dill snapshot, so this proves the same interpreter, not a revived one.
- mutation: forcing `clientOwned: true` fails that test (65s, no kernel survives); restoring passes (12s). Reverting protocol to 8 fails the schema/version assertions.

Two pre-existing failures in `acp-features` (goal timeout, post-close new session) reproduce on `main` and are unrelated.

---

> [!NOTE]
> **High Risk**
> Changes daemon session lifecycle, worker recovery, and on-disk descriptors plus credential handling; mistakes could leak secrets, break ACP/embedder model access, or leave workers in the wrong lifecycle after disconnect.
> 
> **Overview**
> **ACP daemon lifecycle** now treats session-file ACP runs as **resident** workers (survive client disconnect; `--continue` can reattach), while ACP `--no-session` and other modes stay **client-owned**. `isClientOwnedDaemonSession` encodes that rule at the daemon create call site.
> 
> **Launch environment** is always sent on create (`launchEnv: collectDaemonLaunchEnv()` for both lifecycles) so the first worker spawn gets caller endpoint/credentials/proxy settings. Only **non-secret** keys are written to resident worker descriptors via `filterPersistedDaemonLaunchEnv` / `DAEMON_PERSISTED_LAUNCH_ENV_KEYS`; on load, descriptors are sanitized and rewritten if needed. Supervisor **recovery** relaunches resident workers from the persisted allowlist (not ambient supervisor secrets); client-owned workers keep transient env from the reconnecting owner. Promoting owned→resident clears `launchEnv` on the descriptor.
> 
> **Tests** cover lifecycle negotiation, allowlist filtering, supervisor restart + worker recovery without credential persistence, and a kernel-heavy ACP disconnect/`--continue` reattach integration test.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0626d45292471a0d446efb6041583542c06582c5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Keep ACP daemon sessions resident across supervisor restarts
> - ACP sessions with a session file are now classified as resident (not client-owned) via the new `isClientOwnedDaemonSession` utility; `--no-session` ACP invocations remain client-owned.
> - Resident workers persist and recover a safe allowlist of launch environment variables (defined in `DAEMON_PERSISTED_LAUNCH_ENV_KEYS`) across supervisor restarts, while client credentials and secrets are excluded.
> - When a resident worker is auto-recovered, the supervisor builds its environment from a filtered copy of `process.env` to avoid leaking ambient secrets from the previous worker process.
> - Promoting a client-owned worker to resident clears its `launchEnv` from the descriptor so transient credentials are not persisted.
> - Behavioral Change: daemon worker descriptors now include an optional `launchEnv` field; existing descriptors are sanitized on load to remove any non-allowlisted keys.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 0626d45.</sup>
> 
>

### DIFF
--- a/packages/coding-agent/src/modes/daemon/daemon-supervisor.ts
+++ b/packages/coding-agent/src/modes/daemon/daemon-supervisor.ts
@@ -2096,8 +2096,7 @@ export class DaemonSupervisor {
 			throw new Error(`Session worker ${existing.descriptor.workerId} recovery was cancelled`);
 		}
 		const recoveryStopRevision = existing?.stopRevision;
-		const launchEnv =
-			ownerClientId || existing?.descriptor.ownerClientId ? (command.launchEnv ?? existing?.launchEnv) : undefined;
+		const launchEnv = command.launchEnv ?? existing?.launchEnv;
 		const createCommand: DaemonCreateCommand = {
 			...withoutSupervisorCreateFields(command),
 			config: mergeAgentSessionRuntimeConfig(this.defaultSessionConfig, command.config),
--- a/packages/coding-agent/src/modes/daemon/daemon-supervisor.ts
+++ b/packages/coding-agent/src/modes/daemon/daemon-supervisor.ts
@@ -416,6 +416,12 @@ function isDaemonWorkerDescriptor(value: unknown, socketPath: string): value is 
 		(descriptor.pid ?? 0) > 0 &&
 		(descriptor.processStartId === undefined || typeof descriptor.processStartId === "string") &&
 		(descriptor.ownerClientId === undefined || typeof descriptor.ownerClientId === "string") &&
+		(descriptor.launchEnv === undefined ||
+			(typeof descriptor.launchEnv === "object" &&
+				descriptor.launchEnv !== null &&
+				Object.entries(descriptor.launchEnv).every(
+					([key, value]) => typeof key === "string" && typeof value === "string",
+				))) &&
 		typeof descriptor.socketPath === "string" &&
 		typeof descriptor.authenticationToken === "string" &&
 		typeof descriptor.rootActiveSessionId === "string" &&
@@ -936,6 +942,7 @@ export class DaemonSupervisor {
 					snapshotLoads: new Map(),
 					intentionalStop: descriptor.stopRequestedAt !== undefined,
 					stopRevision: 0,
+					launchEnv: descriptor.launchEnv,
 				});
 			} catch (error) {
 				this.log(`Ignoring invalid worker descriptor ${path}: ${String(error)}`);
@@ -2070,7 +2077,7 @@ export class DaemonSupervisor {
 			throw new Error("Session is not owned by this client");
 		}
 		const previousDescriptor = worker.descriptor;
-		worker.descriptor = { ...previousDescriptor, ownerClientId: undefined };
+		worker.descriptor = { ...previousDescriptor, ownerClientId: undefined, launchEnv: undefined };
 		try {
 			this.persistWorker(worker);
 		} catch (error) {
@@ -2096,7 +2103,8 @@ export class DaemonSupervisor {
 			throw new Error(`Session worker ${existing.descriptor.workerId} recovery was cancelled`);
 		}
 		const recoveryStopRevision = existing?.stopRevision;
-		const launchEnv = command.launchEnv ?? existing?.launchEnv;
+		const launchEnv = command.launchEnv ?? existing?.launchEnv ?? existing?.descriptor.launchEnv;
+		const ownerClientIdForDescriptor = existing?.descriptor.ownerClientId ?? ownerClientId;
 		const createCommand: DaemonCreateCommand = {
 			...withoutSupervisorCreateFields(command),
 			config: mergeAgentSessionRuntimeConfig(this.defaultSessionConfig, command.config),
@@ -2173,7 +2181,8 @@ export class DaemonSupervisor {
 				supervisorSocketPath: this.socketPath,
 				authenticationToken: token,
 				rootActiveSessionId,
-				ownerClientId: existing?.descriptor.ownerClientId ?? ownerClientId,
+				ownerClientId: ownerClientIdForDescriptor,
+				...(ownerClientIdForDescriptor === undefined && launchEnv ? { launchEnv } : {}),
 				createdAt: existing?.descriptor.createdAt ?? now,
 				updatedAt: now,
 				lifecycle: "starting",
--- a/packages/coding-agent/src/modes/daemon/daemon-supervisor.ts
+++ b/packages/coding-agent/src/modes/daemon/daemon-supervisor.ts
@@ -54,6 +54,7 @@ import { DAEMON_CATALOG_ROLE_ENV, DaemonCatalogClient } from "./daemon-catalog-p
 import { deserializeDaemonError, serializeDaemonError } from "./daemon-errors.js";
 import {
 	collectDaemonClientEnv,
+	collectDaemonLaunchEnv,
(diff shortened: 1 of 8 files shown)


packages/coding-agent/src/modes/daemon/daemon-supervisor.ts
```diff
@@ -2172,7 +2179,8 @@ export class DaemonSupervisor {
 				supervisorSocketPath: this.socketPath,
 				authenticationToken: token,
 				rootActiveSessionId,
-				ownerClientId: existing?.descriptor.ownerClientId ?? ownerClientId,
+				ownerClientId: ownerClientIdForDescriptor,
+				...(ownerClientIdForDescriptor === undefined && launchEnv ? { launchEnv } : {}),
```
snimu: This persists the caller's entire environment to the descriptor file. `collectDaemonLaunchEnv()` is essentially all of `process.env` minus `PRIME_AGENT_INTERNAL_*`, and since `createDaemonClientConnection` now sends `launchEnv` unconditionally, this isn't ACP-only: the interactive daemon path goes through the same function, so normal TUI sessions get their full env — every API key, `AWS_*`, `GH_TOKEN`, ssh agent sockets, whatever is in the user's shell — written to `~/.prime/agent/daemon-workers/<id>.json`.

The files are 0600 and provider keys already live in `auth.json` next door, so it's not a new class of exposure, but it's a much broader at-rest secret surface, it isn't scoped to what the worker needs, and rotated credentials get silently resurrected on relaunch. The changelog bullet ("keeping client-owned credentials out of descriptor files") describes the client-owned side and rather buries that resident descriptors now contain everything.

There's also a second-order behavior change worth stating in the PR description: resident interactive workers were previously launched with the supervisor's env; now they're launched with the client's env.

At minimum this needs an explicit callout in the changelog and daemon docs; better would be persisting a filtered/allowlisted env.

packages/coding-agent/test/acp-resident-lifecycle.test.ts
```diff
@@ -0,0 +1,332 @@
...
+			stdio: ["pipe", "pipe", "pipe"],
+		},
+	);
+	children.add(child);
+	return new AcpStdioClient(child);
+}
+
+describe("ACP daemon lifecycle negotiation", () => {
+	it("uses resident lifecycle only for ACP sessions that can be reattached", () => {
+		const source = readFileSync(resolve(__dirname, "../src/main.ts"), "utf8");
```
snimu: These two tests grep `main.ts` for literal source strings. They break on any refactor of that code and keep passing on behavioral regressions that happen to preserve the text — asserting incidental details rather than behavior. The kernel-heavy re-attach test and the supervisor launchEnv test carry the real weight here; I'd drop these two or replace them with behavior-level assertions, e.g. checking the create command a fake daemon client receives for each appMode/noSession combination.

packages/coding-agent/test/daemon-supervisor-process.test.ts
```diff
@@ -399,6 +399,78 @@ describe("daemon supervisor resident workers", () => {
...
+		while (!recoveredSummary && Date.now() < recoveryDeadline) {
+			const listed = await replacementClient.request({ type: "list" });
+			recoveredSummary = requireSessionList(listed.success ? listed.data : undefined).find(
+				(session) => session.workerPid !== adoptedSummary.workerPid,
+			);
+			if (!recoveredSummary) await new Promise((resolveDelay) => setTimeout(resolveDelay, 50));
+		}
+		expect(recoveredSummary).toBeDefined();
+		const markerLines = readFileSync(markerPath, "utf8").trim().split("\n");
+		expect(markerLines).toHaveLength(2);
```
snimu, line 491: This assertion is racy: it reads the marker file as soon as the recovered worker appears in `list`, but the extension writes the marker during worker boot, which happens after the summary becomes visible. It fails deterministically on my Mac (`expected [ Array(1) ] to have a length of 2`) and passes with a bounded poll loop around the marker read — confirmed locally. CI was green (Linux timing), but this file only runs in the process-smoke job and will flake there. Suggest polling for `markerLines.length === 2` (with a deadline) before asserting.

packages/coding-agent/src/modes/daemon/daemon-supervisor.ts
sethkarten: Addressed in 0626d4529. Initial launches still receive the caller environment transiently, but descriptors and automatic recovery now use an explicit operational allowlist (runtime paths, agent dir, and non-secret behavior settings). Legacy descriptors are scrubbed on load, and replacement supervisors cannot re-inject ambient API/AWS/GH/SSH credentials. The process regression proves both transient first-launch behavior and credential-free recovery.

packages/coding-agent/test/acp-resident-lifecycle.test.ts
sethkarten: Addressed in 0626d4529. The source-grep assertions were replaced by a behavior-level lifecycle decision test covering reattachable ACP, ACP --no-session, RPC, and print modes; the live kernel reattachment test remains the end-to-end proof.

packages/coding-agent/test/daemon-supervisor-process.test.ts
sethkarten: Addressed in 0626d4529. The recovery assertion now boundedly polls the marker for up to 15 seconds before checking both boot lines. The full daemon process suite passes locally with 9 passed / 6 intentionally skipped.

sethkarten:
Closing this superseded ACP implementation. Resident recovery coverage continues in #1236, with the exact causal prompt lifecycle successor in #1239. This PR remains available as design and review history.

PR #806, opened 2026-08-07
feat(coding-agent): expose ACP quiescence metadata

sethkarten:
## Why

An ACP client cannot currently tell whether a completed turn is safe to act on. Prime Agent subagents are fire-and-forget and keep working after the spawning turn ends, and autonomous mode may still have continuation budget left. A consumer that scores or ends the session at `end_turn` can therefore cut off live work, and the result depends on a race.

This is not derivable client-side: `tool_execution_update`, `auto_retry_*`, `auth_stale` and `recap_update` have no ACP representation, so there is no event stream a client could reconstruct quiescence from.

## What changed

Adds `quiescence` to the existing namespaced envelope:

```
_meta["ai.primeintellect.prime-agent"].quiescence = {
  outstandingSubagents: number,
  remainingAutonomousContinuations: number,
}
```

Counted from the authoritative live roster (queued plus running) rather than reconstructed from child deltas, and emitted after headless completion so it describes the state at the point the turn is reported done.

The turn **reports** rather than **waits**. Blocking `session/prompt` on quiescence would deadlock against detached subagent semantics — a child can legitimately outlive many turns — so the client decides what to do with the signal.

Additive and namespaced: vanilla ACP clients ignore unknown `_meta`. No previously-declared-but-unemitted field is touched.

## Validation

- `npm run check`: Biome 898 files, installer render, browser smoke all pass
- `test/suite/acp-mode.test.ts`: 2/2
- mutation: changing the outstanding count to `+ 1` fails the assertion; restoring passes

Stacked on #805.

---

> [!NOTE]
> **Medium Risk**
> Changes ACP `session/prompt` and `session/new` failure modes (roster read failures now reject prompts) and introduce authoritative child-roster plumbing used by external clients; behavior is additive in `_meta` but correctness-sensitive for scoring/automation.
> 
> **Overview**
> ACP clients get a **quiescence** signal in namespaced `_meta` after each `session/prompt` turn completes headless work: **outstanding subagents** (queued/running from a live roster) and **remaining autonomous continuations**. The turn still reports `end_turn` without blocking on detached children; clients choose how to act on the signal.
> 
> **`AgentSession.getRlmChildSnapshots()`** builds a recursive live child roster from active runs and retained child sessions (with model on runs), and connection snapshots/daemon clients merge **`rlm_child_update`** events so empty attach snapshots do not erase live children.
> 
> **ACP session lifecycle** hardening: subscribe before the initial snapshot, reject `session/new` on snapshot failure (unsubscribe cleanup), guard concurrent `session/new` with an in-flight flag, and **fail `session/prompt`** if the emission-time roster read fails instead of emitting a false zero count.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6447725af40a3354c373f0ee926b182803f6f098. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Expose ACP quiescence metadata with live subagent roster in `session/prompt` responses
> - `session/prompt` now emits `_meta.quiescence` after completion, reporting `outstandingSubagents` and `remainingAutonomousContinuations` from a live authoritative child roster snapshot.
> - `AgentSession.getRlmChildSnapshots` provides a recursive, flattened list of child agent snapshots (id, label, status, model, sessionDir) from both the active run registry and retained child sessions.
> - `DaemonAgentConnection` now updates its cached `latestSnapshot.children` as `rlm_child_update` events arrive, keeping the in-memory roster current.
> - `session/new` now rejects concurrent creation attempts while a first request is pending, and rejects (without claiming the session slot) if the initial roster snapshot read fails.
> - Risk: `session/prompt` and `session/new` now fail hard if the live roster snapshot read fails, rather than succeeding with incomplete metadata.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 6447725.</sup>
> 
>

### DIFF
--- a/packages/coding-agent/src/modes/acp/acp-mode.ts
+++ b/packages/coding-agent/src/modes/acp/acp-mode.ts
@@ -100,7 +100,6 @@ interface AcpSessionEntry {
 	id: string;
 	abort: AbortController | undefined;
 	unsubscribe: (() => void) | undefined;
-	children: Map<string, AgentConnectionRlmChildAgentSnapshot>;
 }
 
 /**
@@ -313,9 +312,10 @@ export async function runAcpModeWithConnection(
 				}
 			}
 			const sessionId = randomUUID();
-			const initialSnapshot = await connection.getInitialSnapshot().catch(() => undefined);
-			const children = new Map((initialSnapshot?.children ?? []).map((child) => [child.id, child]));
-			const entry: AcpSessionEntry = { id: sessionId, abort: undefined, unsubscribe: undefined, children };
+			// Install the listener before fetching the snapshot. Child updates can arrive
+			// while the snapshot request is in flight; the connection remains the
+			// authoritative source used when quiescence is emitted below.
+			const entry: AcpSessionEntry = { id: sessionId, abort: undefined, unsubscribe: undefined };
 			// Subscribe for the session lifetime, not per prompt turn: prime-agent
 			// subagents are fire-and-forget and keep reporting after the spawning turn
 			// ends, so a turn-scoped subscription would drop their updates. One
@@ -333,13 +333,15 @@ export async function runAcpModeWithConnection(
 					return;
 				}
 				if (event.type !== "session_event") return;
-				if (event.event.type === "rlm_child_update") children.set(event.event.child.id, event.event.child);
 				for (const update of acpUpdatesForSessionEvent(event.event, mappingState)) {
 					notify(update);
 				}
 			});
-			// Claim the single-session slot only once the subscription exists, so a
-			// failed subscribe cannot leave the slot occupied and unusable.
+			// Reconcile after subscribing so updates cannot be lost while the snapshot
+			// request is in flight. Do not turn a failed read into an empty roster.
+			await connection.getInitialSnapshot();
+			// Claim the single-session slot only once the subscription and snapshot are
+			// ready, so a failed setup cannot leave it occupied and unusable.
 			entry.unsubscribe = unsubscribe;
 			session = entry;
 			return {
@@ -374,9 +376,12 @@ export async function runAcpModeWithConnection(
 				// Snapshot after headless completion: detached subagents can publish a
 				// terminal update after the parent model turn has gone idle.
 				const autonomous = autonomousMeta(status);
+				// Read the authoritative live roster at emission time. A session-local
+				// shadow map can miss detached children or updates during the turn.
+				const liveSnapshot = await connection.getInitialSnapshot();
 				const meta = primeAgentMeta({
 					...(autonomous ? { autonomous } : {}),
-					quiescence: quiescenceMeta(status, [...entry.children.values()]),
+					quiescence: quiescenceMeta(status, liveSnapshot.children),
 				});
 				await ctx.client
 					.notify(acp.methods.client.session.update, {
--- a/packages/coding-agent/src/modes/acp/acp-mode.ts
+++ b/packages/coding-agent/src/modes/acp/acp-mode.ts
@@ -10,10 +10,10 @@ import type { AgentSessionRuntime } from "../../core/agent-session-runtime.js";
 import type { AgentAutonomousStatus } from "../../core/autonomous.js";
 import { takeOverStdout, writeRawStdout } from "../../core/output-guard.js";
 import { InProcessAgentConnection } from "../agent-connection/in-process-agent-connection.js";
-import type { AgentConnection } from "../agent-connection/types.js";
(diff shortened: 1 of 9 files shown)


packages/coding-agent/src/modes/acp/acp-mode.ts
```diff
@@ -320,8 +337,11 @@ export async function runAcpModeWithConnection(
 					notify(update);
 				}
 			});
-			// Claim the single-session slot only once the subscription exists, so a
-			// failed subscribe cannot leave the slot occupied and unusable.
+			// Reconcile after subscribing so updates cannot be lost while the snapshot
+			// request is in flight. Do not turn a failed read into an empty roster.
+			await connection.getInitialSnapshot();
+			// Claim the single-session slot only once the subscription and snapshot are
+			// ready, so a failed setup cannot leave it occupied and unusable.
```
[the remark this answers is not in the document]
sethkarten: Fixed in b8ba0c8bc: the `getInitialSnapshot()` await is wrapped so a throw calls `unsubscribe()` before rethrowing, and the session slot is still never claimed on a failed setup. A focused test asserts the listener is released when the snapshot throws; removing the `unsubscribe()` call fails it.

packages/coding-agent/src/modes/acp/acp-mode.ts
```diff
@@ -353,15 +391,23 @@ export async function runAcpModeWithConnection(
...
-							update: { sessionUpdate: "session_info_update", _meta: meta },
-						})
-						.catch(() => undefined);
-				}
+				// Snapshot after headless completion: detached subagents can publish a
+				// terminal update after the parent model turn has gone idle.
+				const autonomous = autonomousMeta(status);
+				// Read the authoritative live roster at emission time. A session-local
+				// shadow map can miss detached children or updates during the turn.
+				const liveSnapshot = await connection.getInitialSnapshot();
```
snimu, line 399: Reading the roster via `getInitialSnapshot()` doesn't work on either real transport, so `outstandingSubagents` is always 0 in practice:

- `InProcessAgentConnection` (the `runAcpMode` fallback): `createAgentConnectionSnapshot()` in `snapshot.ts` never populates `children` at all.
- `DaemonAgentConnection` (the normal ACP path once #805 lands): `getInitialSnapshot()` re-fetches state/messages/context but reuses `this.latestSnapshot?.children` — per the comment there, children only travel in the attach snapshot, and nothing merges live `rlm_child_update` events into that cache. ACP attaches at startup before any children exist, so the emission-time roster is the empty attach-time seed.

The tests that show a non-zero count all use the hand-rolled `fakeAcpConnection`, whose snapshot conveniently includes `children`; the one harness-based test asserts `outstandingSubagents: 0`, which is what you always get. So the suite proves acp-mode faithfully counts whatever the connection hands it, not that the connection hands it anything real. For the motivating verifiers use case this reports exactly the false "all quiet" the metadata exists to prevent.

Fix direction: populate `children` in `createAgentConnectionSnapshot()` from the runtime's child registry, and for the daemon path either merge `rlm_child_update` events into `latestSnapshot.children` or add a live-children read to the snapshot rebuild. `remainingAutonomousContinuations` is unaffected — it comes from the completion status, not the snapshot. A test with a real connection and a live child would lock this in.

snimu:
One more thing: this adds user-visible ACP surface (`quiescence` in `_meta`, and `session/prompt` now rejecting when the emission-time roster read fails even though the turn completed) but has no CHANGELOG entry. Please add a bullet under `## [Unreleased]` in `packages/coding-agent/CHANGELOG.md`.

packages/coding-agent/src/modes/acp/acp-mode.ts
sethkarten: Addressed in 6447725af. In-process snapshots now read the live recursive AgentSession child registry, and daemon connections merge `rlm_child_update` events into the refreshed snapshot. Quiescence therefore counts post-attach queued/running children on both transports. Regressions cover a real ACP connection with a live child and a daemon snapshot updated from an initially empty roster.

sethkarten:
Addressed in 6447725af: `packages/coding-agent/CHANGELOG.md` now documents ACP `_meta.quiescence` and the prompt rejection behavior when the emission-time live roster read fails.

sethkarten:
Closing this superseded ACP implementation. Resident recovery coverage continues in #1236, with the exact causal prompt lifecycle successor in #1239. This PR remains available as design and review history.

PR #815, opened 2026-08-07
fix(coding-agent): namespace daemon socket by agent dir

### LINKED ISSUE #768: Daemon socket identity ignores PRIME_AGENT_CODING_AGENT_DIR — sessions silently run under another install's daemon

zan22ye:
## Summary

- Namespace the default daemon socket by the resolved Prime Agent config directory.
- Keep the shared per-user socket directory unchanged so worker socket paths do not get longer.
- Add regression coverage for distinct `PRIME_AGENT_CODING_AGENT_DIR` values producing distinct default daemon sockets.

Fixes #768

## Testing

- `npm run check`
- `cd packages/coding-agent && npx tsx ../../node_modules/vitest/dist/cli.js --run test/daemon-socket.test.ts test/daemon-ps.test.ts test/daemon-ps-format.test.ts test/daemon-launch.test.ts`

> [!NOTE]
> ### Namespace daemon socket path by agent directory to prevent supervisor sharing
> Previously, all coding agent instances used a fixed daemon socket name, causing instances with different `PRIME_AGENT_CODING_AGENT_DIR` values to share the same supervisor. The fix adds an 8-hex-character SHA-256 suffix derived from the normalized agent directory path to the socket name. On Windows the pipe becomes `\\.\pipe\prime-agent-daemon-<hash>`; on Unix the socket becomes `daemon-<hash>.sock`.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized b649247.</sup>
> 
>

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [Unreleased]
 
 - Fixed the bundled `websearch` skill description and missing-key guidance omitting the `/login` → **MCP Connections** step required to configure Serper.
+- Fixed default daemon sockets sharing one supervisor across distinct `PRIME_AGENT_CODING_AGENT_DIR` values ([#768](https://github.com/PrimeIntellect-ai/prime-agent/issues/768)).
 
 ## [0.7.0] - 2026-08-05
 
--- a/packages/coding-agent/src/modes/daemon/daemon-socket.ts
+++ b/packages/coding-agent/src/modes/daemon/daemon-socket.ts
@@ -1,8 +1,10 @@
+import { createHash } from "node:crypto";
 import { chmodSync, existsSync, lstatSync, mkdirSync, unlinkSync } from "node:fs";
 import { createConnection } from "node:net";
 import { tmpdir } from "node:os";
-import { dirname, join } from "node:path";
+import { dirname, join, resolve } from "node:path";
 import lockfile from "proper-lockfile";
+import { getAgentDir } from "../../config.js";
 
 const DAEMON_SOCKET_MODE = 0o600;
 const DAEMON_SOCKET_DIR_MODE = 0o700;
@@ -34,10 +36,17 @@ export interface DaemonSocketIdentity {
 }
 
 export function defaultDaemonSocketPath(): string {
+	const suffix = agentDirSocketSuffix();
 	if (process.platform === "win32") {
-		return "\\\\.\\pipe\\prime-agent-daemon";
+		return `\\\\.\\pipe\\prime-agent-daemon-${suffix}`;
 	}
-	return join(defaultDaemonSocketDir(), "daemon.sock");
+	return join(defaultDaemonSocketDir(), `daemon-${suffix}.sock`);
+}
+
+function agentDirSocketSuffix(): string {
+	const agentDir = resolve(getAgentDir());
+	const normalized = process.platform === "win32" ? agentDir.toLowerCase() : agentDir;
+	return createHash("sha256").update(normalized).digest("hex").slice(0, 8);
 }
 
 export async function acquireDaemonSocketPathLease(socketPath: string): Promise<DaemonSocketPathLease | undefined> {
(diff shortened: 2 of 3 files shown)


snimu:
Thanks — the fix works, but #786 landed the same daemon-socket change for #768 first (same files, same approach). Closing as duplicate of #786; if the win32 case-normalization detail matters, please raise it as a comment there.

PR #816, opened 2026-08-07
fix(coding-agent): survive provider outages in auto-retry (jitter, backoff cap, Retry-After)

### LINKED ISSUE #688: Auto-retry gives up after 14s of backoff, with no jitter, no `Retry-After`, and no delay cap
On `main` @ `c22549a` (v0.7.0), `AgentSession._handleRetryableError` ([agent-session.ts#L10123](https://github.com/PrimeIntellect-ai/prime-agent/blob/main/packages/coding-agent/src/core/agent-session.ts#L10123)) computes:
 
```ts
const delayMs = settings.baseDelayMs * 2 ** (this._retryAttempt - 1);   // L10163
```
 
With the defaults `maxRetries: 3, baseDelayMs: 2000` ([settings-manager.ts#L876](https://github.com/PrimeIntellect-ai/prime-agent/blob/main/packages/coding-agent/src/core/settings-manager.ts#L876)) that's `2s, 4s, 8s` — **14s total, then the turn fails.** Provider `529`/`429` episodes routinely last minutes, so a long run dies on a blip it could have waited out.

anyerqi:
Fixes #688.

Session-level auto-retry gave up after 14s of backoff (3 attempts, 2s/4s/8s), with no jitter, no delay cap, and no `Retry-After` handling. Provider 429/529 episodes routinely last minutes, so long runs died on blips they could have waited out.

## Changes

**`packages/ai`** — `extractStreamFailureParts` already reads `err.headers` for the request id, so `Retry-After` extraction rides the existing path: parse `retry-after-ms` (milliseconds) and `retry-after` (delta-seconds or HTTP-date) into a new `StreamFailureInfo.retryAfterMs`. The value flows into the `provider_stream_failure` diagnostic that the session layer already consumes; no new plumbing.

**`packages/coding-agent`** — `_handleRetryableError` now computes:

```
exponential = min(baseDelayMs * 2^(attempt-1), maxBackoffMs)   # cap bounds only our own guesswork
jittered    = random() * exponential                            # full jitter
delay       = max(retryAfterMs, jittered)                       # Retry-After is the floor, honored past the cap
```

A `Retry-After` above `retry.maxRetryAfterMs` (default 5 min, `0` disables) stops the retry loop with an informative error instead of waiting, mirroring the provider layer's `retry.provider.maxRetryDelayMs` fail-fast. The sleep stays abortable (esc/ctrl+c) and `auto_retry_start.delayMs` drives the TUI countdown, so long honored waits stay visible.

Defaults: `maxRetries` 3 → 10, `retry.maxBackoffMs` 60000, `retry.maxRetryAfterMs` 300000. Worst-case backoff-only budget goes from 14s to ~6 min (2+4+8+16+32+60×5, roughly half that in expectation under full jitter), enough to ride out typical provider incidents. Reference points: Claude Code retries up to 10 attempts with backoff reaching ~30-60s and honors server retry guidance; Codex defaults to 4 request + 5 stream retries with jittered exponential backoff and honors server-provided delays. The 60s cap mirrors the existing `retry.provider.maxRetryDelayMs` default.

## Notes for review

- The setting is named `maxBackoffMs` rather than `maxDelayMs` because settings migration moves legacy `retry.maxDelayMs` into `retry.provider.maxRetryDelayMs` and deletes the key; reusing the name would silently migrate away user-set values.
- `auto_retry_start`/`auto_retry_end` event shapes are unchanged, so the TUI countdown and daemon protocol are unaffected. A wall-clock retry budget (`maxElapsedMs`) was considered but rejected for this PR: it changes `attempt/maxAttempts` semantics, which are part of the daemon protocol surface.

## Testing

- `packages/ai/test/stream-failure.test.ts`: header parsing for both `Headers` and plain-record shapes, delta-seconds, HTTP-date, `retry-after-ms` precedence, invalid/past values.
- `packages/coding-agent/test/suite/regressions/688-retry-backoff.test.ts`: new defaults, cap on backoff, full-jitter bounds, `Retry-After` as the floor and honored beyond `maxBackoffMs`, retry loop stopping when `Retry-After` exceeds `maxRetryAfterMs` (fresh and mid-sequence).
- Existing retry suites pass unchanged (`agent-session-retry-events`, regressions 3317/4491, settings-manager).
- `npm run check` clean.


anyerqi:
Good catch — the cap was doing double duty and the comment overpromised. Reworked in f97efc32a:

- `maxBackoffMs` now bounds only the exponential term; `Retry-After` is the delay floor and is honored past the cap.
- For the absurd-value case I took the `retry.maxRetryAfterMs` route (default 5 min, `0` disables): a longer server-requested wait stops the retry loop and appends the reason to the error message, mirroring how `retry.provider.maxRetryDelayMs` already fails fast instead of waiting silently.
- On surfacing: the wait was already user-visible — `auto_retry_start.delayMs` drives the TUI countdown and the sleep is abortable — so honoring long waits hides nothing.
- Code comment, docs table, and the pinning test now state the actual precedence; added coverage for beyond-cap honoring and for the ceiling stop, both fresh and mid-sequence.

BILLKISHORE:
The `Retry-After` extraction and the move to full jitter both look right, and rejecting a past HTTP-date rather than producing a negative delay is a nice touch.

One thing worth reconsidering: the cap is applied on top of `Retry-After`, so the agent can retry earlier than the provider explicitly asked. The comment on `_computeRetryDelay` says the header is a "floor so we never retry before the provider asked, but the cap still wins", and those two clauses disagree whenever `retryAfterMs > cap`. `caps the server Retry-After at maxBackoffMs` pins the behaviour: a provider asking for 5000ms is retried after 40ms.

With the 60s default that is not just a test artifact. Anthropic and OpenAI 429/529 episodes regularly ask for more than a minute, so the common case during a real outage is retrying at 60s against a server that asked for longer. That is the scenario this PR is trying to survive, and retrying early is what turns one 429 into a series of them.

Worth separating the two concerns: the cap exists to bound *our* exponential guesswork, which is unbounded by construction, whereas `Retry-After` is a specific instruction from the only party that knows when capacity returns. Bounding the first does not imply overriding the second.

If the worry is a provider sending an absurd value and hanging the session, a separate and much larger ceiling for header-driven waits, say `retry.maxRetryAfterMs`, would cover that without capping ordinary ones. Alternatively, honoring `Retry-After` as-is and surfacing the wait in the retry event would keep the user informed rather than silently retrying early.

Either way the code comment and the docs table should say which rule actually wins, since they currently read as though the header does.

kevinjosethomas:

PR #827, opened 2026-08-07
fix(ai): repair invalid unicode escape prefixes

reyanshgupta:
## Problem

`repairJson()` validates `\uXXXX` escapes separately, but the generic escape lookup also treated every `\u` prefix as valid. Malformed streamed tool arguments such as `{"path":"C:\users\workspace"}` therefore remained invalid, and `parseStreamingJson()` eventually returned an empty object instead of the tool arguments.

## Fix

- Preserve `\u` escapes only when they contain exactly four hexadecimal digits.
- Escape malformed `\u` prefixes using the existing invalid-escape repair path.
- Add regression coverage for unescaped Windows paths and valid Unicode escapes.
- Record the user-visible fix under the AI package's Unreleased changelog.

## Validation

- `npx tsx ../../node_modules/vitest/dist/cli.js --run test/json-parse.test.ts` in `packages/ai` (3 tests passed)
- `npm run check`

> [!NOTE]
> ### Fix `repairJson` to preserve malformed Unicode escape prefixes in streamed tool-call arguments
> Removes `'u'` from the `VALID_JSON_ESCAPES` lookup in [`json-parse.ts`](https://github.com/PrimeIntellect-ai/prime-agent/pull/827/files#diff-77e0781eb08fb6dbe9c71a2343a1245f2481e3525365953f99dcdd8228affa94) so that `\u` not followed by 4 hex digits is repaired to `\\u` instead of being accepted as a valid escape. This prevents streamed tool-call arguments containing Windows paths (e.g. `C:\Users\...`) from being silently discarded.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 293d582.</sup>
> 
>

### DIFF
--- a/packages/ai/CHANGELOG.md
+++ b/packages/ai/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## [Unreleased]
 
+- Fixed malformed Unicode escape prefixes causing streamed tool-call arguments such as Windows paths to be discarded.
+
 ## [0.7.0] - 2026-08-05
 
 ## [0.6.1] - 2026-08-05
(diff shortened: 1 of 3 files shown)


kevinjosethomas:

PR #833, opened 2026-08-07
Add per-spawn service tier selection to RLM

paralin:
Allow each RLM spawn to request a service tier with `auto`, `default`, `flex`, `scale`, `priority`, or `None`. Omitted requests continue to inherit the parent tier, and priority requests retain the existing fast-mode clamp for unsupported child models.

Gate every explicit override with the new `rlmAllowedServiceTiers` settings array. When the array is unset, its effective value contains only `defaultServiceTier`, which defaults to `default`, so costly tiers such as `priority` require explicit user opt-in in the settings file.

Validate the option in both the Python runtime shim and the TypeScript host so invalid cross-boundary requests fail clearly. Document the configuration and call shape, and cover selection, inheritance, allowlist defaults, rejection, clamping, and forwarding.

> [!NOTE]
> ### Add per-spawn `service_tier` selection to RLM subagent calls
> - `rlm.run` now accepts a `service_tier` kwarg (`'auto'`, `'default'`, `'flex'`, `'scale'`, `'priority'`); invalid values are rejected immediately in both the Python runtime and the TypeScript agent session.
> - Allowed tiers are enforced via a new `rlmAllowedServiceTiers` setting; unset defaults to `[defaultServiceTier]`. A new `SettingsManager.getRlmAllowedServiceTiers` method exposes the effective allowlist.
> - If `service_tier` is omitted, the child run inherits the parent session's tier. If `'priority'` is requested for a model without fast-mode support, the tier is clamped to `'default'`.
> - The RLM system prompt is updated to document valid tier values and allowlist behavior.
> - Behavioral Change: `rlm.run` calls specifying a tier outside `rlmAllowedServiceTiers` now throw/raise instead of proceeding.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 2940377.</sup>
> 
>

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [Unreleased]
 
 - Fixed URLs not opening on click in fullscreen mode on terminals such as Ghostty; clicking a link in the transcript, dock, or overlays now opens it in the browser.
+- Added config-gated per-spawn RLM service-tier selection with validation, inheritance, and fast-mode clamping.
 
 ## [0.7.2] - 2026-08-11
 
--- a/packages/coding-agent/docs/rlm-runtime.md
+++ b/packages/coding-agent/docs/rlm-runtime.md
@@ -23,7 +23,7 @@ flowchart TD
 When the model delegates work:
 
 ```python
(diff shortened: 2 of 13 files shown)


kevinjosethomas:

PR #834, opened 2026-08-07
fix(coding-agent): preserve harness state on interrupted writes

ssynb:
## Summary

- write Python-side continual harness state through a same-directory temporary file
- atomically replace the state file only after JSON serialization succeeds
- preserve the previous valid state when serialization is interrupted

## Root cause

`HarnessState.save()` opened `harness_state.json` with `"w"` before JSON serialization completed. An exception or interruption during `json.dump()` therefore truncated the last valid file. Because corrupt harness JSON is intentionally loaded as empty state, a later save could turn that transient failure into persistent memory loss.

## User impact

Interrupted kernel-side memory or refinement writes now leave the last valid continual harness state intact instead of potentially losing all persisted entries.

## Validation

- `uv run --quiet --isolated --no-project --python 3.12 --with ipykernel --with nest-asyncio --with tyro --with mcp --with httpx python -m unittest discover -s test -p 'test_*.py'` (65 tests)
- `npm run check` (Biome, TypeScript, installer render, browser smoke)
- deterministic interrupted-serialization reproduction: old state remained byte-for-byte unchanged

> [!NOTE]
> ### Fix `HarnessState.save` to preserve state file on interrupted writes
> - Replaces direct JSON writes in [`harness.py`](https://github.com/PrimeIntellect-ai/prime-agent/pull/834/files#diff-441d64dfc5d74f553c86b609beff62210693b468d021d8bded095dea768138cf) with an atomic write: JSON is written to a `NamedTemporaryFile` in the same directory, then swapped into place via `os.replace`.
> - If the write fails mid-way, the original state file is left untouched and the temp file is cleaned up in a `finally` block.
> - When overwriting an existing file, its permission bits are copied to the temp file before replacement.
> - Adds a test that patches `json.dump` to raise `OSError` mid-write and asserts the prior state is preserved.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 36b0b63.</sup>
> 
>

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [Unreleased]
 
 - Fixed the bundled `websearch` skill description and missing-key guidance omitting the `/login` → **MCP Connections** step required to configure Serper.
+- Fixed interrupted Python harness writes truncating the last valid continual harness state ([#834](https://github.com/PrimeIntellect-ai/prime-agent/pull/834) by [@ssynb](https://github.com/ssynb)).
 
 ## [0.7.0] - 2026-08-05
 
--- a/prime-agent-runtime/src/rlm/harness.py
+++ b/prime-agent-runtime/src/rlm/harness.py
@@ -10,6 +10,7 @@ from __future__ import annotations
 
 import json
 import os
+import tempfile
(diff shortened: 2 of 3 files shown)


kevinjosethomas:

PR #837, opened 2026-08-07
fix(ai): follow RFC 9728 protected-resource metadata in MCP OAuth discovery

### LINKED ISSUE #766: [coding-agent] MCP OAuth discovery ignores RFC 9728 protected-resource metadata; 17 of 36 hosted servers cannot be logged into
### Summary

MCP OAuth discovery only looks for authorization-server metadata at the **MCP endpoint's own
origin**. Per [RFC 9728](https://datatracker.ietf.org/doc/html/rfc9728) (and the MCP authorization
spec, which normatively requires it), a server advertises its authorization server via
*protected-resource* metadata, commonly on a different host. Prime Agent never reads that
document, so `/mcp login <server>` is impossible for any server whose authorization server isn't
co-located with its MCP endpoint.

Measured against the 36 hosted MCP servers our org has provisioned: **17 fail discovery**, all of
them resolvable from the protected-resource document that the servers do publish.

### Environment

- prime-agent **0.7.0** (also reviewed `main`)
- macOS 15 (arm64)

### The code

[`packages/ai/src/mcp/oauth.ts:66-88`](https://github.com/PrimeIntellect-ai/prime-agent/blob/main/packages/ai/src/mcp/oauth.ts#L66-L88):

```ts
/** Try the protected-resource and auth-server well-known docs at the URL's origin. */
async function discover(url: string): Promise<AuthServerMetadata> {
	const origin = new URL(url).origin;
	const candidates = [
		`${origin}/.well-known/oauth-authorization-server`,
		`${origin}/.well-known/openid-configuration`,
	];
	...
}
```

The doc comment says protected-resource metadata is tried; the code never requests it.
`oauth-protected-resource` and `resource_metadata` appear nowhere in the repository.

Three things are missed:

1. **`authorization_servers` indirection**. The AS is frequently a different host.
2. **The `WWW-Authenticate` hint**. An unauthenticated `initialize` returns
   `401 Bearer resource_metadata="https://…/.well-known/oauth-protected-resource/…"`, the exact
   pointer RFC 9728 defines. It is discarded.
3. **Path-suffixed well-knowns**. RFC 8414/9728 place metadata at
   `/.well-known/oauth-protected-resource{path}`, not only at the bare root. For an MCP URL with a
   path (`/mcp/v1`), the root-only probe can 404 while the path-suffixed one is 200.

### Reproduction

Todoist's hosted MCP (`https://ai.todoist.net/mcp`) is a public example:

```
404  https://ai.todoist.net/.well-known/oauth-authorization-server     <- tried by discover()
404  https://ai.todoist.net/.well-known/openid-configuration           <- tried by discover()
200  https://ai.todoist.net/.well-known/oauth-protected-resource/mcp   <- not tried
     -> {"authorization_servers": ["https://todoist.com"]}
401  POST https://ai.todoist.net/mcp
     WWW-Authenticate: Bearer resource_metadata="https://ai.todoist.net/.well-known/oauth-protected-resource/mcp"
```

`discover()` throws `Could not discover OAuth metadata for https://ai.todoist.net`, even though
`https://todoist.com/.well-known/oauth-authorization-server` resolves and advertises
`registration_endpoint: https://todoist.com/oauth/register`; this server would otherwise work
end-to-end with the existing DCR + PKCE flow.

Google's hosted MCP servers are the same shape with a path-suffixed document:

```
404  https://drivemcp.googleapis.com/.well-known/oauth-protected-resource
200  https://drivemcp.googleapis.com/.well-known/oauth-protected-resource/mcp/v1
     -> {"authorization_servers": ["https://accounts.google.com/"]}
```

A related failure mode: some origins answer the well-known path with a `200` HTML SPA catch-all
(Ashby's `https://mcp.ashbyhq.com/.well-known/oauth-authorization-server` returns
`text/html`). `fetchJson` throws on the parse, the loop swallows it, and the final message blames
"Could not discover OAuth metadata" without mentioning that a non-JSON `200` was received. That
server's protected-resource document cleanly points at `https://mcp-auth.ashbyhq.com/oidc`.

### Full measurement

MCP endpoint → real authorization server, for the servers where `discover()` currently throws:

| MCP endpoint | `authorization_servers` |
|---|---|
| `https://ai.todoist.net/mcp` | `https://todoist.com` |
| `https://drivemcp.googleapis.com/mcp/v1` | `https://accounts.google.com/` |
| `https://gmailmcp.googleapis.com/mcp/v1` | `https://accounts.google.com/` |
| `https://calendarmcp.googleapis.com/mcp/v1` | `https://accounts.google.com/` |
| `https://mcp.box.com` | `https://api.box.com/` |
| `https://mcp.ashbyhq.com/mcp/v1` | `https://mcp-auth.ashbyhq.com/oidc` |
| `https://mcp.metaview.ai/mcp` | `https://auth.metaview.ai/` |
| `https://mcp.na1.ironcladapp.com/mcp` | `https://ironcladapp.com` |
| `https://mcp.gowindmill.com/mcp` | `https://api.gowindmill.com` |
| `https://mcp.zip.com/mcp` | `https://app.ziphq.com` |
| `https://mcp.shippo.com` | `https://goshippo.com` |
| `https://mcp.mail.superhuman.com/mcp` | `https://mcp.auth.mail.superhuman.com` |
| `https://app.tryordinal.com/mcp` | `https://mletdxmpsjlbfpnymbcu.supabase.co/auth/v1` |
| `https://<workspace>.cloud.databricks.com/api/2.0/mcp/sql` | `https://<workspace>.cloud.databricks.com/oidc` |
| `https://mcp.mixpanel.com/mcp` | *(no PRM published — genuinely undiscoverable)* |
| `https://8889.apim.mcp.workato.com` | *(no PRM published)* |

(One internal server omitted.) Breakdown of the 16:

- **8 need nothing but this fix**: Todoist, Ashby, Metaview, Windmill, ZipHQ, Shippo, Superhuman,
  Ordinal. Their authorization servers advertise a `registration_endpoint`, so once discovery finds
  them the existing DCR + PKCE flow completes.
- **6 become discoverable but still need a pre-registered client id**: Google Drive / Gmail /
  Calendar, Box, Ironclad, Databricks. (See the companion issue on `mcpServers` not accepting a
  `clientId`.)
- **2 publish no protected-resource document at all**: Mixpanel, Workato. Genuinely undiscoverable;
  the only improvement available is a clearer error.

### Proposal

In `discover()`, follow RFC 9728 before falling back to today's behaviour:

1. If available, use the `resource_metadata` URL from the `WWW-Authenticate` header of a
   `401`/`WWW-Authenticate`-bearing response. (`login()` would need to make, or be handed, one
   unauthenticated probe request; alternatively skip straight to step 2.)
2. `GET {origin}/.well-known/oauth-protected-resource{path}`, then
   `GET {origin}/.well-known/oauth-protected-resource`.
3. Take `authorization_servers[0]` as the issuer and fetch its metadata at both the path-suffixed
   and root forms of `/.well-known/oauth-authorization-server` and `/.well-known/openid-configuration`
   (RFC 8414 defines the path-suffixed form for issuers with a path, e.g.
   `https://host/oidc` → `https://host/.well-known/oauth-authorization-server/oidc`).
4. Fall back to today's two origin candidates.
5. Validate `content-type`/parse separately from transport errors so the thrown message can say
   "returned non-JSON" rather than a bare "could not discover".

Optionally record the resolved issuer on the credential (alongside the existing `tokenEndpoint` /
`clientId`) so refresh never has to re-discover.

Happy to send a PR with unit tests over recorded fixtures for the metadata documents above.

yurekami:
Fixes #766.

## The defect

`discover()` probed only two documents, both at the MCP endpoint's own origin:

```ts
/** Try the protected-resource and auth-server well-known docs at the URL's origin. */
async function discover(url: string): Promise<AuthServerMetadata> {
	const origin = new URL(url).origin;
	const candidates = [
		`${origin}/.well-known/oauth-authorization-server`,
		`${origin}/.well-known/openid-configuration`,
	];
```

The doc comment names protected-resource metadata; `oauth-protected-resource` appears nowhere in the tree at `b9a4461`. RFC 9728 makes that document the resource's own statement of which authorization server to use, and it routinely names a different host, so `/mcp login` was impossible for those servers.

## The change

1. Fetch the protected-resource document first: path-suffixed (`/.well-known/oauth-protected-resource/mcp/v1`), then bare root. Both are needed — see ironclad below.
2. Resolve each advertised issuer with the RFC 8414 path-insertion form (`https://host/.well-known/oauth-authorization-server/oidc`) and the OpenID Connect Discovery path-append form.
3. Fall back to today's two origin-rooted probes, unchanged, for servers publishing no such document.
4. Split `fetchJson`'s transport failure from its JSON-parse failure.

`refreshToken()` inherits this through its existing `discover()` call; no call site changed. Point 4 is not cosmetic: Ashby returns **HTTP 200 with an HTML sign-in page** at `/.well-known/oauth-authorization-server`, so today's failure is a bare `SyntaxError: Unexpected token '<'` with no indication of which document was bad.

## Evidence

I ran the real `createMcpOAuthProvider().login()` against production servers with `fetch` restricted to `GET /.well-known/` requests, so discovery ran end to end and the DCR/token requests were blocked. Same harness on both sides.

Before, at `b9a4461`:

| Endpoint | origin `oauth-authorization-server` | origin `openid-configuration` | Result |
|---|---|---|---|
| `https://ai.todoist.net/mcp` | 404 | 404 | fails |
| `https://mcp.ashbyhq.com/mcp/v1` | 200 (HTML) | 200 (HTML) | fails, `SyntaxError` |
| `https://mcp.box.com` | 401 | 401 | fails |
| `https://drivemcp.googleapis.com/mcp/v1` | 404 | 404 | fails |
| `https://mcp.na1.ironcladapp.com/mcp` | 404 | 404 | fails |
| `https://mcp.shippo.com` | 403 | 403 | fails |
| `https://mcp.linear.app/mcp` | 200 | — | resolves |

After:

| Endpoint | Protected-resource document | Authorization server resolved |
|---|---|---|
| `https://ai.todoist.net/mcp` | `…/oauth-protected-resource/mcp` | `https://todoist.com` (DCR available) |
| `https://mcp.ashbyhq.com/mcp/v1` | `…/oauth-protected-resource/mcp/v1` | `https://mcp-auth.ashbyhq.com/oidc` (DCR available) |
| `https://mcp.box.com` | `…/oauth-protected-resource` | `https://api.box.com/` (no DCR) |
| `https://drivemcp.googleapis.com/mcp/v1` | `…/oauth-protected-resource/mcp/v1` | `https://accounts.google.com/` (no DCR) |
| `https://mcp.na1.ironcladapp.com/mcp` | suffixed 404, then bare root 200 | `https://ironcladapp.com` (no DCR) |
| `https://mcp.shippo.com` | `…/oauth-protected-resource` | `https://goshippo.com` (DCR available) |
| `https://mcp.linear.app/mcp` | `…/oauth-protected-resource/mcp` | `https://mcp.linear.app` (unchanged) |

Ashby, Box and Google exercise the three edge cases the helpers exist for: an issuer carrying a path, an issuer advertised with a trailing slash, and a path-suffixed document where the bare root 404s. Ironclad is the reverse — its suffixed document 404s and only the bare root answers.

Three of these discover successfully but advertise no `registration_endpoint`, so they still need a pre-registered client id. That is the separate gap the reporter notes; this PR does not address it.

## Tests

Five cases added to `packages/ai/test/mcp-oauth.test.ts` (10 pass, 5 pre-existing unchanged). I mutation-checked them — each mutant is caught by exactly one test, and each passes at HEAD only when it should:

| Mutation | Caught by |
|---|---|
| drop the bare-root protected-resource fallback | `falls back to the bare protected-resource document…` |
| probe the endpoint origin before the advertised issuer | `resolves the authorization server from protected-resource metadata` |
| drop the RFC 8414 path-insertion form | `resolves an issuer that carries a path` |
| revert the `fetchJson` non-JSON split | `reports a non-JSON metadata document distinctly…` |

The precedence test serves a valid authorization-server document at the endpoint origin too, so it pins that the protected-resource document wins rather than merely that the origin probe failed.

`npm run check` is clean.

## Verification

- [x] Quoted code and the `oauth-protected-resource` absence checked against `main` @ `b9a4461`.
- [x] Before/after table produced by running this branch and `b9a4461` through the same harness against live servers.
- [x] `npm run check`; `packages/ai/test/mcp-oauth.test.ts` 10/10.
- [ ] **No login completed end to end.** I have accounts on none of these services, so the harness stops at client registration. What is verified is that discovery resolves an authorization server advertising `authorization_endpoint`, `token_endpoint`, and `S256` among its `code_challenge_methods_supported` — not that the subsequent DCR + PKCE exchange succeeds.
- [ ] The "17 of 36" figure is the reporter's measurement, not mine. I verified the seven endpoints above.

## Deliberately out of scope

- **`WWW-Authenticate` `resource_metadata` (step 1 of the proposal).** Needs an extra unauthenticated probe and reaches into `mcp-manager.ts`. Every server above resolves without it.
- **Recording the resolved issuer on the credential.** Changes the persisted `McpCredentials` shape.
- **Validating that the authorization-server document's `issuer` matches the advertised one** (RFC 9728 SHOULD). Not added because a strict comparison would reject servers that work: Box advertises `https://api.box.com/` and its document reports `https://api.box.com`, and Google advertises `https://accounts.google.com/` against a reported `https://accounts.google.com`. Happy to add it with normalization if you want it here rather than separately.

One note on the fan-out: every advertised issuer is tried in order rather than only `authorization_servers[0]`, with no cap. Say the word if you would rather bound it.

> [!NOTE]
> ### Fix MCP OAuth discovery to follow RFC 9728 protected-resource metadata for cross-host authorization servers
> - The `discover` function in [`packages/ai/src/mcp/oauth.ts`](https://github.com/PrimeIntellect-ai/prime-agent/pull/837/files#diff-a147fd0fbdc02e28e95472286bcb945e037e1a21e97e9083ad14a6a3a06524a9) now fetches RFC 9728 protected-resource metadata from the MCP endpoint first, reads listed `authorization_servers` issuers, and builds candidate discovery URLs using both RFC 8414 and OIDC discovery patterns before falling back to the endpoint's own origin.
> - Adds support for issuers with path components, path-suffixed well-known URLs, and deduplication of candidate URLs; error messages now enumerate each attempted URL with its failure reason.
> - `fetchJson` is updated to throw a distinct error when a well-known endpoint returns non-JSON content, and includes response body text for non-OK responses.
> - New tests cover cross-host authorization servers, path-based issuers, fallback from path-suffixed to bare protected-resource documents, and non-JSON metadata error reporting.
> - Behavioral Change: discovery now prefers the endpoint's protected-resource metadata over the endpoint origin, which changes resolution order for existing MCP endpoints that serve a protected-resource document.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized b1c38eb.</sup>
> 
>

### DIFF
--- a/packages/ai/CHANGELOG.md
+++ b/packages/ai/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## [Unreleased]
 
+- Fixed MCP OAuth discovery to follow RFC 9728 protected-resource metadata, so servers whose authorization server is on another host can be logged into ([#766](https://github.com/PrimeIntellect-ai/prime-agent/issues/766)).
+
 ## [0.7.0] - 2026-08-05
 
 ## [0.6.1] - 2026-08-05
--- a/packages/ai/src/mcp/oauth.ts
+++ b/packages/ai/src/mcp/oauth.ts
@@ -26,12 +26,17 @@ interface AuthServerMetadata {
 	scopes_supported?: string[];
 }
 
+/** Protected-resource metadata (RFC 9728). The resource names its authorization servers. */
+interface ProtectedResourceMetadata {
+	authorization_servers?: string[];
+}
+
 export interface McpOAuthConfig {
 	/** MCP server name; provider id becomes `mcp:<server>`. */
 	server: string;
 	/** Human label for UI. */
 	label?: string;
-	/** The MCP endpoint URL — discovery is rooted at its origin. */
+	/** The MCP endpoint URL — discovery starts from its protected-resource metadata. */
 	url: string;
 	/** Pre-registered client id (servers without DCR, e.g. Slack). */
 	clientId?: string;
@@ -47,10 +52,19 @@ interface McpCredentials extends OAuthCredentials {
 
 async function fetchJson(url: string, init?: RequestInit): Promise<unknown> {
 	const res = await fetch(url, init);
+	const body = await res.text();
 	if (!res.ok) {
-		throw new Error(`${init?.method ?? "GET"} ${url} failed: ${res.status} ${await res.text()}`);
+		throw new Error(`${init?.method ?? "GET"} ${url} failed: ${res.status} ${body}`);
 	}
-	return res.json();
+	try {
+		return JSON.parse(body);
+	} catch {
+		// A sign-in page served at a well-known URI is a different failure from a 404,
+		// and saying so is the difference between a fixable report and "could not discover".
+		throw new Error(
+			`${init?.method ?? "GET"} ${url} returned non-JSON (content-type ${res.headers.get("content-type") ?? "unset"})`,
+		);
+	}
 }
 
 /** Random, URL-safe CSRF `state` value, independent of the PKCE verifier. */
@@ -63,28 +77,76 @@ function randomState(): string {
 		.replace(/=/g, "");
 }
 
-/** Try the protected-resource and auth-server well-known docs at the URL's origin. */
+/** Path a well-known URI is suffixed with, normalized so a bare origin contributes nothing. */
+function wellKnownSuffix(url: URL): string {
+	return url.pathname.replace(/\/+$/, "");
+}
+
+/** Protected-resource metadata locations for an MCP endpoint (RFC 9728 section 3.1). */
+function protectedResourceUrls(endpoint: URL): string[] {
+	const base = `${endpoint.origin}/.well-known/oauth-protected-resource`;
+	const suffix = wellKnownSuffix(endpoint);
+	return suffix ? [`${base}${suffix}`, base] : [base];
+}
+
+/**
+ * Authorization-server metadata locations for an issuer. RFC 8414 inserts the well-known
+ * segment between host and path; OpenID Connect Discovery appends it to the issuer instead.
+ */
+function authServerUrls(issuer: string): string[] {
+	let url: URL;
+	try {
+		url = new URL(issuer);
+	} catch {
+		return [];
+	}
+	const oauth = `${url.origin}/.well-known/oauth-authorization-server`;
+	const oidc = `${url.origin}/.well-known/openid-configuration`;
+	const suffix = wellKnownSuffix(url);
+	if (!suffix) return [oauth, oidc];
+	return [`${oauth}${suffix}`, `${oidc}${suffix}`, `${url.origin}${suffix}/.well-known/openid-configuration`];
+}
+
+/**
+ * Resolve the authorization server for an MCP endpoint. The endpoint's protected-resource
+ * document is authoritative and routinely names a different host, so it is consulted first;
+ * the endpoint's own origin remains the fallback for servers that publish no such document.
+ */
 async function discover(url: string): Promise<AuthServerMetadata> {
-	const origin = new URL(url).origin;
-	const candidates = [
-		`${origin}/.well-known/oauth-authorization-server`,
-		`${origin}/.well-known/openid-configuration`,
-	];
-	let lastError: unknown;
-	for (const candidate of candidates) {
+	const endpoint = new URL(url);
+	const failures: string[] = [];
+	const attempt = async <T>(candidate: string): Promise<T | undefined> => {
 		try {
-			const meta = (await fetchJson(candidate)) as AuthServerMetadata;
-			if (meta.authorization_endpoint && meta.token_endpoint) {
-				return meta;
-			}
+			return (await fetchJson(candidate)) as T;
 		} catch (error) {
-			lastError = error;
+			failures.push(`${candidate} (${error instanceof Error ? error.message : String(error)})`);
+			return undefined;
+		}
+	};
+
+	const issuers: string[] = [];
+	for (const candidate of protectedResourceUrls(endpoint)) {
+		const meta = await attempt<ProtectedResourceMetadata>(candidate);
+		// The document is server-controlled, so a malformed field must not cost us the origin fallback.
+		const listed = meta?.authorization_servers;
+		const advertised = Array.isArray(listed) ? listed.filter((issuer) => typeof issuer === "string" && issuer) : [];
+		if (advertised.length) {
+			issuers.push(...advertised);
+			break;
 		}
 	}
-	throw new Error(
-		`Could not discover OAuth metadata for ${origin}. ` +
-			`Tried ${candidates.join(", ")}. Last error: ${String(lastError)}`,
-	);
+
+	const seen = new Set<string>();
+	for (const candidate of [...issuers.flatMap(authServerUrls), ...authServerUrls(endpoint.origin)]) {
+		if (seen.has(candidate)) continue;
+		seen.add(candidate);
+		const meta = await attempt<AuthServerMetadata>(candidate);
+		if (!meta) continue;
+		if (meta.authorization_endpoint && meta.token_endpoint) return meta;
+		failures.push(`${candidate} (no authorization_endpoint or token_endpoint)`);
+	}
+
+	throw new Error(`Could not discover OAuth metadata for ${endpoint.origin}. Tried ${failures.join("; ")}`);
 }
 
 /** Dynamic client registration (RFC 7591). Returns the issued client_id. */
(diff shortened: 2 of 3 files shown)


sethkarten:
Thank you for the report and proposed work. This root cause is now covered by maintainer-owned stacked PR #1164, authored independently from `upstream/main`.

We did not inspect or reuse this PR's diff, branch, commits, implementation code, or tests; its public description/comments were used only as a bug report. To keep one review surface, this PR is superseded by #1164 and is being closed.

PR #840, opened 2026-08-07
feat(coding-agent): add session-only model switch

paralin:
The `/model` command changes the active model and saves that selection as the configured default. This adds `/switch`, which uses the same exact-match lookup, autocomplete, provider authentication, and selector UI while changing only the current session model.

The persistence choice now reaches the real session backend rather than suppressing only InteractiveMode's duplicate settings write. In-process and daemon connections carry `persistDefault`; `AgentSession.setModel` still records the session's `model_change`, but skips the global settings update for `/switch`. `/model` keeps its existing behavior.

Daemon clients capability-gate session-only model selection so an older daemon fails clearly instead of silently persisting it. A real-session regression verifies that `/switch` leaves the configured default unchanged, alongside focused command, connection, protocol, and UI tests.

This addresses the `/switch` persistence defect reported in #1248. That issue's separate mid-run SDK/extension loop-switching proposal is outside this PR.

Verification: 6 focused test files, 255 tests passed; `npm run check` passed through the pre-commit hook.

> [!NOTE]
> ### Add `/switch` slash command for session-only model selection without persisting the default
> - Adds a new `/switch` slash command alongside the existing `/model` command; `/model` saves the selection as the configured default, `/switch` changes the model for the current session only.
> - Threads a `persistDefault` flag through `AgentSession.setModel`, `AgentConnection.setModel`, and the interactive mode model selection flow to control whether the global default is updated.
> - Bumps `DAEMON_SCHEMA_REVISION` to 17 and adds a `session_model_selection` server capability; daemon connections throw `DaemonCapabilityUnavailableError` if session-only selection is requested against a server that lacks this capability.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 77ea195.</sup>
> 
>

### DIFF
--- a/packages/coding-agent/src/core/agent-session.ts
+++ b/packages/coding-agent/src/core/agent-session.ts
@@ -904,6 +904,7 @@ export interface ModelCycleResult {
 
 interface ModelSelectOptions {
 	waitForExtensions?: boolean;
+	persistDefault?: boolean;
 }
 
 interface ToolDefinitionEntry {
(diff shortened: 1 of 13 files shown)


kevinjosethomas:

PR #842, opened 2026-08-07
fix(coding-agent): run %%bash cells in Git Bash on Windows, not WSL bash

FiveTechSoft:
## Problem

On Windows, a bare `%%bash` cell in the IPython kernel resolves `bash` through CreateProcess, which finds `C:\Windows\System32\bash.exe` (**WSL**) before anything on PATH — even when Git Bash is installed and earlier in PATH.

Inside WSL there is no Git Credential Manager, so `git push` (or any command that prompts) blocks on a username/password prompt that the kernel can never answer — the kernel runs with `allow_stdin: false` and no TTY. The cell hangs until the user aborts, and the kernel can die with it. Observed in a real session as a **3-minute hang on `git push`**, with commits authored as `root@<host>.localdomain` (the WSL default identity).

## Fix

- **Bare `%%bash` cells are rewritten to `%%script <Git Bash>` by default on Windows** via a new `getWindowsIpythonBashScriptPath()` helper (`utils/shell.ts`). An explicit `shellPath` setting still wins. Git Bash is located in the standard install dirs or derived from `git.exe` on PATH.
- The resolved path is converted to its **8.3 short path with forward slashes** (`C:/PROGRA~1/Git/bin/bash.exe`). This is required because IPython parses `%%script` lines on Windows with shlex in non-posix mode, which keeps quotes literally and mangles backslash escapes — verified against the installed IPython: only a space-free, forward-slash path survives parsing (`quoteScriptMagicArgument` now also keeps `~` unquoted for this).
- **The kernel env now defaults to `GIT_TERMINAL_PROMPT=0`** (`kernel/index.ts`): when no credential helper can answer, git fails fast with a clear error instead of hanging the cell forever. An explicit host or per-kernel value still takes precedence.

## Verification

- 3 new tests in `ipython-provisioner.test.ts` (Windows default applied, explicit `shellPath` wins, no Git Bash → cell untouched). They fail against the old code and pass with the fix.
- End-to-end simulation of the script magic with the real kernel venv (`arg_split` + `create_subprocess_exec` + cell via stdin): runs Git Bash (`/mingw64/bin/git`, `credential.helper=manager`) and a `git ls-remote` to a nonexistent repo authenticates via GCM and fails fast in ~2s — no hang.
- `npm run check` (biome + tsgo + installer + browser-smoke) passes. The 6 remaining test failures on Windows (fake-python sh stubs, chmod-based EACCES) are pre-existing on main and unrelated.

> [!NOTE]
> ### Fix `%%bash` cells to use Git Bash on Windows instead of WSL bash
> - On Windows, bare `%%bash` cells (with no explicit shell arguments) are rewritten to `%%script` invocations using a resolved Git Bash path via the new `getWindowsIpythonBashScriptPath()` helper in [`utils/shell.ts`](https://github.com/PrimeIntellect-ai/prime-agent/pull/842/files#diff-194f91e121efc11aa3db177a21b23875cc1356e8e41a4464533c0da0b9b8e0fe).
> - Git Bash discovery scans standard Program Files locations and also checks for `bash.exe` relative to any `git.exe` found on PATH, replacing previously hardcoded paths.
> - Windows paths with spaces are converted to 8.3 short paths (via PowerShell/COM) to avoid breaking IPython's `%%script` argument parser; `~` is also left unquoted to prevent corruption of Windows short paths.
> - Kernel child processes now default to `GIT_TERMINAL_PROMPT=0` across both forked and directly spawned code paths.
> - Behavioral Change: on Windows, `%%bash` cells will run in Git Bash instead of WSL bash when Git Bash is available; cells are left unchanged if no suitable path is found.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 3bbe038.</sup>
> 
>

### DIFF
--- a/packages/coding-agent/src/core/kernel/index.ts
+++ b/packages/coding-agent/src/core/kernel/index.ts
@@ -233,6 +233,16 @@ function createKernelStartupAbortError(): Error {
 	return new Error("Kernel startup aborted");
 }
 
+/**
+ * Kernel process env. The kernel has no interactive terminal, so a git
+ * credential prompt could never be answered and would hang the cell until the
+ * user aborts — default GIT_TERMINAL_PROMPT=0 so git fails fast instead. An
+ * explicit GIT_TERMINAL_PROMPT in the host env or per-kernel overrides wins.
+ */
+function buildKernelEnv(overrides?: Record<string, string>): NodeJS.ProcessEnv {
+	return { GIT_TERMINAL_PROMPT: "0", ...process.env, ...overrides };
+}
+
 function raceStartupWithAbort<T>(promise: Promise<T>, signal: AbortSignal | undefined): Promise<T> {
 	if (!signal) {
 		return promise;
@@ -622,7 +632,7 @@ export class KernelManager {
 					// Match the direct-spawn env exactly: merge the current host env with
 					// the per-kernel overrides, applied fresh in the child (the template's
 					// inherited env snapshot may be stale by fork time).
-					env: this.options.env ? { ...process.env, ...this.options.env } : { ...process.env },
+					env: buildKernelEnv(this.options.env),
 				});
 				forked = true;
 			} catch (err) {
@@ -646,7 +656,7 @@ export class KernelManager {
 		if (!forked) {
 			const kernel = spawn(python, ["-m", "ipykernel_launcher", "-f", connection.path], {
 				cwd: this.options.cwd,
-				env: this.options.env ? { ...process.env, ...this.options.env } : process.env,
+				env: buildKernelEnv(this.options.env),
 				stdio: ["ignore", "pipe", "pipe"],
 			});
 			this.kernel = kernel;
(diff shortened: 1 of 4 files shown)


khanra17:
Validated the integration on Windows 11 x64 with Git for Windows:

- Bare `%%bash` cells resolve to `C:/PROGRA~1/Git/bin/bash.exe`, so they do not fall through to WSL's `System32\\bash.exe`.
- An explicit `shellPath` still wins.
- `GIT_TERMINAL_PROMPT=0` makes headless Git credential failures return instead of hanging the kernel.

When combined with #744, the IPython path should remain the public Git Bash launcher, while daemon Job Object execution should retain #744's direct `usr\\bin\\bash.exe` path. The short-path conversion and `git.exe`-relative discovery are integration details; I am not opening a separate PR for them.

kevinjosethomas:

PR #854, opened 2026-08-07
fix(runtime): accept RLMSpawnHandle in delete_subagent and align name attributes

### LINKED ISSUE #824: [rlm] RLMSpawnHandle.name vs RLMSubagent.session_name: rlm.delete_subagent(handle) raises TypeError and RLMSubagent.name raises AttributeError
## Summary

The Python `rlm` API names the same concept two different ways depending on where the object came from. `rlm()` / `rlm.run()` return an `RLMSpawnHandle` whose child name field is `name`, while `rlm.list_subagents()` returns `RLMSubagent` objects whose child name field is `session_name`. Neither type carries the other's attribute, and `delete_subagent()` accepts only `str | RLMSubagent`, so the natural call `await rlm.delete_subagent(handle)` on the handle you just received from `rlm(...)` raises `TypeError`. The shipped prompts and docs teach both spellings without ever noting they belong to different types.

## Environment

- prime-agent 0.7.0, installed from npm
- Component: `prime-agent-runtime` (Python kernel package `rlm`), plus the coding-agent prompts/docs that document it
- Verified against `main` at commit `b9a4461149419156599d60174dddf15458e2b9ee`. The `rlm/__init__.py` shipped in the 0.7.0 npm package is byte-identical to that commit, so the line numbers below apply to both.

## Reproduction

From the IPython kernel of any prime-agent session:

1. Spawn a child and keep the handle:

```python
handle = await rlm("Review the public API", name="api-reviewer")
print(handle.name)              # "api-reviewer" - works
print(handle.session_name)      # AttributeError
```

2. Try to delete the child using the handle you were just given:

```python
await rlm.delete_subagent(handle)
# TypeError: target must be str or RLMSubagent, got RLMSpawnHandle
```

3. Now go the other direction - recover the child from the registry and try the attribute the spawn-time docs taught:

```python
children = await rlm.list_subagents()
sub = children[0]
print(sub.session_name)         # "api-reviewer" - works
print(sub.name)                 # AttributeError: 'RLMSubagent' object has no attribute 'name'
await rlm.delete_subagent(sub)  # works
```

The `TypeError` in step 2 is raised by an `isinstance` check before any host request is issued, so the attribute and type behaviour can also be confirmed with no daemon and no live child by constructing the two dataclasses directly:

```python
from pathlib import Path
from rlm import RLMSpawnHandle, RLMSubagent, delete_subagent

h = RLMSpawnHandle(rlm_child_id="c1", name="api-reviewer", session_dir=Path("/x"), model="m")
s = RLMSubagent(rlm_child_id="c1", active_session_id="a", session_id="s",
                session_name="api-reviewer", session_dir=Path("/x"), status="running")

hasattr(h, "session_name")   # False
hasattr(s, "name")           # False
await delete_subagent(h)     # TypeError: target must be str or RLMSubagent, got RLMSpawnHandle
```

The exception strings quoted above are the actual messages produced by running that snippet against the source at the commit named in Environment.

## Expected

The object returned by `rlm()` is accepted by `rlm.delete_subagent()`, and one attribute spelling for the child's name works on both types. Spawning a child and then deleting it is the most obvious two-line lifecycle in the API and arguably should not require a type conversion or a detour through an id field.

## Actual

- `rlm.delete_subagent(handle)` raises `TypeError: target must be str or RLMSubagent, got RLMSpawnHandle`. The workaround is `await rlm.delete_subagent(handle.rlm_child_id)`, since the `str` branch accepts the child id - but nothing in the docs or prompts suggests this is necessary.
- `RLMSubagent.name` raises `AttributeError`, and `RLMSpawnHandle.session_name` raises `AttributeError`.

Worth noting how asymmetric this is: the `str` branch is permissive. Host-side, `_rlmSubagentMatchesTarget` (`packages/coding-agent/src/core/agent-session.ts` L9146-9153) resolves a string selector against `rlm_child_id`, `active_session_id`, `session_id`, **or** `session_name`, so even `await rlm.delete_subagent(handle.name)` succeeds. Only the typed-object path is narrow, and it is narrow against the one type the spawn call hands back.

Both failures surface only at runtime inside a REPL cell, mid-turn. The annotations are present and a static type checker would flag the `delete_subagent(handle)` call, but nothing type-checks kernel cells.

## Code pointers

`prime-agent-runtime/src/rlm/__init__.py` - the two dataclasses, neither of which has an alias property (L27-32 and L43-50):

```python
@dataclass(frozen=True)
class RLMSpawnHandle:
    rlm_child_id: str
    name: str
    session_dir: Path
    model: str
```

```python
@dataclass(frozen=True)
class RLMSubagent:
    rlm_child_id: str
    active_session_id: str | None
    session_id: str | None
    session_name: str
    session_dir: Path
    status: str
```

`prime-agent-runtime/src/rlm/__init__.py` L219-230, `delete_subagent` - the narrow union and the rejecting `else` branch (first ten lines shown):

```python
async def delete_subagent(target: str | RLMSubagent) -> RLMSubagent:
    """Delete one running or retained direct child from the current parent session."""
    if isinstance(target, RLMSubagent):
        selector = target.rlm_child_id
    elif isinstance(target, str):
        selector = target.strip()
        if not selector:
            raise ValueError("target must not be empty")
    else:
        raise TypeError(f"target must be str or RLMSubagent, got {type(target).__name__}")
```

`prime-agent-runtime/src/rlm/__init__.py` L297, `_RLMCallable.delete_subagent` forwards the same narrow type, so the `rlm.delete_subagent(...)` spelling used everywhere in the docs inherits the restriction:

```python
    async def delete_subagent(self, target: str | RLMSubagent) -> RLMSubagent:
        return await delete_subagent(target)
```

Every spawn entry point is typed `-> RLMSpawnHandle` - module-level `run` (L143), `_RLMCallable.run` (L288), `_RLMCallable.__call__` (L300), and `_CallableModule.__call__` (L309) - confirming that the object a caller holds at spawn time is always the type `delete_subagent` rejects.

The mismatch is reinforced by shipped guidance, which teaches `.name` at spawn time and `.session_name` at list time and never flags that these are different types:

- `packages/coding-agent/src/core/prompts/rlm.ts` L136 and L149 are pushed into the same generated RLM prompt section (both inside the `allowRecursion && hasIpython` block; L136 additionally requires `hasAgentMessage`), and both bind the identifier `child`, but no single binding satisfies both. L136: ``"Use `await agent_message.list_agents()` to discover family and `await rlm.list_subagents()` to recover direct child handles. Use `agent_message.send(..., receiver_role='child', receiver_name=child.name)` for follow-ups."`` - `.name` requires an `RLMSpawnHandle`, yet the same sentence sources `child` from `list_subagents()`, which yields `RLMSubagent`. L149: ``"Delete a direct child explicitly with `await rlm.delete_subagent(child)` when it is no longer needed."`` - requires an `RLMSubagent`.
- `packages/coding-agent/src/core/refinement/refinement.ts` L138 and L458 both instruct: ``handle = await rlm('sub-task')`` returning ``rlm_child_id``, ``name``, ``session_dir``, ``model``, then ``await agent_message.send(..., receiver_role='child', receiver_name=handle.name)``. That guidance is self-consistent, but it is what establishes `.name` as the spelling an agent carries forward.
- `packages/coding-agent/docs/rlm.md` L59 prints `handle.rlm_child_id, handle.name, handle.session_dir, handle.model`; L84 uses `receiver_name=api_review.name` on a spawn handle; L97 prints `child.session_name, child.status, child.active_session_id`; L103 calls `await rlm.delete_subagent(children[0])` on a registry entry. No example anywhere shows `delete_subagent(handle)`, and none warns that it fails.
- `packages/coding-agent/skills/agent-observe/SKILL.md` L21-24 uses `child.session_name` then `await rlm.delete_subagent(child)`, correct only because `child` came from `list_subagents()`.

## Suggested fix

Three small, backward-compatible changes in `prime-agent-runtime/src/rlm/__init__.py`. Properties are class attributes, not fields, so they compose with `@dataclass(frozen=True)` without changing the generated `__init__` signature.

1. Alias on the spawn handle:

```python
@dataclass(frozen=True)
class RLMSpawnHandle:
    rlm_child_id: str
    name: str
    session_dir: Path
    model: str

    @property
    def session_name(self) -> str:
        """Alias for `name`; matches RLMSubagent.session_name."""
        return self.name
```

2. Alias on the registry entry:

```python
@dataclass(frozen=True)
class RLMSubagent:
    ...

    @property
    def name(self) -> str:
        """Alias for `session_name`; matches RLMSpawnHandle.name."""
        return self.session_name
```

3. Widen `delete_subagent` (and the `_RLMCallable.delete_subagent` forwarder at L297) to accept either object, selecting `rlm_child_id` from both:

```python
async def delete_subagent(target: str | RLMSubagent | RLMSpawnHandle) -> RLMSubagent:
    if isinstance(target, (RLMSubagent, RLMSpawnHandle)):
        selector = target.rlm_child_id
    elif isinstance(target, str):
        selector = target.strip()
        if not selector:
            raise ValueError("target must not be empty")
    else:
        raise TypeError(
            f"target must be str, RLMSubagent, or RLMSpawnHandle, got {type(target).__name__}"
        )
```

Optionally, once the aliases exist, `docs/rlm.md` could add one line stating that `rlm()` returns `RLMSpawnHandle` and `list_subagents()` returns `RLMSubagent`, that `name` and `session_name` are interchangeable, and that both are accepted by `delete_subagent`.

If aliasing both directions is unattractive (two spellings for one concept is its own cost), widening `delete_subagent` alone would fix the lifecycle break, and a one-line docs note could cover the attribute difference. Would you prefer that narrower shape?

## Impact

This lands hardest on long-running multi-agent sessions. A parent that spawns children across many turns naturally keeps the handles it got back and cleans them up later; that cleanup path is exactly the one that raises `TypeError`. The failure appears mid-turn inside a REPL cell, so the agent has to notice the traceback, infer that `rlm_child_id` is the escape hatch, and retry - spending a turn on an API asymmetry rather than on the task.

Because the spelling that fails is one the shipped prompts and docs teach, this is not obviously a mistake a careful reader avoids; the guidance in `prompts/rlm.ts` is internally inconsistent within a single generated prompt section. Agents that skip cleanup after the error would accumulate retained children, since completed daemon-backed children stay addressable while the parent session is open (`docs/rlm.md` L100).

Related and worth cross-linking, though distinct: #762 (`agent_message` / `agent_observe` skill API surface not discoverable) is the same family of "the Python-side surface is not learnable from what ships". #799 is a comparable small-but-load-bearing correctness gap in `rlm.find_models()`.

sashankh:
## Summary

`rlm()` returns an `RLMSpawnHandle` whose child name field is `name`, while `rlm.list_subagents()` returns `RLMSubagent` objects whose child name field is `session_name`. Neither type carried the other's attribute, and `delete_subagent()` accepted only `str | RLMSubagent`, so the natural call `await rlm.delete_subagent(handle)` on the handle just returned by `rlm(...)` raised `TypeError`.

This matters because the shipped system prompt teaches exactly the call that fails:

> Delete a direct child explicitly with `await rlm.delete_subagent(child)` when it is no longer needed.

## Reproduction

Confirmed live from the IPython kernel of a running prime-agent session, before the fix:

```
handle type: RLMSpawnHandle | .name = pi-arch
handle.session_name    -> AttributeError: 'RLMSpawnHandle' object has no attribute 'session_name'
delete_subagent(handle) -> TypeError: target must be str or RLMSubagent, got RLMSpawnHandle
registry type: RLMSubagent | .session_name = pi-arch
RLMSubagent.name       -> AttributeError: 'RLMSubagent' object has no attribute 'name'
```

## Changes

- `delete_subagent()` accepts `RLMSpawnHandle` in addition to `str` and `RLMSubagent`, resolving the selector from `rlm_child_id` exactly as it already does for `RLMSubagent`.
- `RLMSpawnHandle.session_name` and `RLMSubagent.name` added as read-only aliases, so either spelling works on either type and both documented spellings become correct.
- Type hints widened on both the module-level function and the `rlm` proxy method.

Behavior for existing `str` and `RLMSubagent` targets is unchanged. The `TypeError` message now lists all three accepted types; the one existing assertion pinned to the old message was updated.

## Tests

Added `RlmSpawnHandleInteropTest` to `prime-agent-runtime/test/test_subagent_registry.py` covering each symptom in the issue: the `session_name` alias, the `name` alias, both types agreeing on the same child, `delete_subagent` accepting a handle, and handle/subagent resolving to an identical host selector.

```
$ uv run python -m unittest discover -s test -p 'test_subagent_registry.py'
Ran 15 tests in 0.009s
OK
```

Full runtime suite goes from 64 to 69 tests. Two pre-existing `ModuleNotFoundError: No module named 'mcp'` errors in `test_mcp_base.py` are present identically on unmodified `main` and are unrelated to this change.

fixes #824

> [!NOTE]
> ### Fix `delete_subagent` to accept `RLMSpawnHandle` and align `name`/`session_name` attributes
> - `delete_subagent` now accepts an `RLMSpawnHandle` (the value returned by `rlm(...)`) in addition to `str` and `RLMSubagent`, resolving it via `rlm_child_id`.
> - `RLMSpawnHandle` gains a `session_name` property aliasing `name`; `RLMSubagent` gains a `name` property aliasing `session_name`, making the two types interchangeable for name lookups.
> - New test class `RlmSpawnHandleInteropTest` in [test_subagent_registry.py](https://github.com/PrimeIntellect-ai/prime-agent/pull/854/files#diff-75abf03ff4e1d298c5622c9acffeb219d7cef6b86503d10ad261e1993a64a7dd) covers the alias properties and `delete_subagent` behavior for spawn handles.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized f0b90d8.</sup>
> 
>

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [Unreleased]
 
 - Fixed the bundled `websearch` skill description and missing-key guidance omitting the `/login` → **MCP Connections** step required to configure Serper.
+- Fixed `rlm.delete_subagent()` rejecting the handle returned by `rlm(...)`, and added matching `name`/`session_name` aliases so either spelling works on spawn handles and subagent records ([#854](https://github.com/PrimeIntellect-ai/prime-agent/pull/854) by [@sashankh](https://github.com/sashankh)).
 
 ## [0.7.0] - 2026-08-05
 
--- a/prime-agent-runtime/src/rlm/__init__.py
+++ b/prime-agent-runtime/src/rlm/__init__.py
@@ -31,6 +31,11 @@ class RLMSpawnHandle:
     session_dir: Path
     model: str
 
+    @property
+    def session_name(self) -> str:
+        """Alias for :attr:`name`, matching :attr:`RLMSubagent.session_name`."""
+        return self.name
+
 
 @dataclass(frozen=True)
 class RLMModel:
@@ -48,6 +53,11 @@ class RLMSubagent:
     session_name: str
     session_dir: Path
     status: str
+
+    @property
+    def name(self) -> str:
+        """Alias for :attr:`session_name`, matching :attr:`RLMSpawnHandle.name`."""
+        return self.session_name
 
 
 def _install_control_comm_handlers() -> None:
@@ -216,16 +226,23 @@ async def list_subagents() -> list[RLMSubagent]:
     return [_subagent_from_payload(entry) for entry in entries]
 
 
-async def delete_subagent(target: str | RLMSubagent) -> RLMSubagent:
-    """Delete one running or retained direct child from the current parent session."""
-    if isinstance(target, RLMSubagent):
+async def delete_subagent(target: str | RLMSubagent | RLMSpawnHandle) -> RLMSubagent:
+    """Delete one running or retained direct child from the current parent session.
+
+    Accepts the child name, an :class:`RLMSubagent` from :func:`list_subagents`, or the
+    :class:`RLMSpawnHandle` returned by ``rlm(...)``/:func:`run`.
+    """
+    if isinstance(target, (RLMSubagent, RLMSpawnHandle)):
         selector = target.rlm_child_id
     elif isinstance(target, str):
         selector = target.strip()
         if not selector:
             raise ValueError("target must not be empty")
     else:
-        raise TypeError(f"target must be str or RLMSubagent, got {type(target).__name__}")
+        raise TypeError(
+            "target must be str, RLMSubagent, or RLMSpawnHandle, "
+            f"got {type(target).__name__}"
+        )
     payload = await host_request("rlm.delete_subagent", {"target": selector})
     return _subagent_from_payload(payload.get("subagent"), "rlm.delete_subagent")
 
@@ -294,7 +311,9 @@ class _RLMCallable:
     async def list_subagents(self) -> list[RLMSubagent]:
         return await list_subagents()
 
-    async def delete_subagent(self, target: str | RLMSubagent) -> RLMSubagent:
+    async def delete_subagent(
+        self, target: str | RLMSubagent | RLMSpawnHandle
+    ) -> RLMSubagent:
         return await delete_subagent(target)
 
     async def __call__(self, prompt: str, **kwargs: Any) -> RLMSpawnHandle:
--- a/prime-agent-runtime/test/test_subagent_registry.py
+++ b/prime-agent-runtime/test/test_subagent_registry.py
@@ -191,7 +191,9 @@ class RlmSubagentRegistryTest(unittest.TestCase):
 
         with self.assertRaisesRegex(ValueError, "target must not be empty"):
             asyncio.run(rlm_module.delete_subagent("   "))
-        with self.assertRaisesRegex(TypeError, "target must be str or RLMSubagent"):
+        with self.assertRaisesRegex(
+            TypeError, "target must be str, RLMSubagent, or RLMSpawnHandle"
+        ):
             asyncio.run(rlm_module.delete_subagent(123))
 
     def test_rejects_invalid_registry_payload(self) -> None:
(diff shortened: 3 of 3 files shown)


kevinjosethomas:

Issue #810, opened 2026-08-07
Subagent model request "gpt 5.6 sol" not recognized; agents spawned with mangled selector "5.6 gpt sol"


guala19:
## Summary

Using the prime-agent TUI with the **opencode-go** subscription (parent model `opencode-go/deepseek-v4-flash`) and an **OpenAI Pro x5** subscription linked (which serves the `openai-codex` catalog), I asked the orchestrator to spawn a subagent to work with **"gpt 5.6 sol"**. The model was **not recognized in the model list** and, instead of failing loudly, subagents were spawned with the **mangled selector `5.6 gpt sol`** (reordered tokens, non-existent model).

## Environment

- prime-agent **0.7.0** (stable), interactive TUI session, macOS
- Parent model: `opencode-go/deepseek-v4-flash` (opencode-go subscription)
- Also linked via `/login`: OpenAI Pro x5 → `openai-codex` provider (`prime-agent model list` shows the full catalog, including `openai-codex/gpt-5.6-sol`)

## Repro (verified from IPython in the same install/credentials)

1. `prime-agent model list` → shows `openai-codex/gpt-5.6-sol`, `openai-codex/gpt-5.6-luna`, etc., and `opencode-go/deepseek-v4-flash`.
2. RLM-side model discovery never surfaces the `openai-codex` provider, even though the subscription is linked and one-shots on those models work:

```python
await rlm.find_models("", limit=20)
# -> only opencode-go and prime-inference selectors; openai-codex never appears
await rlm.find_models("gpt 5.6 sol", limit=20)
# -> ['prime-inference/openai/gpt-5.6-sol', 'prime-inference/openai/gpt-5.6-sol-pro']
#    (a *different provider's* same-named model — NOT the openai-codex one my subscription serves)
await rlm.find_models("5.6 gpt sol", limit=20)
# -> []   (word order matters: matching is done on the flattened, non-alphanumeric-stripped string)
```

3. In the TUI, prompting the parent (`opencode-go/deepseek-v4-flash`) to launch a subagent with "gpt 5.6 sol": it reports the model is not in the list, and the spawned subagent entries carry the garbled selector **`5.6 gpt sol`** instead of any real model id.

## Expected

- `rlm.find_models("gpt 5.6 sol")` should surface `openai-codex/gpt-5.6-sol`, which the linked OpenAI Pro x5 credential demonstrably serves (`prime-agent model list` and direct runs both confirm it).
[...]

PR #855, opened 2026-08-07
fix(coding-agent): make rlm.find_models represent every provider and ignore word order

### LINKED ISSUE #799: rlm.find_models() returns an alphabetical head slice that reads as the full reachable set
## Summary

With no query, `rlm.find_models` scores every model equally and falls through to an alphabetical tiebreak, then truncates to `limit` (default 8). The result is a single-vendor list that looks like a complete answer to "what can I reach", with nothing in the return value marking it as a page.

## Observed

prime-agent 0.7.0, six providers configured, **155 reachable models across 8 providers**:

```python
await rlm.find_models()             # -> 8 results, all anthropic
await rlm.find_models("", limit=20) # -> 20 results: 15 anthropic, 2 minimax, 3 local
```

`openai-codex`, `openrouter`, `qwencloud`, `zai` never appear — they sort after `anthropic` and `Archie-*`. The default call omits even `anthropic/claude-sonnet-5`, which sorts behind `claude-opus-*`.

## Cause

`findRlmModelMatches` in `packages/coding-agent/src/core/rlm-runtime.ts` assigns `score = 0` to every candidate when the query is empty, so ordering is decided entirely by `a.selector.localeCompare(b.selector)` before `.slice(0, limit)`.

## Why it matters

An orienting agent calls this bare — it is the natural first move before delegating. In a real run, an orchestrator did exactly that, concluded only the anthropic family was natively reachable, and routed three delegations through shell-CLI fallbacks that were unnecessary. It then reported the detour as correct harness behaviour, because nothing in the response contradicted the inference. A second, query-by-name pass revealed all four vendors spawn natively via `rlm(model=...)`.

The failure mode is confident wrongness rather than an error: no exception, no empty result, just a plausible subset.

## Suggestions

1. Make the unqualified call a summary — providers with model counts — instead of an arbitrary first page.
2. Return a truncation marker (`total`, or `truncated: true`) so a caller can distinguish a page from a catalogue.
3. Rank the default slice for provider diversity rather than alphabetically.
4. At minimum, state in the docstring and in `rlm-runtime.md` that an unqualified call is a sample, and that reachability is verified by querying a specific model id.

The docstring currently reads *"Search a bounded list of models backed by active user credentials"*, which reads as if credentials were the bound. In practice the bound is `limit`, applied after an alphabetical sort.

## Related

Distinct from #702 / #639 / #696 / #740 (openai-codex discovery) and from #724 / #539 (interactive model-selector ranking) — this is `findRlmModelMatches` on the RLM path.

sashankh:
## Summary

With no query, `findRlmModelMatches` scored every candidate `0`, so ordering fell through to an alphabetical tiebreak on the selector before `.slice(0, limit)`. With 155 models across 8 providers, a bare `rlm.find_models()` returned 8 results from a single provider, and nothing in the return value marked it as a page.

The failure mode is confident wrongness rather than an error. An orienting agent calls this bare as its first move before delegating, and can reasonably conclude the providers it cannot see are unreachable. Issue #799 records exactly that: an orchestrator routed three delegations through shell-CLI fallbacks that were unnecessary.

While in the same matcher, this also fixes the token-order half of #810: the query was normalized into a single string, so `"5.6 gpt sol"` returned `[]` while `"gpt 5.6 sol"` matched. The `openai-codex` reachability half of #810 is out of scope here.

## Changes

- Round-robin ranked candidates by provider when the query is empty, so a bounded page reaches every provider instead of the alphabetical head of one.
- Return `total`, `truncated`, and per-provider match counts for the whole match set, so a page cannot be mistaken for the full catalog.
- Score a multi-word query token-wise as a final tier, keeping exact, prefix, and partial ranking ahead of it, so word order no longer decides whether a model matches.
- Surface the page summary to Python as `RLMModelPage`, a `list` subclass, so existing code that treats the result as a list keeps working while `total` / `truncated` / `providers` are available.

Non-empty-query ranking is otherwise unchanged.

## Tests

New regression `packages/coding-agent/test/suite/regressions/799-find-models-provider-coverage.test.ts` covers an unqualified page sampling every provider and reporting what it was cut from, a reordered query matching the same models as the original, and query ranking still taking precedence over provider coverage.

```
$ npx tsx ../../node_modules/vitest/dist/cli.js --run \
    test/suite/regressions/799-find-models-provider-coverage.test.ts \
    test/suite/regressions/4649-subagent-model-selection.test.ts
Test Files  2 passed (2)
```

```
$ npm run check
Checked 899 files. No fixes applied.
Installer render check passed.
```

```
$ cd prime-agent-runtime && uv run python -m unittest discover -s test -p 'test_subagent_registry.py'
Ran 11 tests - OK
```

Two pre-existing `ModuleNotFoundError: No module named 'mcp'` errors in the Python runtime suite, and 4 pre-existing failures in `test/resource-loader.test.ts`, reproduce identically on unmodified `origin/main` and are unrelated to this change.

closes #799
refs #810

> [!NOTE]
> ### Fix `rlm.find_models` to cover every provider and match query tokens in any order
> - `findRlmModelMatches` adds an order-independent token-matching tier after exact/prefix/partial ranking, so queries like `gpt 4o` and `4o gpt` return the same results.
> - Unqualified (empty) queries now interleave results across providers via round-robin rather than sorting alphabetically, ensuring every provider appears in the page.
> - `find_models` returns `RLMModelPage` (a `list` subclass) with `total`, `truncated`, and `providers` metadata instead of a plain list, in both Python (`prime-agent-runtime`) and TypeScript (`rlm-runtime.ts`).
> - The runtime readiness check in [`bootstrap.ts`](https://github.com/PrimeIntellect-ai/prime-agent/pull/855/files#diff-e87b9a04ab9b4ac02e8dd33af2f2f58320d8b7f6be6853acf4a3e05a34623cae) now requires `RLMModelPage` to be importable from `rlm`.
> - Behavioral Change: callers of `find_models` / `findRlmModels` that expected a plain list or `{ models }` object now receive a richer result type with page metadata fields.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized cca59fd.</sup>
> 
>


kevinjosethomas:

PR #858, opened 2026-08-07
fix(coding-agent): make /login sign-in URL copyable from headless terminals

parithosh:
## Problem

When logging in to Claude (Anthropic) or OpenAI from a headless machine over SSH, the OAuth sign-in URL shown by `/login` cannot be copied cleanly into a browser on another machine.

Root cause: the login dialog renders the URL through the TUI's `Text` component, which hard-wraps it at panel width. Each wrapped segment becomes a separate physical terminal row (the renderer requires this — it throws on over-width lines), so mouse selection copies the URL **with embedded newlines**, and pasting it into a browser fails. On headless SSH the OSC 8 hyperlink was also disabled, because capability detection is conservative for unknown terminals, so there was no clickable escape hatch either.

## Changes

**`packages/tui`** — promote the private OSC 52 selection-copy fallback to a public `TUI.copyToClipboard(text)`. It routes through the host `onCopy` hook when set, else writes `OSC 52`, which reaches the clipboard of the machine the terminal is running on — i.e. the local machine, even over SSH.

**`packages/coding-agent` login dialog:**
- New `c` shortcut while the sign-in link is on screen: copies the full URL via `tui.copyToClipboard`. The hint line becomes *"Copy sent via OSC 52 — if pasting shows nothing, your terminal blocked it (tmux: set-clipboard on)."* — deliberately honest, since OSC 52 has no ack and is refused by some setups (xterm's default `disallowWindowOps`, iTerm2 without "Applications may access clipboard", tmux without `set-clipboard on`).
- The shortcut only fires while no paste field is visible (verification codes / API keys may contain `c`) and is disarmed when the screen is replaced, so it can never copy a stale URL.
- Emit the OSC 8 hyperlink **unconditionally** instead of gating on detected capabilities. The gate exists for labeled links whose URL would vanish on OSC-8-ignorant terminals; here the visible label *is* the URL, so degradation is plain text. From a headless SSH session (where `TERM` can't identify the outer terminal) a modern local terminal can now ctrl+click the link — the browser opens locally and the existing "paste the redirect URL" fallback completes the flow. `pi-tui` already re-opens OSC 8 across wrapped lines, so the whole wrapped region is one link.

## Prior art

Surveyed how other agent CLIs handle this (source-verified):
- **Codex CLI** renders the auth URL inside its ratatui TUI (also hard-wrapped) and overlays **OSC 8 unconditionally** across the wrapped cells (`tui/src/onboarding/auth.rs`, `mark_url_hyperlink`), plus a device-code flow for headless. This PR's OSC 8 ungate mirrors that.
- **Claude Code**'s TUI login has a copy affordance with "(Copied!)" feedback (`urlToCopy` string pool) — the `c` shortcut mirrors that.
- **Gemini CLI**, **Amp**, and **opencode** avoid the problem by printing the URL as a raw stdout write *outside* the TUI so the terminal soft-wraps it (Gemini reorders startup specifically "to make copying the link possible").

## Not in this PR (follow-up candidates)

- Terminal soft-wrap of the URL is impossible inside the current renderer (one array line = one terminal row, enforced by a throw), so a Gemini-style pre-TUI plain print or a Codex-style device-code flow would be the next escalation if OSC 52 + OSC 8 + paste-redirect prove insufficient in practice.

## Testing

- `packages/tui`: 753 tests pass; typecheck clean.
- `packages/coding-agent`: `login-dialog`, `auth-flows`, `oauth-selector` suites pass (42 tests), including new cases for: copy-on-`c` + feedback, hint rendering, `c` routing to the paste field while input is visible, stale-URL disarm, OSC 8 emitted regardless of detected capabilities, and full URL text surviving a hard wrap at width 44.
- Repo pre-commit checks (biome, tsgo, installer render, browser smoke) pass.

> [!NOTE]
> ### Add 'c' key to copy the /login sign-in URL from headless terminals
> - The sign-in URL in the login dialog now always renders as an OSC 8 hyperlink (previously gated on terminal capability) and shows a `c copy link` hint while visible.
> - Pressing `c` calls the new `TUI.copyToClipboard` method, which uses the host `onCopy` callback if provided, or falls back to OSC 52 with a base64-encoded payload (dropped if the encoded payload exceeds 100,000 characters).
> - The copy hint and `c` keybinding are suppressed while the manual-paste or prompt input fields are visible, routing `c` into the input instead.
> - Behavioral Change: OSC 8 hyperlink sequences are now emitted unconditionally for the sign-in URL, regardless of reported terminal capabilities.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized aa61398.</sup>
> 
>


kevinjosethomas:

Issue #751, opened 2026-08-06
v0.7.0 & v0.7.1 release tarball is missing dist/bundle/amazon-bedrock.js, breaking the amazon-bedrock provider


HenryL-system1:
## Description

The officially published v0.7.0 release artifact fails at runtime for the `amazon-bedrock` provider:

```
Error: Cannot find module '/opt/homebrew/lib/node_modules/prime-agent/dist/bundle/amazon-bedrock.js' imported from
/opt/homebrew/lib/node_modules/prime-agent/dist/bundle/chunk-ALQBG3TN.js
```

## Root cause

`dist/bundle/chunk-ALQBG3TN.js` dynamically imports the bedrock provider chunk:

```js
bedrockProviderModulePromise ||= importNodeOnlyProvider("./amazon-bedrock.js").then(...)
```

but `dist/bundle/amazon-bedrock.js` is **not present** in the shipped package. Every other provider entry point exists as its own chunk (`anthropic-JMWPKDC4.js`, `google-vertex-SQF6LM63.js`, `mistral-OBA3N6GZ.js`, `openai-completions-5PBGGQ35.js`, etc.) — only the bedrock one is missing.

This affects both distribution channels, which serve byte-identical artifacts:
- The R2 CDN URL used by `prime-agent update` (`https://pub-728493de92a943e2a9b2d17b4719f318.r2.dev/releases/v0.7.0/prime-agent-0.7.0.tgz`)
- The GitHub release asset `prime-agent-0.7.0.tgz` on the [v0.7.0 release](https://github.com/PrimeIntellect-ai/prime-agent/releases/tag/v0.7.0)

Both have SHA256 `88b6578518c72cd51a825bc80f28e0fef9a64c67de4a7d6fd7afd7ca1b34da0b`, confirming it's the same build, not a corrupted local install.

## Likely tie-in to #651

I believe this was caused by the same underlying issue reported in #651: `@aws-sdk/core@^3.977.6` did not exist on the npm registry at the time v0.7.0 was built in CI. That issue was closed because the npm package now exists and a **fresh `npm install`** succeeds — but the **already-published v0.7.0 tarball was never rebuilt/republished**, so it still ships without the bedrock bundle chunk. It looks like the build step for that entry point failed silently (e.g. esbuild couldn't resolve the then-missing `@aws-sdk/core`) without failing the overall build/release.

[...]

Issue #843, opened 2026-08-07
packages/coding-agent: track official Homebrew core submission readiness


fettpl:
## Summary

Prime Agent already has an official [`homebrew/core` submission](https://github.com/Homebrew/homebrew-core/pull/297372) for `v0.7.0`. The formula passed the current Homebrew macOS and Linux CI matrix, but review is blocked on the vendored ZeroMQ dependency. This issue tracks the remaining upstream and packaging work through merge and final installation verification.

`homebrew/core` is the correct route because Prime Agent is open-source, command-line-only software. Homebrew normally requires that category to be built from source as a formula; a rejected core formula would not automatically become eligible for `homebrew/cask`.

## Current state

- [`Homebrew/homebrew-core#297372`](https://github.com/Homebrew/homebrew-core/pull/297372) is open and remains the canonical submission.
- Its current CI passed on Linux arm64, Linux x86_64, macOS arm64, and macOS x86_64.
- Homebrew review requested use of the existing `zeromq` formula instead of the native library vendored by the `zeromq` npm package.
- [`zeromq/zeromq.js#736`](https://github.com/zeromq/zeromq.js/issues/736) now tracks that request. A ZeroMQ maintainer says the `zmq_shared` npm option exists but has not been tested recently.
- The formula installs the published `prime-agent-0.7.0.tgz` asset with SHA-256 `88b6578518c72cd51a825bc80f28e0fef9a64c67de4a7d6fd7afd7ca1b34da0b`.
- [`#751`](https://github.com/PrimeIntellect-ai/prime-agent/issues/751) reports that this exact `v0.7.0` tarball is missing `dist/bundle/amazon-bedrock.js`, so `--version` passing is not enough to establish that the release artifact is suitable for Homebrew.

### 1. Build the Node binding against Homebrew ZeroMQ

- Test the exact `zeromq@6.5.0` package used by Prime Agent with its `zmq_shared` configuration.
- Declare the Homebrew `zeromq` formula and only the build dependencies demonstrated by a clean build.
- Verify the installed native module with `otool -L` on macOS and `ldd` on Linux.
- Confirm no private `libzmq` copy remains in the keg.
- If the existing option is broken, prepare a focused `zeromq.js` fix linked to `zeromq.js#736`.

### 2. Use a complete stable release artifact

- Resolve `#751` and publish a corrected stable release rather than asking Homebrew to merge a formula around a known-broken tarball.
[...]

Issue #651, opened 2026-08-05
v0.7.0 cannot be installed: @aws-sdk/core@^3.977.6 does not exist (broken transitive dependency)


AppleLamps:
## Summary

The v0.7.0 release cannot be installed because of a broken transitive dependency: `@aws-sdk/client-bedrock-runtime@^3.1083.0` resolves to `3.1104.0`, which requires `@aws-sdk/core@^3.977.6` — a version that does not exist on the npm registry (the latest published is `3.977.3`).

This affects all platforms, not just Windows, since the release is distributed as an npm tarball (`prime-agent-0.7.0.tgz`) installed via `npm install -g`.

## Reproduction

```bash
npm install -g prime-agent-0.7.0.tgz
```

Fails with:

```
npm error code ETARGET
npm error notarget No matching version found for @aws-sdk/core@^3.977.6.
```

## Root cause

`packages/ai` depends on `@aws-sdk/client-bedrock-runtime@^3.1083.0`. The caret range resolves to the latest `3.1104.0`, whose `package.json` declares `"@aws-sdk/core": "^3.977.6"`. However, the highest published `@aws-sdk/core` is `3.977.3`, so npm cannot satisfy the dependency graph.

Verified:

- `@aws-sdk/client-bedrock-runtime@3.1083.0` requires `@aws-sdk/core@^3.975.1` (satisfiable)
- `@aws-sdk/client-bedrock-runtime@3.1104.0` requires `@aws-sdk/core@^3.977.6` (NOT satisfiable)

## Workaround (local)

Install into a project with npm `overrides`:

```json
{
  "dependencies": {
    "prime-agent": "file:./prime-agent-0.7.0.tgz"
  },
  "overrides": {
    "@aws-sdk/client-bedrock-runtime": "3.1083.0",
    "@aws-sdk/core": "3.977.3"
  }
}
```

## Suggested fix

Pin `@aws-sdk/client-bedrock-runtime` to a version whose `@aws-sdk/core` requirement is satisfiable (e.g. `3.1083.0`), or wait for `@aws-sdk/core@3.977.6+` to be published and re-release. The `^3.1083.0` caret range is too loose and currently resolves to a broken build.

AppleLamps:
Closing: the underlying issue is resolved. @aws-sdk/core@3.977.6 has since been published to the npm registry, so the v0.7.0 release now installs cleanly without any workaround (verified with a fresh install). Thanks!

PR #860, opened 2026-08-07
fix(coding-agent): include Bedrock in Node bundle

imMamdouhaboammar:
## Summary

- add the compiled Amazon Bedrock provider as an explicit esbuild entry alongside the CLI
- keep the existing lazy node-only provider loading path intact
- fail the bundle step when an `importNodeOnlyProvider("./x.js")` reference points to an asset that was not emitted

## Root cause

`packages/ai/src/providers/register-builtins.ts` loads Bedrock through `importNodeOnlyProvider("./amazon-bedrock.js")`, where the helper calls `import(specifier)`. Because that import specifier is dynamic from esbuild's perspective, the coding-agent bundle does not emit `dist/bundle/amazon-bedrock.js` even though the generated CLI chunk still references it.

This is reproducible on current `main`: a clean build produced no Bedrock bundle asset while the CLI chunk contained the unresolved `./amazon-bedrock.js` reference.

## Validation

Before the fix, the new integrity check failed with:

```text
Bundle contains missing node-only provider assets:
- chunk-27BW5CFV.js -> ./amazon-bedrock.js
```

After adding the explicit entry:

- root `npm run build` passed
- `dist/bundle/amazon-bedrock.js` was emitted and imports successfully
- the emitted module exports `streamBedrock` and `streamSimpleBedrock`
- root `npm run check` passed
- `npm pack --workspace packages/coding-agent --dry-run --json` includes `dist/bundle/amazon-bedrock.js`
- `git diff --check` passed

This addresses the source-side cause of #751. The already-published v0.7.0 artifact still requires a replacement release or a new release after this lands.

> [!NOTE]
> ### Include Amazon Bedrock provider in coding-agent Node bundle
> - Adds `amazon-bedrock` as a second esbuild entry point in [bundle.mjs](https://github.com/PrimeIntellect-ai/prime-agent/pull/860/files#diff-8606e8936a1a9a467bc717dab8da4d6bafd17f129a8e044e3054e4567b18803d), alongside the existing `cli` entry, so the Bedrock provider is emitted into the bundle output.
> - Adds a `verifyNodeOnlyProviderAssets` post-build check that scans output `.js` files for `importNodeOnlyProvider` calls and throws if any referenced asset files are missing.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 36842d0.</sup>
> 
>

### DIFF
--- a/packages/coding-agent/scripts/bundle.mjs
+++ b/packages/coding-agent/scripts/bundle.mjs
@@ -11,9 +11,9 @@
  * compiled Bun binary), keyed off the __PI_BUNDLED__ define below, so extension
  * imports of pi packages share the bundle's module instances.
  */
-import { chmodSync, readFileSync, rmSync } from "node:fs";
+import { chmodSync, existsSync, readFileSync, readdirSync, rmSync } from "node:fs";
 import { execFileSync } from "node:child_process";
-import { dirname, join } from "node:path";
+import { dirname, join, resolve } from "node:path";
 import { fileURLToPath } from "node:url";
 import { build } from "esbuild";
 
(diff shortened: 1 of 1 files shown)


seankumar-tl:
Confirming this reproduces on the released v0.7.1, not just `main`, so the fix is still needed. A fresh 0.7.1 install has all eight other provider entry points in `dist/bundle/` and no `amazon-bedrock.js`, and any Bedrock model dies with `Cannot find module .../dist/bundle/amazon-bedrock.js`.

Relevant to #751, where the missing chunk is still attributed to the `@aws-sdk/core` registry gap from #651: 0.7.1 was built after that cleared and the chunk is still missing, which points at your root cause rather than that one.

I landed on the same diagnosis from the shipped artifact before I found this PR, and confirmed the emitted module only has to satisfy `streamBedrock` / `streamSimpleBedrock`. A local re-export at `dist/bundle/amazon-bedrock.js` pointing at the `packages/ai` copy is enough to make Bedrock stream correctly, which matches your validation notes.

One small thing on `verifyNodeOnlyProviderAssets`: the regex requires a literal argument, so it covers today's call sites but would pass silently if anyone ever passed a variable into `importNodeOnlyProvider`, which is the exact shape that caused this bug. Probably fine as is, just noting the check inherits the blind spot it's guarding against.

fettpl:
@badlogic this is now the remaining source fix for #751 and the official Homebrew submission tracked in #843. The released `v0.7.1` archive still omits `dist/bundle/amazon-bedrock.js`, matching this PR's root cause.

CI is currently `action_required` because the branch comes from a fork. Could you approve [run 31189426821](https://github.com/PrimeIntellect-ai/prime-agent/actions/runs/31189426821) and review this PR? A corrected stable release after this and #846 will unblock the Prime Agent side of Homebrew/homebrew-core#297372.

kevinjosethomas:

PR #861, opened 2026-08-07
fix(ipython): terminate %%bash subprocesses on interrupt

### LINKED ISSUE #849: Esc/interrupt leaves `%%bash` subprocesses running forever: forceAbort resolves the tool call after 1s but never reaches the cell's grandchildren
## Summary

Interrupting an `ipython` cell that is running a `%%bash` command returns control to the model but does not stop the work. The subprocess keeps running — indefinitely, at full CPU — and because `forceAbort` deliberately leaves `activeExecution` set, the next `ipython` call hits the busy-kernel prompt and offers to *wait* on something that will never be interrupted.

**Tested at:** v0.7.0 (`c98941a2`) · macOS 26.0 (Darwin 25.4.0) · uid 501

## Observed

A cell running `grep -ril <pattern> ~/Dev --include=...` was interrupted with Esc. Seven minutes later:

```
  PID  PPID  PGID %CPU ELAPSED STAT ARGS
90511 90510 90067 99.0   07:05    R  grep -ril ... /Users/ee/Dev --include=*.md ...
90510 90192 90067                    bash
90192 90067 90067                    .../kernel-venv/bin/python -m ipykernel_launcher -f .../connection.json
90067 13639 90067                    prime-agent
```

The only recovery is killing the leaf by hand or choosing "Kill kernel and restart" (which loses the namespace).

## Mechanism

`core/kernel/index.ts:905` — `onAbort` sends `interrupt_request` on control, then arms a 1s timer:

```ts
abortTimer = globalThis.setTimeout(forceAbort, KERNEL_ABORT_GRACE_MS);  // = 1000
```

`forceAbort` calls `resolveExecution(execution, { clearActive: false })`. The `clearActive: false` is deliberate — prime knows the cell may still be running — so the tool call is **abandoned**, not stopped.

The `interrupt_request` raises `KeyboardInterrupt` in the kernel's main thread, but the whole tree shares one PGID (no separate process group is created for the cell), so nothing can group-signal the cell without also signalling prime itself, and a signal to `bash` alone is not forwarded by a non-interactive shell to its foreground child. The leaf never sees it.

## Impact

- Orphans survive the session and keep burning a core; they accumulate across interrupts.
- The subsequent busy-kernel choice (`core/tools/ipython.ts:150-156`) presents "Wait and preserve state" as a real option, but waiting cannot terminate.
- Combined with #848 (cells have no timeout at any layer — `ipythonSchema` has only `code`, and `executeInner` awaits with no deadline), there is no bound anywhere: a cell can neither time out nor be interrupted.

## Suggested direction

Run each cell's shell subprocesses in their own process group (`start_new_session` / `setsid` at the script-magic layer), so an interrupt can signal the group, and escalate SIGINT → SIGTERM → SIGKILL across the existing grace window instead of only resolving the promise locally.

Akashmdshetty:
## Summary

This PR fixes an issue where interrupting a `%%bash`/`%%script` cell returns control to the agent but leaves child subprocesses running indefinitely.

## Changes

- Ensure subprocesses spawned by `%%bash`/`%%script` execute in their own process group.
- Clean up the entire process group when a `KeyboardInterrupt` occurs.
- Preserve existing IPython behavior by wrapping the existing script magic implementation.
- Add regression tests covering the interrupt behavior.

## Motivation

Previously, interrupting long-running `%%bash` cells could leave orphaned subprocesses running after the tool call was abandoned, requiring manual cleanup.

Closes #849

> [!NOTE]
> ### Fix `%%bash` cell interrupts to terminate subprocess groups in IPython
> - Patches `ScriptMagics.shebang` in the IPython bootstrap to wrap `subprocess.Popen`, creating a new process group for child processes spawned during shebang cell execution.
> - On `KeyboardInterrupt`, sends SIGINT then SIGKILL to the entire child process group (POSIX) or uses `taskkill` (Windows), ensuring `%%bash` and similar cells stop reliably.
> - Also updates the abort timeout in `KernelManager` to re-attempt `interrupt()` before calling `forceAbort()` if the same execution is still active.
> - The bootstrap patch is wrapped in `try/except` to be best-effort and avoid breaking the kernel on failure.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized cbf0eb5.</sup>
> 
>


sethkarten:
Thank you for the report and proposed work. This root cause is now covered by maintainer-owned stacked PR #1162, authored independently from `upstream/main`.

We did not inspect or reuse this PR's diff, branch, commits, implementation code, or tests; its public description/comments were used only as a bug report. To keep one review surface, this PR is superseded by #1162 and is being closed.

PR #871, opened 2026-08-07
fix(ai): map Codex stream errors from nested error fields

### LINKED ISSUE #863: Codex stream errors dump raw JSON and drop plan/reset info: mapCodexEvents reads event.code/event.message instead of event.error.*
On `prime-agent` 0.7.0, a Codex subscription quota error surfaces as a raw JSON blob, and the plan/reset information the server supplied is discarded — even though this file already contains code that formats it correctly.

## What the user sees

```
Error: Codex error: {"type":"error","error":{"type":"usage_limit_reached","message":"The usage limit has been
reached","plan_type":"prolite","resets_at":1786160051,"eligible_promo":null,"resets_in_seconds":44419},
"status_code":429,"headers":{"X-Codex-Active-Limit":"premium","X-Codex-Plan-Type":"prolite",
"X-Codex-Primary-Used-Percent":"100","X-Codex-Primary-Window-Minutes":"10080", ...}}
```

Repeated four times, then `Retry failed after 3 attempts: <same blob>`.

Everything needed for a good message is in that payload: the plan (`prolite`), the exact reset instant (`resets_at` → 2026-08-08T03:34:11Z, ~12h away), and the window (`X-Codex-Primary-Window-Minutes: 10080`, i.e. a 7-day cap rather than a burst limit).

## Root cause

`packages/coding-agent` bundles `openai-codex-responses`, whose `mapCodexEvents` reads the wrong nesting level:

```js
if (type === "error") {
  const code = event.code || "";          // undefined — the class is at event.error.type
  const message = event.message || "";    // undefined — the text is at event.error.message
  throw new CodexApiError(`Codex error: ${message || code || JSON.stringify(event)}`, {
    code: code || undefined,
    payload: event
  });
}
```

For this payload both reads miss, so the message falls through to `JSON.stringify(event)` and `code` is left `undefined`.

The correct handling already exists a few functions away, in `parseErrorResponse` (used by the non-streaming path):

```js
if (/usage_limit_reached|usage_not_included|rate_limit_exceeded/i.test(code) || response.status === 429) {
  const plan = err.plan_type ? ` (${err.plan_type.toLowerCase()} plan)` : "";
  const mins = err.resets_at ? Math.max(0, Math.round((err.resets_at * 1e3 - Date.now()) / 6e4)) : undefined;
  const when = mins !== undefined ? ` Try again in ~${mins} min.` : "";
  friendlyMessage = `You have hit your ChatGPT usage limit${plan}.${when}`.trim();
}
```

That would have produced `You have hit your ChatGPT usage limit (prolite plan). Try again in ~740 min.` The streaming path never reaches it.

Minor, in the same function: `message = err.message || friendlyMessage || message` prefers the vague `"The usage limit has been reached"` over the `friendlyMessage` that carries the plan and reset time. Callers that use `info.message` therefore lose the reset time even on the path that computes it.

## Why this is not covered by the open quota work

- #795 (fail closed on provider quota exhaustion) fixes the *classification and retry* half, and I confirmed its `isQuotaExhaustionMessage` regex does match this text. But it matches only because the whole payload was stringified into the error message and happens to contain `usage_limit_reached ... has been reached`. Classification is riding on a JSON dump rather than on `event.error.type`, and #795 does not touch the Codex provider file.
- #688 / #816 concern backoff shape (jitter, cap, `Retry-After`), not payload parsing.

So even with #795 merged, a Codex quota stop still shows a JSON blob and never tells the user when service returns.

## Suggested fix

1. In `mapCodexEvents`, read `event.error?.type ?? event.code` and `event.error?.message ?? event.message`, and carry `event.status_code`, so `CodexApiError.code` is `usage_limit_reached` rather than `undefined`.
2. Reuse the existing `parseErrorResponse` formatting on the stream path so the thrown message names the plan and reset time instead of dumping JSON.
3. In `parseErrorResponse`, prefer `friendlyMessage` over the generic `err.message` when one was built.

Classification would then key off the structured `code` instead of incidental substring matches in a serialized payload.

## Environment

- `prime-agent` 0.7.0 (npm), macOS, node 26.1.0
- Provider `openai-codex` (Codex OAuth), plan `prolite`, streaming path
- Observed 2026-08-07 ~15:00 UTC

## Additional signal in the payload (possible follow-up)

The same response reports a second, non-exhausted bucket on the same account:

```
X-Codex-Primary-Used-Percent: 100          (active limit: premium)
X-Codex-Bengalfox-Primary-Used-Percent: 25 (X-Codex-Bengalfox-Limit-Name: GPT-5.3-Codex-Spark)
X-Codex-Credits-Balance: 0
```

Surfacing remaining per-bucket quota, or naming the still-usable model in the error, would let a run degrade instead of stopping. Happy to split that into its own issue if preferred.

teddytennant:
## Summary
Streamed Codex errors put `type`/`message` under `error`, but `mapCodexEvents` read them at the top level. That turned usage-limit failures into a raw JSON dump and dropped plan/reset info.

The non-stream path already formats these. This shares that formatting for stream errors and prefers the friendly quota text when we build one.

Fixes #863

## Test plan
- [x] Unit test SSE usage_limit_reached payload
- [x] Unit test HTTP 429 path still uses friendly quota text
- [x] packages/ai openai-codex-stream tests pass (19/19)

> [!NOTE]
> ### Fix Codex stream error mapping to extract nested error fields and produce quota-aware messages
> - Updates `mapCodexEvents` in [openai-codex-responses.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/871/files#diff-69623a8c4e616b30339182296d1b69ed62a9d7fcce14b744ed6e638ecf357abc) to read nested `event.error` and top-level `status_code` fields from SSE error events, replacing the previous top-level-only `code`/`message` extraction.
> - Adds a `formatCodexError` helper that normalizes error details into a structured response, generating user-friendly quota and retry messages for `usage_limit_reached`, `usage_not_included`, and `rate_limit_exceeded` codes (or HTTP 429), including plan label and approximate retry minutes from `resets_at`.
> - Refactors `parseErrorResponse` to delegate to `formatCodexError`, so HTTP error responses with quota/rate-limit codes also return friendly messages instead of raw server text.
> - Behavioral Change: `CodexApiError` messages for quota and rate-limit conditions now surface plan and retry timing; raw JSON is no longer used as a fallback message.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized c6c1f65.</sup>
> 
>

### DIFF
--- a/packages/ai/src/providers/openai-codex-responses.ts
+++ b/packages/ai/src/providers/openai-codex-responses.ts
@@ -495,10 +495,21 @@ async function* mapCodexEvents(events: AsyncIterable<Record<string, unknown>>): 
 		if (!type) continue;
 
 		if (type === "error") {
-			const code = (event as { code?: string }).code || "";
-			const message = (event as { message?: string }).message || "";
-			throw new CodexApiError(`Codex error: ${message || code || JSON.stringify(event)}`, {
-				code: code || undefined,
+			const nested = (event as { error?: CodexErrorDetails }).error;
+			const topLevel = event as { code?: string; message?: string; status_code?: number };
+			const status = typeof topLevel.status_code === "number" ? topLevel.status_code : undefined;
+			const formatted = formatCodexError(
+				{
+					code: nested?.type ?? nested?.code ?? topLevel.code,
+					type: nested?.type ?? nested?.code ?? topLevel.code,
+					message: nested?.message ?? topLevel.message,
+					plan_type: nested?.plan_type,
+					resets_at: nested?.resets_at,
+				},
+				{ status, fallbackMessage: JSON.stringify(event) },
+			);
+			throw new CodexApiError(formatted.friendlyMessage || `Codex error: ${formatted.message}`, {
+				code: formatted.code,
 				payload: event,
 			});
 		}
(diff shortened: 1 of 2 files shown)


kevinjosethomas:

PR #876, opened 2026-08-07
fix(tui): enable negotiated tmux hyperlinks

### LINKED ISSUE #686: Allow OSC 8 hyperlinks in supported tmux sessions
## Problem

Prime Agent unconditionally disables terminal hyperlinks whenever it detects tmux or screen:

```ts
const inTmuxOrScreen = !!process.env.TMUX || term.startsWith("tmux") || term.startsWith("screen");
if (inTmuxOrScreen) {
    const trueColor = colorTerm === "truecolor" || colorTerm === "24bit";
    return { images: null, trueColor, hyperlinks: false };
}
```

This is conservative for older or unconfigured multiplexers, but modern tmux can preserve OSC 8 hyperlinks when its client advertises the `hyperlinks` feature. In that supported configuration, Prime Agent still renders Markdown links using the fallback `label (URL)` form instead of making the label clickable.

Source: `packages/tui/src/terminal-image.ts`

## Verified environment

- Prime Agent 0.7.0
- Ghostty 1.3.1
- tmux 3.5a
- macOS
- tmux `mouse` enabled
- tmux client terminal: `xterm-ghostty`
- tmux client features include `hyperlinks`

The tmux configuration declares:

```tmux
set -as terminal-features ",xterm-ghostty:RGB:hyperlinks"
```

After detaching and reattaching, `#{client_termfeatures}` reports `hyperlinks`.

## Reproduction

1. Start Prime Agent inside the supported tmux configuration above.
2. Render a Markdown link such as `[Prime Agent](https://github.com/PrimeIntellect-ai/prime-agent)`.
3. Observe that Prime Agent prints the URL in parentheses rather than emitting OSC 8 for the label.
4. Confirm that `detectCapabilities()` returns `hyperlinks: false` solely because `TMUX` is present.

## Local verification

As a narrow local experiment, I changed the tmux/screen branch to enable hyperlinks when both `TMUX` and Ghostty's `GHOSTTY_RESOURCES_DIR` marker are present.

The installed bundled renderer and module renderer then both reported:

```text
{ images: null, trueColor: true, hyperlinks: true }
```

A private tmux test confirmed that:

- the OSC 8 destination was preserved in tmux;
- the Markdown label rendered without the fallback URL;
- the label opened successfully in Ghostty using `Command+Shift+click` (`Shift` bypasses tmux mouse capture); and
- a negative control without the Ghostty marker retained `hyperlinks: false`.

Images remained disabled inside tmux.

## Expected behavior

Users with a hyperlink-capable tmux configuration should have a supported way to enable clickable Markdown labels without patching the installed bundle.

Possible approaches include:

- detecting tmux hyperlink capability when it can be established reliably;
- adding an explicit `terminal.hyperlinks` auto/on/off setting or environment override; or
- combining conservative auto-detection with an explicit opt-in.

The Ghostty-marker experiment proves the rendering path works, but it may be too narrow to be the general upstream solution. The default should remain conservative when hyperlink support cannot be established.

rbalachandar:
Fixes #686.

## Summary

- Detect tmux's negotiated `hyperlinks` feature via `tmux display-message`.
- Enable OSC 8 links only when tmux explicitly advertises support.
- Keep terminal images disabled under tmux and hyperlinks disabled under screen.
- Fall back safely when tmux is missing, fails, or exceeds the 250 ms probe timeout.

## Tests

- Focused terminal capability tests: 47/47.
- Full TUI test suite: 742/742.
- `npm run check` passes, including TypeScript, installer, and browser smoke checks.
- Manual tmux verification confirmed clickable OSC 8 output.

image

### DIFF
--- a/packages/tui/CHANGELOG.md
+++ b/packages/tui/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## [Unreleased]
 
+- Added OSC 8 hyperlink support for tmux clients that advertise the `hyperlinks` terminal feature ([#686](https://github.com/PrimeIntellect-ai/prime-agent/issues/686)).
+
 ## [0.7.1] - 2026-08-07
 
 ## [0.7.0] - 2026-08-05
--- a/packages/tui/src/terminal-image.ts
+++ b/packages/tui/src/terminal-image.ts
@@ -1,3 +1,5 @@
+import { spawnSync } from "node:child_process";
+
 export type ImageProtocol = "kitty" | "iterm2" | null;
 
 export interface TerminalCapabilities {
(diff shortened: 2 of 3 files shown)


kevinjosethomas:

PR #889, opened 2026-08-07
fix(coding-agent): make continual harness state reachable from the prompt

### LINKED ISSUE #819: Continual harness memory is unreachable in practice: the prompt overview is a hardcoded 6-entry alphabetical head slice, and `rlm.harness` defaults to the session-local store, which is empty in a spawned child
## Summary

Two defects in the continual harness compound, and each is the reason the other has no workaround. They are filed together because the mitigation for either one is the other one.

**A. The system-prompt overview renders an alphabetical head slice of 6 entries per kind at a hardcoded cap.** `formatHarnessStateForPrompt()` renders the continual harness overview into the system prompt using a default cap of 6 entries per kind and a 180-character body clip. The 6 that survive are chosen by an alphabetical sort on `[path, title, id]` — not recency, version, or relevance — so which lessons the model sees is decided by how their file paths happen to be spelled. The function's options object already accepts `maxEntriesPerKind`/`maxContentLength`, but neither production call site passes them, and grepping the published sources turns up no setting, env var, or config key that reaches them — so in practice the limits behave as hardcoded. (If there is a supported way to raise them that we missed, that would resolve most of A.)

**B. `rlm.harness` resolves to the session-local store, which is empty for a spawned child.** `rlm.harness` resolves to the **session-local** harness store by default, and for an `rlm()`-spawned child that store is new and empty at spawn. A child that does the natural thing — `rlm.harness.overview()` or `rlm.harness.list("memory")` — gets `Harness state (local): ...` with `memory: 0`, even though the same child's system prompt was built from the *merged* global + local state and lists global memory entries. Nothing in the local output mentions that a global store exists, how many entries it holds, or that `global_=True` exists.

The prompt text itself says the overview lines are "compact summaries, not full descriptions ... inspect ... the underlying continual harness entry only when detail matters". **A** is why the summaries are not sufficient; **B** is why the inspect path returns zero. Together, there is no obvious path by which a spawned child reads what the global store holds.

To be precise about what is *not* broken, since it bounds the ask:

- Neither defect is silent in the strict sense. A emits an overflow line per kind; B correctly labels its output `(local)`. But the model has no in-prompt way to reach the hidden entries and no documented way to raise the cap, and the local output names neither the global scope nor the flag that reads it.
- Global reads work fine from a child when asked for explicitly. `RLM_GLOBAL_HARNESS_STATE_DIR` is exported to every kernel including children, so `rlm.harness.list("memory", global_=True)`, `rlm.harness.overview(global_=True)`, and `rlm.harness.get("memory", "global:<id>")` all resolve correctly (verified — see repro B). Defect B is purely that the **no-argument default is local-only and silent about it** — which is plausibly the call a child makes first.

A third, sharper gap falls out of B's design: parent-**local** entries do not reach children at all, while the system prompt explicitly recommends local scope for "session coordination".

Defect A looks like the same pattern as #799 (`rlm.find_models()` returning an alphabetical head slice): a head slice of an alphabetical ordering presented where a representative sample is expected.

## Environment

- prime-agent 0.7.0, installed from the published npm package `prime-agent`
- Components: the coding-agent workspace package `@earendil-works/pi-coding-agent` at `packages/coding-agent` (continual harness / `/refine` overview injection into the system prompt) and `prime-agent-runtime` (the Python `rlm` package installed into the kernel venv)
- Claims below were checked against `main` at `b9a4461149419156599d60174dddf15458e2b9ee` ("fix(coding-agent): show MCP login path for websearch (#809)"). The installed `prime-agent-runtime/src/rlm/harness.py` and `__init__.py` are byte-identical to that ref.

### A. Prompt overview cap

The behavior is fully determined by the exported function, so it reproduces without a live session. Save as `packages/coding-agent/test/overview-cap.test.ts` and run with `npm test` (vitest):

```ts
import { expect, it } from "vitest";
import { formatHarnessStateForPrompt, type HarnessEntry } from "../src/core/refinement/index.js";

const mk = (id: string, title: string, path: string, content: string, version = 1): HarnessEntry => ({
  id, kind: "memory", title, content, path, scope: "global", version,
  reference: {}, arguments: {}, metadata: {}, source: "test",
  created_at: new Date().toISOString(), updated_at: new Date().toISOString(),
});

it("renders only the 6 alphabetically-first memory entries", () => {
  const entries: Record<string, HarnessEntry> = {};

  // 48 pre-existing entries spread across the alphabet.
  for (let i = 0; i < 48; i++) {
    const letter = String.fromCharCode(97 + (i % 26));
    const id = `mem-${String(i).padStart(3, "0")}`;
    entries[id] = mk(id, `${letter}-lesson-${i}`, `memory/${letter}/${letter}-lesson-${i}.md`,
      `Older lesson ${i}. `.repeat(12));
  }

  // The newest entry: highest version, written last, but its path starts with "w".
  entries["mem-new"] = mk("mem-new", "workspace-resolver-fix", "memory/w/workspace-resolver-fix.md",
    "CRITICAL, just learned: the resolver error is caused by a stale lockfile. Re-resolve from a clean cache.", 7);

  const state = { entries: { prompt: {}, memory: entries, skill: {}, subagent: {} }, refinements: [] };
  const out = formatHarnessStateForPrompt(state as never);

  const shown = out.split("\n").filter((l) => l.startsWith("- [global:"))
    .map((l) => l.match(/\[global:([\w-]+)\]/)![1]);

  console.log("rendered:", shown.join(", "));
  expect(shown).toHaveLength(6);            // 6 of 49
  expect(shown).toContain("mem-new");       // FAILS — newest entry is not rendered
});
```

### B. A child reads an empty store

The child kernel environment is fully determined by `_rlmKernelEnv()`, so this reproduces standalone without spawning — set `RLM_HARNESS_STATE_DIR` to an empty directory (what a fresh child gets) and `RLM_GLOBAL_HARNESS_STATE_DIR` to a populated one (what every kernel gets):

1. Create two harness dirs, one empty and one holding 40 memory entries:

```bash
mkdir -p /tmp/child-harness /tmp/global-harness
python - <<'PY'
import json
empty = {"entries": {"prompt": {}, "memory": {}, "skill": {}, "subagent": {}}, "refinements": []}
json.dump(empty, open("/tmp/child-harness/harness_state.json", "w"))

full = {"entries": {"prompt": {}, "memory": {}, "skill": {}, "subagent": {}}, "refinements": []}
for i in range(40):
    eid = f"mem-{i:03d}"
    full["entries"]["memory"][eid] = {
        "id": eid, "kind": "memory", "title": f"Memory {i}",
        "content": f"Durable fact number {i}. " * 20,
        "path": "global/memory", "version": 1, "scope": "global",
        "metadata": {}, "arguments": {}, "reference": {},
        "created_at": "2026-01-01T00:00:00Z", "updated_at": "2026-01-01T00:00:00Z",
    }
json.dump(full, open("/tmp/global-harness/harness_state.json", "w"))
PY
```

2. Run the calls a child would make, with the kernel env a child receives:

```bash
RLM_HARNESS_STATE_DIR=/tmp/child-harness \
RLM_GLOBAL_HARNESS_STATE_DIR=/tmp/global-harness \
python -c "
from rlm.harness import get_harness_state
s = get_harness_state()                       # what rlm.harness resolves to
print(s.overview())
print('default list:', len(s.list(\"memory\")))
print('global  list:', len(get_harness_state(global_=True).list(\"memory\")))
"
```

Equivalently, in a session whose global store holds 40+ memory entries: spawn `handle = await rlm("Check harness memory for what we already know about X, then report")` and have the child run `print(rlm.harness.overview())` as its first step.

## Expected

**A.** The overview that reaches the system prompt should either (a) include recently written or high-priority entries with a guaranteed slot, so a lesson recorded by `/refine` is visible to the next session, or (b) let the operator raise the cap when the store outgrows it. Ideally both.

**B.** The default local overview should not read as "the harness is empty" when a populated global store is one keyword argument away. A child asking the harness what it knows should either be shown the global store or be told, in the same output, that a global store exists with N entries and how to read it.

### A

Selection is an alphabetical head slice at a fixed depth of 6, with bodies hard-cut at 180 characters. Measured output of the repro above (49-entry store):

```
rendered: mem-000, mem-026, mem-001, mem-027, mem-002, mem-028
- +43 more memory entries
```

The render order is strictly alphabetical by path (`memory/a/a-lesson-0.md`, `memory/a/a-lesson-26.md`, `memory/b/b-lesson-1.md`, …). The entry written most recently, at `version 7`, is absent. A newly recorded lesson is included only if its path sorts into the first 6 for its kind — a property of the filename, unrelated to how recently or how urgently it was learned.

Measured on that same fixture, comparing the default render against the same call with `maxEntriesPerKind`/`maxContentLength` raised past the store size:

| render | chars |
| --- | --- |
| default (6 entries / 180 chars) | 3,951 |
| uncapped | 15,290 |

So the prompt carried 25.8% of the stored harness text for this fixture. That percentage is fixture-dependent: bodies here are only ~191-203 characters, barely over the 180-char clip, so nearly all of the loss is the 43 dropped entries rather than truncation. A store with longer entries retains proportionally less, since the surviving entry count stays fixed at 6 and each surviving body is still cut at 180.

### B

Measured output of step 2, verbatim:

```
Harness state (local): /tmp/child-harness/harness_state.json
Call contract: installed Python skills use await <skill_import>(...) or a matching shell CLI; ...
prompt: 0
memory: 0
skill: 0
subagent: 0
refinements: 0
default list: 0
global  list: 40
```

`memory: 0` against a global store holding 40 entries. There is no `global_=True` anywhere in the output, no global path, and no count. Meanwhile that same child's system prompt was rendered from global + local merged state and does list global memory entries — so the child is presented with two contradictory views of its own memory, one of which it obtained by direct measurement.

Aggravating factor: **parent-local entries do not reach children.** Each session's local store lives under its own session-artifact dir, and neither `_rlmKernelEnv()` nor `_createInlineRlmSubagentRuntime()` passes or exports a parent harness dir — yet the prompt instructs agents to "Default to local continual harness refinement for ... session coordination" (refinement.ts L450). Coordination notes written at the *recommended* scope are structurally invisible to the children they are meant to coordinate.

## Code pointers

Line numbers are against `main` at the ref above.

### A. Overview render cap

**`packages/coding-agent/src/core/refinement/refinement.ts`** — `formatHarnessStateForPrompt` (L429), constants at L26-28:

```ts
const DEFAULT_OVERVIEW_ENTRY_LIMIT = 6;
const DEFAULT_OVERVIEW_REFINEMENT_LIMIT = 5;
const DEFAULT_OVERVIEW_CONTENT_LIMIT = 180;
```

The sort key (L467-469) is purely lexicographic — no recency, version, or relevance term:

```ts
const entries = Object.values(state.entries[kind]).sort((a, b) =>
    [a.path, a.title, a.id].join("\0").localeCompare([b.path, b.title, b.id].join("\0")),
);
```

followed by the head slice at L481 and the overflow notice at L497-500:

```ts
for (const entry of entries.slice(0, maxEntriesPerKind)) {
```

```ts
const overflow = entries.length - Math.min(entries.length, maxEntriesPerKind);
if (overflow > 0) {
    lines.push(`- +${overflow} more ${kind} entries`);
}
```

Bodies pass through `compactText` (L421), which collapses whitespace and hard-cuts with an ellipsis:

```ts
function compactText(text: string, maxLength: number): string {
	const normalized = text.replace(/\s+/g, " ").trim();
	if (normalized.length <= maxLength) {
		return normalized;
	}
	return `${normalized.slice(0, Math.max(0, maxLength - 3))}...`;
}
```

**Why the limits are unreachable in practice.** `formatHarnessStateForPrompt` does expose the knobs (L432-434), defaulting at L440-442:

```ts
const maxEntriesPerKind = options.maxEntriesPerKind ?? DEFAULT_OVERVIEW_ENTRY_LIMIT;
const maxContentLength = options.maxContentLength ?? DEFAULT_OVERVIEW_CONTENT_LIMIT;
```

But both production call sites in **`packages/coding-agent/src/core/system-prompt.ts`** (L106 and L141) pass only the example-inclusion flags:

```ts
prompt += `\n\n${formatHarnessStateForPrompt(harnessState, { includeIpythonExamples: hasIpython, includeShellExamples: hasBash, includeRefineExamples: hasIpython && hasRefineSkill })}`;
```

Grepping `maxEntriesPerKind|maxContentLength|DEFAULT_OVERVIEW_` across the published package sources returns hits only inside `refinement.ts` itself (plus its bundled copies) — no settings key, no env var, no docs reference. Is there a supported override we are missing?

**Recency data is already present and unused by the sort.** `HarnessEntry` (L34-48) carries both timestamps and a version, none of which participate in the ordering:

```ts
export interface HarnessEntry {
	id: string;
	kind: RefinementKind;
	title: string;
	content: string;
	path: string;
	...
	created_at: string;
	updated_at: string;
	version: number;
}
```

**Limits disagree with the Python side.** `prime-agent-runtime/src/rlm/harness.py`, `HarnessState.overview()` (L721, class defined at L141) defaults to `max_entries_per_kind: int = 20` and clips bodies at 120 characters:

```python
summary = entry.content.strip().replace("\n", " ")
if len(summary) > 120:
    summary = f"{summary[:117]}..."
```

So the in-kernel view an agent gets from `harness.overview()` and the system-prompt view disagree on both dimensions (20/120 vs 6/180). An agent that inspects its own store in IPython sees a different, larger set than the one it was primed with — which makes the prompt view hard to reason about from inside the session, and which interacts directly with defect B: the in-kernel view that disagrees is also the one that defaults to the empty local store. Is the divergence intentional?

**Existing test coverage does not exercise the cap.** `packages/coding-agent/test/refinement.test.ts` L607 calls `formatHarnessStateForPrompt(merged)` against a 2-entry merged state, so it never crosses either limit.

### B. Local-by-default resolution

**`prime-agent-runtime/src/rlm/__init__.py`** — `_HarnessProxy._resolve()` (L250-252). `rlm.harness` is this proxy; every attribute access lands here, and the resolving call takes no scope argument:

```python
def _resolve(self) -> HarnessState:
    try:
        return get_harness_state()
```

**`prime-agent-runtime/src/rlm/harness.py`** — `get_harness_state()` (L785-791) defaults to local:

```python
def get_harness_state(
    state_dir: str | Path | None = None, *, global_: bool = False, **kwargs: Any
) -> HarnessState:
    """Return the cached local harness state, or global when requested."""
    global_ = _resolve_global_flag(global_, kwargs)
    file_path = _state_file(state_dir, global_=global_)
    scope: HarnessScope = "global" if global_ else "local"
```

**`prime-agent-runtime/src/rlm/harness.py`** — `HarnessState.overview()` (L721-738). The header names the scope but never the *other* scope, and the counts come from `self.entries` only:

```python
def overview(self, *, max_entries_per_kind: int = 20, global_: bool = False, **kwargs: Any) -> str:
    if target := self._global_target(global_, kwargs):
        return target.overview(max_entries_per_kind=max_entries_per_kind)
    self._sync_from_disk()
    lines = [
        f"Harness state ({self.scope}): {self.file_path}",
        ...
    ]
    for kind in _KINDS:
        records = self.list(kind)[:max_entries_per_kind]
        lines.append(f"{kind}: {len(self.entries[kind])}")
```

`HarnessState.list()` (L425) and `HarnessState.snapshot()` (L770) have the same shape: `global_` defaults to `False`, and the local result is returned with no indication that a global store was skipped.

**`packages/coding-agent/src/core/agent-session.ts`** — `_rlmKernelEnv()` (L8799-8818) exports both dirs to every kernel, which is why `global_=True` works and why the local dir is the child's own (comments elided):

```ts
const env: Record<string, string> = {
    RLM_DEPTH: String(this._rlmDepth),
    RLM_MAX_DEPTH: String(this._rlmMaxDepth),
    RLM_GLOBAL_HARNESS_STATE_DIR: getGlobalHarnessStateDir(),
};
const rlmSessionDir = this._ensureRlmSessionDir();
if (rlmSessionDir) {
    env.RLM_SESSION_DIR = rlmSessionDir;
    env.RLM_HARNESS_STATE_DIR = this._localHarnessStateDir() ?? getLocalHarnessStateDir(rlmSessionDir)!;
}
```

**`packages/coding-agent/src/core/agent-session.ts`** — `_localHarnessStateDir()` (L7165-7170) resolves to the session's own artifact dir, which for a freshly spawned child is new and empty:

```ts
private _localHarnessStateDir(): string | undefined {
    return (
        getLocalHarnessStateDir(this.sessionManager.getSessionArtifactDir()) ??
        (this._rlmSessionDir ? getLocalHarnessStateDir(this._rlmSessionDir) : undefined)
    );
}
```

**`packages/coding-agent/src/core/agent-session.ts`** — `_loadMergedHarnessState()` (L7547-7554), fed to `buildSystemPrompt` at L4288 as `harnessState: this._loadMergedHarnessState()`. This is the path that gives the child a *prompt* view including global entries, and the reason the two views disagree:

```ts
/** Global harness state overlaid with this session's local state, when persisted. */
private _loadMergedHarnessState(): HarnessState {
    const localHarnessStateDir = this._localHarnessStateDir();
    return mergeHarnessStates(
        loadHarnessState(getGlobalHarnessStateDir(), "global"),
        localHarnessStateDir ? loadHarnessState(localHarnessStateDir, "local") : undefined,
    );
}
```

**`packages/coding-agent/src/core/agent-session.ts`** — `_createInlineRlmSubagentRuntime()` (L8948) builds the child with `SessionManager.create(this._cwd, options.sessionDir)` (L8949) and passes `rlmSessionDir: options.sessionDir` (L8999). No parent harness dir is passed, so the child's local store is unrelated to the parent's.

**`packages/coding-agent/src/core/refinement/refinement.ts`** — guidance line at L450:

```
"Default to local continual harness refinement for current task progress, temporary blockers, and session coordination. Use global continual harness refinement only for stable cross-session lessons, durable user preferences, reusable skills/subagents, or explicitly project-qualified facts."
```

**`packages/coding-agent/docs/rlm.md`** (145 lines) — `grep -ci harness` and `grep -c global_` both return `0`. The local/global scope model and the `global_=True` keyword are not documented on the page an agent is most likely to read.

### A. Overview render cap

1. **Plumb the existing options through.** The parameters already exist; add an overview settings block read by the settings manager and forward it at both `system-prompt.ts` call sites. This alone unblocks operators with large stores and is a small change.
2. **Make selection recency- or priority-aware.** `updated_at` and `version` are already on every `HarnessEntry`. Sorting by `updated_at` descending before the slice — or reserving a portion of the budget for the N most recently updated entries and filling the rest alphabetically — gives every newly refined entry a slot without changing the schema. Keeping a stable alphabetical order *within* the selected set preserves prompt-cache friendliness.
3. **Reconcile with the Python defaults** in `HarnessState.overview()` so the in-kernel and system-prompt views agree, or document why they differ.
4. Consider raising the shipped defaults. 6 entries per kind is low relative to what a store accumulates after even moderate `/refine` use.
5. Optionally, make the overflow actionable by naming the mechanism to reach hidden entries in the overflow line itself (for example pointing at `harness.overview()` or a lookup by kind), so the model has a next step rather than just a count. Note that this is only useful once B is fixed — today that pointer leads to an empty local store.

### B. Local-by-default resolution

6. **Make the local view self-describing.** In `HarnessState.overview()`, when `self.scope == "local"` and a global store resolves, append a line naming the global path, its per-kind counts, and the exact call to read it — e.g. after the `Harness state (local): ...` header:

```
NOTE: this is the LOCAL store. The GLOBAL store (<global path>) holds prompt: N, memory: N, skill: N, subagent: N.
Read it with rlm.harness.overview(global_=True) / rlm.harness.list("memory", global_=True) / rlm.harness.get(kind, "global:<id>").
```

   Local overviews currently hide the (usually larger) global store, and this is the one output an agent is most likely to read first. The same hint is worth attaching to an empty-result `list()` when the global store is non-empty.

7. **Document the scope model in `packages/coding-agent/docs/rlm.md`**, including `global_=True` and the `"global:<id>"` prefix form that `_strip_scope_prefix` (harness.py L59) already accepts — right now the flag appears only in the Python source.

8. **Close the parent-local gap.** Either export a `RLM_PARENT_HARNESS_STATE_DIR` from `_rlmKernelEnv()` and merge it read-only in `_loadMergedHarnessState()` (with scope rendered as `parent:` so children do not try to write it), or amend the refinement.ts L450 guidance to stop recommending local scope for "session coordination" — as written, following the recommendation makes those notes unreadable by the children being coordinated.

9. Optionally, have `rlm()` inline a compact global-memory digest into the child's spawn brief for children whose task mentions prior knowledge. The spawn brief is the one channel that can carry full entry bodies to a child at spawn time, and the harness inlines nothing into it today.

## Impact

The continual harness is the mechanism by which `/refine` is supposed to make lessons persist across sessions and across agents. Both defects work against that, and each blocks the other's mitigation.

From A: with an alphabetical cap of 6, a store that has accumulated a few dozen memories surfaces a near-fixed subset determined by filename spelling, and newly written lessons are the *least* likely to appear — they are added to a store that is already past the cap, and nothing about being new helps them rank. In the fixture above the prompt carried 25.8% of the stored harness text, and a real store with longer bodies would carry less. Every spawned child is primed from the same overview, so all children inherit the identical alphabetical head slice.

From B: the failure mode is a confident empty answer. A child asked to "check what we already know before starting" gets `memory: 0` from the default call, and unless it independently knows to pass `global_=True` it has no signal that a populated global store exists — while the parent, reading its own populated overview, has no signal that the child read an empty one. The parent-local gap compounds it: the system prompt steers coordination notes into a scope that children cannot read, so following the documented guidance reduces what children can see.

Combined: the prompt view is a small fixed slice and tells the agent to inspect the store for detail; the default inspect path returns zero. For long-running multi-agent operation the accumulated global store is a large part of what makes delegation cheaper than doing the work inline, and neither default read path delivers it.

Related: #799 (same alphabetical-head-slice pattern in `rlm.find_models()`), #769 (server-backed long-term memory — different storage backend, same "children cannot get at what the fleet knows" surface), #762 (skill API surface not discoverable — adjacent to the documentation half of this).

Hotragn:
Fixes #819.

Two defects compound so that a spawned child has no reachable path to the continual harness state its own system prompt was built from. Each is the workaround for the other, so they are fixed together.

## A. The prompt overview kept the alphabetically-first entries

`formatHarnessStateForPrompt` sorted every kind by `[path, title, id]` and kept the first 6. Which lessons reached the model was decided by how their paths happened to be spelled — a memory written seconds ago under `memory/w/...` lost its slot to 48 older entries under `memory/a/...`. The options object already accepted `maxEntriesPerKind`/`maxContentLength`, but neither production call site in `system-prompt.ts` passed them and no setting reached them, so the caps were hardcoded in practice.

- Entries are now ranked by recency: `updated_at`, falling back to `created_at` when it is unparseable, then `version`, then the old `[path, title, id]` comparison purely to break exact ties. Selection and render order are both recency-first, so the newest lesson is nearest the model and the output stays deterministic for a given state.
- A new `harnessOverview` setting (`maxEntriesPerKind`, `maxContentLength`) reaches both call sites through `SettingsManager.getHarnessOverviewSettings()` and `BuildSystemPromptOptions`. Values are clamped in `formatHarnessStateForPrompt` (0-500 entries, 40-4000 chars) so a bad settings value cannot blow up or collapse the prompt.
- `overviewForPrompt`, the refiner's own input, had the same defect in a milder form: an unordered 40-entry head slice. It now uses the same ranking, so a truncated list still shows the refiner the entries most likely to need an edit.

The overflow line was `- +42 more memory entries` while the surrounding prompt instructed the model to "inspect the underlying continual harness entry only when detail matters". It named a number and no way to act on it. It now names the call:

```
- +42 more memory entries not shown: read them with `rlm.harness.list("memory")` for this session's local store and `rlm.harness.list("memory", global_=True)` for the cross-session global store, or raise `harnessOverview.maxEntriesPerKind` in settings.json
```

Sessions without IPython get the settings half only, matching the existing `includeIpythonExamples` gating.

## B. `rlm.harness` was silent about the global store

`rlm.harness` resolves to the session-local store. For an `rlm()`-spawned child that store is new and empty, so `overview()` returns `memory: 0` — while the same child's system prompt was built from the merged global + local state and lists global entries. Nothing in the output mentioned that a global store existed, how many entries it held, or that `global_=True` reads it.

A local overview now ends with:

```
global store (not listed above): /home/u/.prime/agent/harness/harness_state.json
  prompt: 0, memory: 40, skill: 0, subagent: 0, refinements: 0
  read it with overview(global_=True), list(<kind>, global_=True), or get(<kind>, 'global:<id>'); pass include_global=True to overview()/list() for the merged view this session's system prompt was built from
```

`overview()` and `list()` also accept `include_global=True`, which returns the same union the host merges to build the prompt — local entries shadow global entries on id collision, matching `mergeHarnessStates`. The pointer is emitted only when a distinct global store resolves, so a global store never points at itself and a local store aliased onto the global file stays quiet.

Running the issue's repro B verbatim against this branch:

```
Harness state (local): /tmp/child-harness/harness_state.json
...
memory: 0
refinements: 0
global store (not listed above): /tmp/global-harness/harness_state.json
  prompt: 0, memory: 40, skill: 0, subagent: 0, refinements: 0
  read it with overview(global_=True), list(<kind>, global_=True), ...
```

## Deliberately not in scope

The issue's third observation — parent-*local* entries never reach children, while the prompt recommends local scope for session coordination — is a scope-inheritance design change rather than a reachability fix, and would need its own decision about whether a child sees a read-only view of its parent's local store. Left for a separate issue.

## Tests

`packages/coding-agent/test/suite/regressions/819-harness-overview-reachability.test.ts` (7 cases): the issue's repro A (the newest entry now renders, first, out of 49); `created_at` fallback and tie-break determinism; the overflow hint in both IPython and non-IPython prompts; clamping of negative, `NaN`, and oversized budgets; `SettingsManager` round-trip; `buildSystemPrompt` honoring the option; and an end-to-end case that writes a global harness file, boots a session with `harnessOverview.maxEntriesPerKind: 12`, and asserts the live `session.systemPrompt` renders all 12.

`prime-agent-runtime/test/test_harness.py` (3 cases): the local overview names the global store and its counts; `include_global=True` merges with local shadowing while `global_=True` still reads global alone; and no pointer is emitted when no distinct global store exists.

Verified: `npm run check` clean; `prime-agent-runtime` 38/38; `test/refinement.test.ts`, `test/system-prompt.test.ts`, `test/settings-manager.test.ts` pass except one pre-existing Windows-only path-separator assertion in `refinement.test.ts` that also fails on `main`.

> [!NOTE]
> ### Make continual harness state reachable from the system prompt via recency ranking and global store pointers
> - Harness overview entries in the system prompt are now ranked by recency (most-recently-updated first) instead of alphabetical path order, using new helpers in [refinement.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/889/files#diff-17b614d876315ad8ac0165562a8f07a5c4ada3b28928eb02d90f6b93a96ea91d).
> - Overflow lines now include explicit `rlm.harness.list` / `rlm.harness.overview` read instructions so agents know how to access withheld entries.
> - `rlm.harness.overview()` and `rlm.harness.list()` in [harness.py](https://github.com/PrimeIntellect-ai/prime-agent/pull/889/files#diff-441d64dfc5d74f553c86b609beff62210693b468d021d8bded095dea768138cf) gain an `include_global=True` parameter that merges the global peer store, with local entries shadowing global ones.
> - Local overviews now append a pointer to the global store (path, per-kind counts, and read guidance) when a distinct global store exists.
> - New `harnessOverview.maxEntriesPerKind` and `harnessOverview.maxContentLength` settings are exposed via `SettingsManager` and forwarded through `buildSystemPrompt`, with clamping deferred to the formatter.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 7534c8b.</sup>
> 
>

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -20,6 +20,10 @@
 - Fixed workers with no live connection reporting as `ready`; stopping workers now report a `stopping` state, are hidden from live sessions, and no longer receive daemon-wide commands ([#850](https://github.com/PrimeIntellect-ai/prime-agent/pull/850)).
 - Fixed timed-out worker stops stranding dead-but-registered workers ("Session worker is not connected"); stops now finalize in the background once the process exits, and zombie processes are no longer counted as alive ([#851](https://github.com/PrimeIntellect-ai/prime-agent/pull/851)).
 - Fixed sessions becoming permanently unopenable after a stale worker registration was left behind; open/resume now self-heals by finishing the old cleanup and starting a fresh worker ([#852](https://github.com/PrimeIntellect-ai/prime-agent/pull/852)).
+- Changed the continual harness overview to rank entries by recency instead of an alphabetical path sort, so the newest lessons survive truncation ([#819](https://github.com/PrimeIntellect-ai/prime-agent/issues/819))
+- Added the `harnessOverview` setting to raise the continual harness overview's per-kind entry cap and body length ([#819](https://github.com/PrimeIntellect-ai/prime-agent/issues/819))
+- Fixed the continual harness overview naming withheld entries by count alone, with no call that reads them ([#819](https://github.com/PrimeIntellect-ai/prime-agent/issues/819))
+- Added `include_global=True` to `rlm.harness.overview()` and `rlm.harness.list()`, and made a local overview name the global store, its counts, and how to read it ([#819](https://github.com/PrimeIntellect-ai/prime-agent/issues/819))
 
 ## [0.7.1] - 2026-08-07
 
--- a/packages/coding-agent/docs/rlm-runtime.md
+++ b/packages/coding-agent/docs/rlm-runtime.md
@@ -212,6 +212,18 @@ Session-local state lives in the session artifact directory under `harness/harne
 
 `/refine` runs a dedicated review over the current trajectory and applies small create/update/delete edits. Rollback uses recorded before/after snapshots. The base system prompt remains immutable; refinements are supplemental state.
 
+### Reading state the prompt did not show
+
+The system-prompt overview is a truncated, most-recently-updated-first sample of the merged local + global state, capped by the `harnessOverview` setting. `rlm.harness` resolves to the session-local store, which is empty for a freshly spawned child, so a local read is not a view of everything the prompt was built from:
+
+```python
+rlm.harness.overview()                            # local store; names the global store and its counts
+rlm.harness.overview(global_=True)                # cross-session store only
+rlm.harness.overview(include_global=True)         # both, matching what built the system prompt
+rlm.harness.list("memory", include_global=True)   # merged entries; local shadows global on id collision
+rlm.harness.get("memory", "global:<id>")          # a single entry by its displayed id
+```
+
 ## Goal Requests
 
 The bundled `goal` Python skill is a thin host-bridge client:
--- a/packages/coding-agent/docs/settings.md
+++ b/packages/coding-agent/docs/settings.md
@@ -126,6 +126,26 @@ prime-agent --offline
 }
 ```
 
+### Continual Harness Overview
+
+Budgets for the continual harness overview injected into the system prompt. Entries are listed most-recently-updated first, so raising the cap adds older entries rather than reshuffling the ones already shown.
+
+| Setting | Type | Default | Description |
+|---------|------|---------|-------------|
+| `harnessOverview.maxEntriesPerKind` | number | `6` | Entries rendered per kind (prompt/memory/skill/subagent), clamped to 0-500 |
+| `harnessOverview.maxContentLength` | number | `180` | Characters of each entry's body, clamped to 40-4000 |
+
+```json
+{
+  "harnessOverview": {
+    "maxEntriesPerKind": 20,
+    "maxContentLength": 400
+  }
+}
+```
+
+Entries beyond the cap are not lost. The overview reports how many were withheld, and the kernel can read them with `rlm.harness.list("<kind>")` for the session-local store, `rlm.harness.list("<kind>", global_=True)` for the cross-session store, or `rlm.harness.list("<kind>", include_global=True)` for the merged view the prompt was built from.
+
 ### Branch Summary
 
 | Setting | Type | Default | Description |
--- a/packages/coding-agent/src/core/agent-session.ts
+++ b/packages/coding-agent/src/core/agent-session.ts
@@ -4310,6 +4310,7 @@ export class AgentSession {
 			rlmDepth: this._rlmDepth,
 			rlmParentAgent: this._rlmParentAgent,
 			harnessState: this._loadMergedHarnessState(),
+			harnessOverview: this.settingsManager.getHarnessOverviewSettings(),
 		};
 		return buildSystemPrompt(this._baseSystemPromptOptions);
 	}
--- a/packages/coding-agent/src/core/refinement/refinement.ts
+++ b/packages/coding-agent/src/core/refinement/refinement.ts
@@ -26,6 +26,12 @@ const REFINEMENT_HISTORY_FILE_NAME = "refinements.jsonl";
 const DEFAULT_OVERVIEW_ENTRY_LIMIT = 6;
 const DEFAULT_OVERVIEW_REFINEMENT_LIMIT = 5;
 const DEFAULT_OVERVIEW_CONTENT_LIMIT = 180;
+const MIN_OVERVIEW_ENTRY_LIMIT = 0;
+const MAX_OVERVIEW_ENTRY_LIMIT = 500;
+const MIN_OVERVIEW_CONTENT_LIMIT = 40;
+const MAX_OVERVIEW_CONTENT_LIMIT = 4_000;
+/** Entries per kind handed to the refiner, which sees far more context than the system prompt. */
+const REFINER_OVERVIEW_ENTRY_LIMIT = 40;
 
 export type RefinementKind = "prompt" | "memory" | "skill" | "subagent";
 export type RefinementAction = "create" | "update" | "delete";
@@ -426,6 +432,61 @@ function compactText(text: string, maxLength: number): string {
 	return `${normalized.slice(0, Math.max(0, maxLength - 3))}...`;
 }
 
+/** Configurable overview budgets, surfaced to users as the `harnessOverview` setting. */
+export interface HarnessOverviewLimits {
+	maxEntriesPerKind?: number;
+	maxContentLength?: number;
+}
+
+function entryRecency(entry: HarnessEntry): number {
+	const updated = Date.parse(entry.updated_at);
+	if (Number.isFinite(updated)) {
+		return updated;
+	}
+	const created = Date.parse(entry.created_at);
+	return Number.isFinite(created) ? created : 0;
+}
+
+/**
+ * Overview slots are scarce, so the entries that survive truncation must be the ones
+ * most likely to matter rather than the ones whose paths sort first: most recently
+ * written, then most refined. The trailing path/title/id comparison only breaks exact
+ * ties, keeping the rendered order stable across builds of the same state.
+ */
+function compareHarnessEntriesForOverview(a: HarnessEntry, b: HarnessEntry): number {
+	const recency = entryRecency(b) - entryRecency(a);
+	if (recency !== 0) {
+		return recency;
+	}
+	if (b.version !== a.version) {
+		return b.version - a.version;
+	}
+	return [a.path, a.title, a.id].join("\0").localeCompare([b.path, b.title, b.id].join("\0"));
+}
+
+function rankHarnessEntriesForOverview(entries: Record<string, HarnessEntry>): HarnessEntry[] {
+	return Object.values(entries).sort(compareHarnessEntriesForOverview);
+}
+
+/**
+ * The overview is the only place the model learns that hidden entries exist, so the
+ * overflow line has to name the call that reads them. Without it the prompt advises
+ * inspecting the underlying entry while leaving no reachable path to do so.
+ */
+function overflowReadHint(kind: RefinementKind, includeIpythonExamples: boolean): string {
+	if (!includeIpythonExamples) {
+		return `raise \`harnessOverview.maxEntriesPerKind\` in settings.json to show more ${kind} entries`;
+	}
+	return `read them with \`rlm.harness.list("${kind}")\` for this session's local store and \`rlm.harness.list("${kind}", global_=True)\` for the cross-session global store, or raise \`harnessOverview.maxEntriesPerKind\` in settings.json`;
+}
+
+function clampOverviewLimit(value: number | undefined, fallback: number, min: number, max: number): number {
+	if (typeof value !== "number" || !Number.isFinite(value)) {
+		return fallback;
+	}
+	return Math.min(max, Math.max(min, Math.trunc(value)));
+}
+
 export function formatHarnessStateForPrompt(
 	state: HarnessState,
 	options: {
@@ -437,9 +498,19 @@ export function formatHarnessStateForPrompt(
 		includeRefineExamples?: boolean;
 	} = {},
 ): string {
-	const maxEntriesPerKind = options.maxEntriesPerKind ?? DEFAULT_OVERVIEW_ENTRY_LIMIT;
+	const maxEntriesPerKind = clampOverviewLimit(
+		options.maxEntriesPerKind,
+		DEFAULT_OVERVIEW_ENTRY_LIMIT,
+		MIN_OVERVIEW_ENTRY_LIMIT,
+		MAX_OVERVIEW_ENTRY_LIMIT,
+	);
 	const maxRefinements = options.maxRefinements ?? DEFAULT_OVERVIEW_REFINEMENT_LIMIT;
-	const maxContentLength = options.maxContentLength ?? DEFAULT_OVERVIEW_CONTENT_LIMIT;
+	const maxContentLength = clampOverviewLimit(
+		options.maxContentLength,
+		DEFAULT_OVERVIEW_CONTENT_LIMIT,
+		MIN_OVERVIEW_CONTENT_LIMIT,
+		MAX_OVERVIEW_CONTENT_LIMIT,
+	);
 	const includeIpythonExamples = options.includeIpythonExamples ?? true;
 	const includeRefineExamples = options.includeRefineExamples ?? includeIpythonExamples;
 	const lines = [
@@ -447,6 +518,7 @@ export function formatHarnessStateForPrompt(
 		"",
 		"Local continual harness entries belong to this Prime Agent session. Global continual harness entries persist across Prime Agent sessions.",
 		"The continual harness entries below are compact summaries, not full descriptions. Use them as routing/context hints; inspect or refine the underlying continual harness entry only when detail matters.",
+		`Each kind lists its most recently updated entries first, capped at ${maxEntriesPerKind} per kind. A "+N more" line means the remaining entries exist but are not shown here; read them from the store instead of assuming they are absent.`,
 		"Default to local continual harness refinement for current task progress, temporary blockers, and session coordination. Use global continual harness refinement only for stable cross-session lessons, durable user preferences, reusable skills/subagents, or explicitly project-qualified facts.",
 		"Use these continual harness prompt notes, memories, skills, and subagent specs when they are relevant. The base system prompt is immutable; prompt entries below are supplemental notes only.",
 		"",
@@ -464,9 +536,7 @@ export function formatHarnessStateForPrompt(
 
 	let totalEntries = 0;
 	for (const kind of Object.keys(state.entries) as RefinementKind[]) {
-		const entries = Object.values(state.entries[kind]).sort((a, b) =>
-			[a.path, a.title, a.id].join("\0").localeCompare([b.path, b.title, b.id].join("\0")),
-		);
+		const entries = rankHarnessEntriesForOverview(state.entries[kind]);
 		totalEntries += entries.length;
 		// Render subagent specs as a task-shaped roster the model can match against — the
 		// analogue of Claude Code's agent-type menu — rather than a bare count. In
@@ -496,7 +566,7 @@ export function formatHarnessStateForPrompt(
 		}
 		const overflow = entries.length - Math.min(entries.length, maxEntriesPerKind);
 		if (overflow > 0) {
-			lines.push(`- +${overflow} more ${kind} entries`);
+			lines.push(`- +${overflow} more ${kind} entries not shown: ${overflowReadHint(kind, includeIpythonExamples)}`);
 		}
 		lines.push("");
 	}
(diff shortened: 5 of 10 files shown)


kevinjosethomas:

PR #890, opened 2026-08-07
fix(coding-agent): defer and report heartbeat fires a busy session declines

### LINKED ISSUE #820: Heartbeat fires that coincide with a busy session are dropped, not deferred, and the delivered prompt has no run counter or skip notice
## Summary

When a scheduled heartbeat (`source: "heartbeat"` or `"rlm_heartbeat"`) comes due while its session is busy, the fire is dropped rather than deferred: `claimDueInState` advances `job.nextRunAt` at claim time, *before* any busy check, so the beat that `runCronJob` then declines is gone with no catch-up, no coalescing, and no backlog counter. Separately, the heartbeat prompt the model actually receives is static text — `createHeartbeatPromptMessage` sets `content: job.prompt` and puts `runCount`/`lastRunAt`/`nextRunAt` only in `details`, which `convertToLlm` drops. The net effect is that a heartbeat-driven agent cannot tell "beat 3 of 3" from "beat 3 of 40", and skipped beats are invisible to it.

Skips are barely visible to a human either. The only persisted evidence is `job.lastSkippedAt`, a single overwritten timestamp, and the only place it is rendered is `formatAgentCronJob`, which is used by the daemon CLI job listing. `/heartbeat status` does not surface it at all.

## Environment

- prime-agent 0.7.0, installed from npm as `prime-agent` (CLI `prime-agent`); upstream source package is `packages/coding-agent` (`@earendil-works/pi-coding-agent`)
- Components: cron/heartbeat scheduler (`core/cron-jobs.ts`), daemon dispatch (`modes/daemon/daemon-mode.ts`), message conversion (`core/messages.ts`), interactive command (`modes/interactive/interactive-mode.ts`)
- Line numbers below are against upstream `main` at commit `b9a4461149419156599d60174dddf15458e2b9ee`

## Reproduction

1. Start a daemon-backed session.
2. Register a heartbeat on a short interval:
   ```
   /heartbeat --every 5m Report which heartbeat number this is and what changed since the previous one.
   ```
   (Same result via the bundled `rlm-heartbeat` skill / `rlm_heartbeat.create`.)
3. In the same session, start a foreground bash command long enough to span two or more heartbeat boundaries:
   ```
   sleep 900
   ```
   This sets `session.isBashRunning`, which `shouldDeferHeartbeatCronJob` classifies as busy. A real long build/test run reproduces identically.
4. Let the bash finish and wait for the next heartbeat to be delivered.
5. Read the delivered heartbeat message text, then run `/heartbeat status`, then inspect `scheduled-jobs.json` (`SESSION_SCHEDULED_JOBS_FILENAME`, `packages/coding-agent/src/core/cron-jobs.ts:111`) in the session artifacts directory.

## Expected

- Fires that land during a busy window are either deferred/coalesced onto the next opportunity, or the next delivered heartbeat states that fires were missed.
- The delivered prompt carries the run counter and the previous delivery time. `runCount` and `lastRunAt` are already computed and already put in `HeartbeatPromptDetails` — they just need to reach the model.
- Some surface reports how many fires were skipped, not only when the most recent skip happened.

## Actual

- A fire that coincides with a busy session appears to be permanently lost. `claimDueInState` moves `nextRunAt` forward the moment the job is claimed; `runCronJob` then returns `"skipped"`; `recordDispatchResult`'s skip branch recomputes `nextRunAt` from `now` again rather than restoring the missed time. I could not find any path that re-arms the missed beat — `recoverInterruptedInState` only clears an outstanding dispatch and sets `lastError`, it does not restore `nextRunAt`. If there is a re-arm path I missed, please point at it.
- The delivered heartbeat prompt is byte-identical on every delivery. It contains no run number, no previous-delivery timestamp, and no notice that fires were skipped.
- The only skip evidence anywhere is `job.lastSkippedAt`, a single overwritten timestamp persisted to `scheduled-jobs.json`. It is rendered only by `formatAgentCronJob` as `skipped=<date>`, which is used by the daemon CLI cron listing (`packages/coding-agent/src/cli/daemon-command.ts:1019`). `/heartbeat status` renders Status / Every / Delivery / Instruction / Next / Last / Runs / Error (`showHeartbeat`, `packages/coding-agent/src/modes/interactive/interactive-mode.ts:9322`) and shows no skip information at all. The in-session agent receives none of it.
- Measurement from one run: on a 5-fire heartbeat job against a busy session, 2 of the 5 scheduled fires were skipped, including the job's first fire. The delivered beats carried no indication of the gap.

## Code pointers

**1. The beat is consumed before the busy check** — `claimDueInState`, `packages/coding-agent/src/core/cron-jobs.ts:1574` (snippet is `:1581`-`:1585`):

```ts
const scheduledFor = job.nextRunAt!;
const nextRunAt = nextRunAtForSchedule(job.schedule, claimedAt)?.toISOString();
const advanced = nextRunAt
    ? { ...job, nextRunAt, updatedAt: claimedAt.toISOString() }
    : withoutNextRunAt({ ...job, updatedAt: claimedAt.toISOString() });
```

`nextRunAt` is advanced unconditionally at claim time. Everything downstream can only decline; nothing appears to put the beat back.

The very next lines (`:1586`) are a third drop path that never reaches the busy check at all — if a dispatch for the same job is still outstanding, the newly due fire is discarded on the spot:

```ts
if (claimedJobIds.has(job.id)) {
    return { ...advanced, lastSkippedAt: claimedAt.toISOString() };
}
```

**2. Declining gates on the heartbeat path** — `DaemonMode.runCronJob`, `packages/coding-agent/src/modes/daemon/daemon-mode.ts:1557`:

```ts
if (isHeartbeatCronJob(runnableJob) && isAgentMessagePromptInProgress) {
    return "skipped";                                    // :1573
}
if (shouldDeferHeartbeatCronJob(runnableJob, session)) {
    return "skipped";                                    // :1576
}
...
    return didPrompt ? undefined : "skipped";            // :1610
```

(There are two further early `return "skipped"` gates at `:1561`/`:1566` for a job that is no longer runnable, and one at `:1587` on the non-heartbeat branch.)

**3. What counts as busy** — `shouldDeferHeartbeatCronJob`, `packages/coding-agent/src/core/cron-jobs.ts:1351`:

```ts
const busyBesidesStreaming =
    activity.isCompacting === true ||
    activity.isRetrying === true ||
    activity.isBashRunning ||
    activity.hasPendingSessionWork ||
    (!activity.isStreaming && activity.unfinishedActionCount > 0);
if (busyBesidesStreaming) {
    return true;
}
```

Plus `activity.isStreaming` for `follow_up`-mode heartbeats. A session running a long bash command loses every heartbeat for the duration.

**4. Skip bookkeeping is a lone timestamp** — `recordDispatchResult`, `packages/coding-agent/src/core/cron-jobs.ts:715`:

```ts
if (result.outcome === "skipped" && result.error === undefined) {
    const nextRunAt = nextRunAtForSchedule(job.schedule, now);
    updated = {
        ...job,
        status: job.schedule.kind === "once" ? "completed" : job.status,
        nextRunAt: nextRunAt?.toISOString(),
        lastSkippedAt: now.toISOString(),
        updatedAt: now.toISOString(),
    };
    return updated;
}
updated = {
    ...job,
    ...
    runCount: job.runCount + 1,
```

Same shape in `recordSkipResult` (`:671`). `runCount` advances only on a real delivery (`:747`), so it is already the correct "beat number" — it is simply never shown to the model. `formatAgentCronJob` (`:1242`) renders `skipped=${new Date(job.lastSkippedAt).toLocaleString()}` at `:1248` — most-recent only, no count.

**5. The prompt carries nothing** — `createHeartbeatPromptMessage`, `packages/coding-agent/src/core/messages.ts:401`:

```ts
return {
    role: "custom",
    customType: HEARTBEAT_PROMPT_CUSTOM_TYPE,
    content: job.prompt,
    display: true,
    details: {
        jobId: job.id,
        schedule: job.schedule.expression,
        status: job.status,
        runCount: job.runCount,
        nextRunAt: job.nextRunAt,
        lastRunAt: job.lastRunAt,
    },
    timestamp,
};
```

**6. `details` never reaches the LLM** — `convertToLlm`, `packages/coding-agent/src/core/messages.ts:430`, `custom` branch:

```ts
const content = typeof m.content === "string" ? [{ type: "text" as const, text: m.content }] : m.content;
return {
    role: "user",
    content,
    timestamp: m.timestamp,
};
```

Only `m.content` survives. Everything in `HeartbeatPromptDetails` (`packages/coding-agent/src/core/messages.ts:157`) is dropped before the model sees the turn.

**7. Delivery path** — `AgentSession.promptHeartbeat`, `packages/coding-agent/src/core/agent-session.ts:4492`:

```ts
const message = createHeartbeatPromptMessage(job);
await this._promptInjectedMessage(job.prompt, message, {
```

Note the prompt text is passed *separately* from the custom message, so a fix must update both arguments, not just the message.

## Suggested fix

**(a) Count the skips.** Add `skippedSinceLastRun: number` to `AgentCronJob` (`packages/coding-agent/src/core/cron-jobs.ts:34`ff, alongside `lastSkippedAt` at `:52`). Increment it wherever `lastSkippedAt` is written — the `outcome === "skipped"` branch of `recordDispatchResult`, `recordSkipResult`, and the already-claimed branch of `claimDueInState` — and reset it to `0` in the same object literal that does `runCount: job.runCount + 1`. Extend `formatAgentCronJob` to `skipped=<date> (<n> since last delivery)`, and add a `Skipped:` line to `showHeartbeat` so `/heartbeat status` shows it too.

**(b) Put it in `content`, not `details`.** In `createHeartbeatPromptMessage`, append a machine-readable block to the prompt text, and have `promptHeartbeat` pass that same augmented string to `_promptInjectedMessage` instead of raw `job.prompt`:

```
[heartbeat · job <id> · run #<runCount + 1> · schedule <schedule.expression> · previous delivery <lastRunAt>]
[WARNING: <skippedSinceLastRun> scheduled fire(s) were SKIPPED since the last delivery (most recent <lastSkippedAt>) — this heartbeat is not a reliable clock]
```

Emit the second line only when `skippedSinceLastRun > 0`. `runCount + 1` should be correct at message-construction time because `recordDispatchResult` increments `runCount` only after `runJob` resolves. Keeping this in `content` is the load-bearing part — per pointer 6, anything in `details` cannot influence the model.

**(c) Optional, larger.** Consider deferring instead of dropping: do not advance `nextRunAt` in `claimDueInState` until the dispatch outcome is known, or re-arm on skip so at most one beat is coalesced forward rather than lost. (b) alone makes the loss legible; (c) makes it rare.

## Impact

A heartbeat is the only self-clock a long-running or unattended agent has, and `rlm_heartbeat` subagents depend on it for the same reason. The busy conditions in `shouldDeferHeartbeatCronJob` — long bash runs, compaction, retries, pending session work — are exactly the states in which a periodic check-in matters most, so skips cluster over precisely the intervals a check-in is meant to cover. Because the delivered prompt is invariant, the agent has no way to detect the gap: it reads each beat as the next tick of an unbroken cadence and builds plans, deadlines, and "since last check" reasoning on a schedule that did not happen. Fix (b) is a few lines and removes the silent-failure property even without (c).

Same failure shape as #799 (a silently truncated result that reads as the complete set), and adjacent to #758/#759 on the general theme of scheduling/completion state that exists internally but never reaches the agent.

Hotragn:
Fixes #820.

A heartbeat fire that lands on a busy session is currently consumed and lost, and the beats that do land carry no way to notice the gap.

## The fire is consumed before the busy check

`claimDueInState` advances `job.nextRunAt` at claim time, before `runCronJob` ever asks `shouldDeferHeartbeatCronJob`. When the session turns out to be busy, `recordDispatchResult`'s skip branch advances `nextRunAt` a *second* time, from `now`. Three consequences:

- The missed beat is never re-armed.
- An interval schedule loses its phase on every skip: `every 5m` skipped 90s late becomes `every 5m` measured from the skip, so the cadence drifts by the length of every busy check.
- A busy window spanning several beats is indistinguishable from one skipped beat. `lastSkippedAt` is a single overwritten timestamp.

`rescheduleAfterSkip` now advances from the dispatch record's `scheduledFor` — the beat that was actually skipped — and steps the schedule forward until the next fire is strictly after `now`. So:

- `every 5m` due at 12:05 and declined at 12:06:30 re-arms at **12:10**, not 12:11:30.
- The same job declined at 12:27 re-arms at 12:30 and reports **5** swallowed fires (12:05, 12:10, 12:15, 12:20, 12:25).
- Because the result is always strictly after `now`, a busy session cannot make the scheduler spin on an already-due job. Dispatch attempts stay at one per interval, exactly as today.
- The step loop is bounded (`MAX_COALESCED_SKIPS`); a pathologically short schedule against a very long busy window falls back to the plain advance while keeping the count it accumulated.

Those fires land in a new `AgentCronJob.missedRunCount`, which accumulates across skips and is cleared when a beat is delivered. All three skip paths feed it: the `recordDispatchResult` skip branch, `recordSkipResult`, and `claimDueInState`'s already-claimed branch (a fire arriving while the previous dispatch is still outstanding is swallowed the same way).

I deliberately did not make a declined beat re-fire immediately. Re-arming at the original `scheduledFor` would leave the job due while the session is still busy, and `runCronJob` declines cheaply, so it would busy-loop for the length of the busy window. Coalescing forward preserves both the user's cadence and the scheduler's load profile.

## The delivered prompt could not report any of this

`createHeartbeatPromptMessage` put `runCount`/`lastRunAt`/`nextRunAt` in `HeartbeatPromptDetails`, which `convertToLlm` drops, and set `content: job.prompt`. Every delivery was byte-identical, so a heartbeat-driven agent could not tell "beat 3 of 3" from "beat 3 of 40".

The heartbeat message content is what `convertToLlm` turns into the provider's user message, so `formatHeartbeatPromptContent` now prefixes the instruction with a compact block:

```text
<heartbeat>
beat 7
schedule every 5m
previous delivery 2026-01-01T12:05:00.000Z
2 scheduled fires skipped while this session was busy (most recent 2026-01-01T12:15:00.000Z)
</heartbeat>

Check the deployment and report meaningful changes
```

The instruction stays last so it remains the final thing the model reads, and the skipped-fires line appears only when there is a backlog. Queue previews and `clearQueue()` are unaffected — they read the prompt text passed to `_promptInjectedMessage`, which is still the bare instruction, so `Heartbeat prompt: <instruction>` renders exactly as before.

## Human-visible surfaces

`/heartbeat status` gained `Missed: <n> since last run, most recent <date>`; previously it showed nothing about skips at all. `formatAgentCronJob` gained `missed=<n>` alongside the existing `skipped=<date>`.

## Daemon protocol

`DaemonCronJob` is an alias of `AgentCronJob`, so `missedRunCount` is published on cron/heartbeat job rows. That is an additive change on an existing response shape — a new daemon's extra field is ignored by an old client, and a new client renders nothing when an old daemon omits it — matching revisions 9/10/12, which also published new fields. `DAEMON_SCHEMA_REVISION` moves to 16 with a matching comment and `DAEMON_SCHEMA_ID` (14 was taken by the telemetry opt-out in #521 and 15 by the queued-message mutation command in #838, both of which landed while this was open; rebased onto each in turn). The schema digest is unchanged from `main` because it hashes the `daemon-protocol.ts` wire-type source, which this PR does not touch. The schema digest hashes the `daemon-protocol.ts` wire-type source only, so it is unchanged; `test/daemon-protocol.test.ts` verifies the identity stays synchronized.

## Tests

New: `packages/coding-agent/test/suite/regressions/820-heartbeat-missed-beats.test.ts` (6 cases) — phase preservation on a declined beat, coalescing five fires across a 22-minute busy window with the re-armed fire proven to be in the future, accumulation across two skips followed by a delivery that clears the backlog, `missed=` in the daemon listing, the composed prompt text in first-delivery/backlog/singular forms, and an end-to-end case asserting the model actually receives the backlog notice.

Updated (intentional behavior changes, not relaxations):

- `test/cron-jobs.test.ts` — two tests asserted the drifting reschedule. One is renamed from "reschedules a skipped dispatch from the skip time" to "...on the original cadence"; both now also assert `missedRunCount`.
- `test/suite/regressions/4482-heartbeat-injected-prompt.test.ts` — four assertions compared the delivered text to the bare instruction; they now compare against `formatHeartbeatPromptContent`. The queue-preview and `clearQueue` assertions were left untouched and still pass, which is the check that previews stayed clean.

Verified: `npm run check` clean; `cron-jobs`, `daemon-protocol`, `daemon-client`, `agent-connection-daemon`, `messages`, `4482`, and the new suite pass (180 + 6). `daemon-mode.test.ts` has 5 failures on this Windows machine (unix-socket bind and symlink cases) that reproduce identically on a clean `main` worktree.

> [!NOTE]
> ### Fix heartbeat cron jobs to preserve schedule cadence and track missed fires on busy session declines
> - When a heartbeat fire is declined (session busy), the scheduler now re-arms from the original scheduled beat rather than the skip time, preserving the schedule phase.
> - Missed fires during a busy window are coalesced into a `missedRunCount` on the job, reset to zero on the next successful delivery.
> - Delivered heartbeat prompts include a `<heartbeat>` block with beat number, schedule expression, previous delivery time, and a missed-fire summary when backlog exists.
> - `missed=<n>` appears in daemon job listings and the interactive `/heartbeat status` view when there is a backlog.
> - Behavioral Change: `recordSkipResult` and `recordDispatchResult` in [`cron-jobs.ts`](https://github.com/PrimeIntellect-ai/prime-agent/pull/890/files#diff-fb783dad61580c3b11d0731df2a325b6e3dd852fc430a309a5f3ae7ebcc826dd) now advance `nextRunAt` along the original cadence instead of from the skip time; daemon schema bumped to revision 17.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized d101619.</sup>
> 
>

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -20,6 +20,9 @@
 - Fixed workers with no live connection reporting as `ready`; stopping workers now report a `stopping` state, are hidden from live sessions, and no longer receive daemon-wide commands ([#850](https://github.com/PrimeIntellect-ai/prime-agent/pull/850)).
 - Fixed timed-out worker stops stranding dead-but-registered workers ("Session worker is not connected"); stops now finalize in the background once the process exits, and zombie processes are no longer counted as alive ([#851](https://github.com/PrimeIntellect-ai/prime-agent/pull/851)).
 - Fixed sessions becoming permanently unopenable after a stale worker registration was left behind; open/resume now self-heals by finishing the old cleanup and starting a fresh worker ([#852](https://github.com/PrimeIntellect-ai/prime-agent/pull/852)).
+- Fixed a heartbeat fire declined by a busy session restarting the schedule from the skip time instead of staying on its original cadence ([#820](https://github.com/PrimeIntellect-ai/prime-agent/issues/820))
+- Added a coalesced count of heartbeat fires skipped since the last delivery, shown in `/heartbeat status` and the daemon job listing ([#820](https://github.com/PrimeIntellect-ai/prime-agent/issues/820))
+- Changed delivered heartbeat prompts to carry the beat number, previous delivery time, and any skipped-fire backlog ([#820](https://github.com/PrimeIntellect-ai/prime-agent/issues/820))
 
 ## [0.7.1] - 2026-08-07
 
--- a/packages/coding-agent/docs/long-running-agents.md
+++ b/packages/coding-agent/docs/long-running-agents.md
@@ -156,6 +156,25 @@ await rlm_heartbeat.update(first["heartbeat"]["id"], status="pause")
 
 RLM heartbeats are distinct from the user's `/heartbeat`; the Python skill cannot replace or clear the user-owned heartbeat.
 
+### Fires that land on a busy session
+
+A fire is declined when the session is busy — streaming, compacting, retrying, running bash, or holding unfinished actions. The declined fire is re-armed on the schedule's original phase rather than restarting the interval from the moment it was declined, and every fire that comes and goes during one busy window is coalesced into a single backlog count.
+
+The next delivered beat carries that context. Each heartbeat prompt is prefixed with a `<heartbeat>` block naming the beat number, the schedule, the previous delivery time, and — when there is a backlog — how many fires were skipped and when the most recent one was:
+
+```text
+<heartbeat>
+beat 7
+schedule every 5m
+previous delivery 2026-01-01T12:05:00.000Z
+2 scheduled fires skipped while this session was busy (most recent 2026-01-01T12:15:00.000Z)
+</heartbeat>
+
+Check the deployment and report meaningful changes
+```
+
+The backlog resets once a beat is delivered. `/heartbeat status` shows it as `Missed: <n> since last run`, and the daemon job listing as `missed=<n>`.
+
 ### General schedules
 
 Schedule a one-time or recurring prompt for an addressable agent:
--- a/packages/coding-agent/src/core/cron-jobs.ts
+++ b/packages/coding-agent/src/core/cron-jobs.ts
@@ -50,6 +50,11 @@ export interface AgentCronJob {
 	nextRunAt?: string;
 	lastRunAt?: string;
 	lastSkippedAt?: string;
+	/**
+	 * Scheduled fires swallowed since the last delivered run, including fires
+	 * coalesced away while the session stayed busy. Reset when a run is delivered.
+	 */
+	missedRunCount?: number;
 	lastError?: string;
 	runCount: number;
 }
@@ -111,6 +116,8 @@ interface CronJobsState {
 export const SESSION_SCHEDULED_JOBS_FILENAME = "scheduled-jobs.json";
 
 const MAX_TIMEOUT_MS = 2_147_483_647;
+/** Upper bound on schedule steps walked while coalescing skipped fires. */
+const MAX_COALESCED_SKIPS = 10_000;
 const ONE_SECOND_MS = 1000;
 const ONE_MINUTE_MS = 60_000;
 export const DEFAULT_HEARTBEAT_SCHEDULE = "every 5m";
@@ -651,7 +658,7 @@ export class AgentCronJobStore {
 					: job.schedule.kind === "interval"
 						? nextRunAtForSchedule(job.schedule, now)
 						: undefined;
-			updated = {
+			updated = withoutMissedRunCount({
 				...job,
 				status: job.schedule.kind === "once" ? "completed" : "active",
 				nextRunAt: nextRunAt?.toISOString(),
@@ -659,7 +666,7 @@ export class AgentCronJobStore {
 				lastError,
 				runCount: job.runCount + 1,
 				updatedAt: now.toISOString(),
-			};
+			});
 			return updated;
 		});
 		if (updated) {
@@ -668,7 +675,7 @@ export class AgentCronJobStore {
 		return updated;
 	}
 
-	recordSkipResult(id: string, result: { now?: Date }): AgentCronJob | undefined {
+	recordSkipResult(id: string, result: { now?: Date; scheduledFor?: string }): AgentCronJob | undefined {
 		const now = result.now ?? new Date();
 		let updated: AgentCronJob | undefined;
 		const jobs = this.readJobs().map((job) => {
@@ -679,13 +686,8 @@ export class AgentCronJobStore {
 				updated = job;
 				return job;
 			}
-			const nextRunAt = nextRunAtForSchedule(job.schedule, now);
-			updated = {
-				...job,
-				nextRunAt: nextRunAt?.toISOString(),
-				lastSkippedAt: now.toISOString(),
-				updatedAt: now.toISOString(),
-			};
+			const { nextRunAt, missed } = rescheduleAfterSkip(job, result.scheduledFor ?? job.nextRunAt, now);
+			updated = withSkipAccounting({ ...job, nextRunAt: nextRunAt?.toISOString() }, now, missed);
 			return updated;
 		});
 		if (updated) {
@@ -729,24 +731,26 @@ export class AgentCronJobStore {
 					return job;
 				}
 				if (result.outcome === "skipped" && result.error === undefined) {
-					const nextRunAt = nextRunAtForSchedule(job.schedule, now);
-					updated = {
-						...job,
-						status: job.schedule.kind === "once" ? "completed" : job.status,
-						nextRunAt: nextRunAt?.toISOString(),
-						lastSkippedAt: now.toISOString(),
-						updatedAt: now.toISOString(),
-					};
+					const { nextRunAt, missed } = rescheduleAfterSkip(job, dispatch.scheduledFor, now);
+					updated = withSkipAccounting(
+						{
+							...job,
+							status: job.schedule.kind === "once" ? "completed" : job.status,
+							nextRunAt: nextRunAt?.toISOString(),
+						},
+						now,
+						missed,
+					);
 					return updated;
 				}
-				updated = {
+				updated = withoutMissedRunCount({
 					...job,
 					status: job.schedule.kind === "once" ? "completed" : job.status,
 					lastRunAt: now.toISOString(),
 					lastError: result.error === undefined ? undefined : errorMessage(result.error),
 					runCount: job.runCount + 1,
 					updatedAt: now.toISOString(),
-				};
+				});
 				return updated;
 			});
 			return [];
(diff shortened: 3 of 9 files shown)


kevinjosethomas:

PR #891, opened 2026-08-07
feat(coding-agent): report in-flight tool-call progress in agent_observe

### LINKED ISSUE #822: [coding-agent] agent_observe reports that a child is in a tool call but never for how long — no elapsed time, no in-flight tool-call count
## Summary

`agent_observe` tells an orchestrator *that* a child session is executing a tool call, but never *how long* it has been executing it. `AgentObserveAgentSummary` collapses the child's entire set of in-flight tool calls into a single `status: "tool"` string — no start timestamp, no elapsed duration, and no count.

A child three seconds into a fast `bash` call and a child forty minutes into a hung network request or a `bash` call waiting on a subprocess that never exits produce byte-identical summaries. The second case does not require anything unusual: the `bash` tool's `timeout` is optional with no default (`packages/coding-agent/src/core/tools/bash.ts:26` — `"Timeout in seconds (optional, no default timeout)"`), so an indefinitely blocking call is ordinary behaviour, not an error path.

To be precise about what *is* already there, since it bounds the ask: a wedged child does **not** look idle. `status` is correctly `"tool"`, and `latestMessage.toolCalls` names the tools the last assistant message requested. The missing piece is strictly the time dimension, plus the count.

## Environment

- prime-agent 0.7.0, installed from npm
- Component: `packages/coding-agent` — daemon mode / the `agent-observe` skill. The underlying state lives in `packages/agent`, which ships as `@earendil-works/pi-agent-core`.
- Line numbers below are against the `main` branch head as of 2026-08-06; every pointer also names the enclosing symbol, so it should survive line drift. The shipped 0.7.0 build matches that source for every snippet quoted here — checked against `dist/modes/daemon/daemon-mode.js`, `dist/core/agent-observe.d.ts`, and the bundled `@earendil-works/pi-agent-core` `dist/agent.js` in the installed package.

## Reproduction

1. From a parent session, spawn a child that will block inside a single tool call:
   ```python
   handle = await rlm("Run the shell command: sleep 100000", name="worker")
   ```
2. Wait a few seconds for the child to enter the tool call, then observe it:
   ```python
   snap = await agent_observe.get_agent(handle.name)
   print(snap)
   ```
3. Wait an arbitrarily long time — 60 seconds, then 30 minutes — and repeat step 2.
4. Compare the two summaries field by field.

`sleep` is just the minimal deterministic case; any tool call that blocks indefinitely reproduces this identically.

## Expected

The summary should carry enough information for the parent to compute how long the current tool call has been running, so that a policy such as "if the child has been in a single tool call for more than N minutes, treat it as wedged" can be written against the API. Something along the lines of an in-flight tool-call count plus a start timestamp or elapsed duration.

## Actual

Every poll after the tool call starts returns the same summary, regardless of elapsed time. Per the object literal built at `daemon-mode.ts:2959-2983`, the fields relevant to progress are:

```
status:          "tool"
isStreaming:     true
isCompacting:    false
isSessionActive: true
messageCount:    <unchanged while inside the tool call>
queuedCount:     0            # session-action queue, not tool calls — see below
latestMessage:   { role: "assistant", timestamp: <fixed>, toolCalls: ["bash"], ... }
```

`messageCount` cannot advance while the call is in flight, because the tool-result message is only appended once the tool returns. Nothing in the summary is a function of elapsed time.

The only workaround available today is to poll `agent_observe.recent_messages()`, read the latest assistant message's `timestamp` — which is optional, emitted only when the message carries one (`agent-observe.ts:111`) — and diff it against wall-clock by hand on every poll, reconstructing a duration the daemon already has in hand.

## Code pointers

**`packages/coding-agent/src/core/agent-observe.ts:7-27` — `AgentObserveAgentSummary`.** The full set of fields; there is no tool-call count and no timing field:

```ts
export interface AgentObserveAgentSummary {
	activeSessionId: string;
	sessionId: string;
	sessionName?: string;
	runtimeKind?: "top-level" | "subagent";
	cwd: string;
	status: string;
	isCurrent: boolean;
	isStreaming: boolean;
	isCompacting: boolean;
	attachedClients: number;
	messageCount: number;
	queuedCount: number;
	isSessionActive: boolean;
	...
	latestMessage?: AgentObserveMessagePreview;
}
```

Note that `queuedCount` is the queued *session action* count (`summary.sessionActions.queuedCount`, `daemon-mode.ts:2971`), unrelated to in-flight tool calls. `packages/coding-agent/skills/agent-observe/SKILL.md:31` describes the summary as including a "pending count", which is easy to read as tool-call pendency; it is not.

**`packages/coding-agent/src/modes/daemon/daemon-mode.ts:2940-2984` — `createAgentObserveSummary`.** The set of in-flight tool calls is reduced to a boolean test and discarded:

```ts
const status = session.isStreaming
	? session.state.pendingToolCalls.size > 0
		? "tool"
		: "model"
	: session.isCompacting
		? "compacting"
		: session.isSessionActive || session.hasRunningRlmChildren()
			? "busy"
			: state.clients.size > 0
				? "user"
				: "idle";
```

`session.state.pendingToolCalls` is right there and `.size` is already being read — only the `> 0` comparison survives into the response. (`daemon-session-list.ts:235` does the same thing for `isRunningTools`.)

**`packages/agent/src/types.ts:342` — `AgentState.pendingToolCalls`.** The underlying state is IDs only, with no timestamps:

```ts
/** Tool call ids currently executing. */
readonly pendingToolCalls: ReadonlySet<string>;
```

**`packages/agent/src/agent.ts:580-585` — the `tool_execution_start` reducer.** The event carries a `toolName` (`types.ts:419`: `{ type: "tool_execution_start"; toolCallId: string; toolName: string; args: any }`, emitted with the real name at `agent-loop.ts:647` and `:702`), but the reducer keeps only the id and records no start time:

```ts
case "tool_execution_start": {
	const pendingToolCalls = new Set(this._state.pendingToolCalls);
	pendingToolCalls.add(event.toolCallId);
	this._state.pendingToolCalls = pendingToolCalls;
	break;
}
```

So the start time is not merely unexposed — it is never captured anywhere.

## Why the parent has no other signal

`agent_message.send` always uses steering delivery, and its receipt can come back `"queued"`, meaning the message "will deliver when the target's current work allows" (`packages/coding-agent/skills/agent-message/SKILL.md`). Reading `packages/agent/src/agent-loop.ts`, the steering queue appears to be drained only at turn boundaries: once before the first turn (`:318`) and once after each turn completes (`:407`). If that reading is right, a child blocked inside a single tool call never reaches the post-turn poll, so a queued message stays undelivered for as long as the call blocks — and the parent cannot distinguish "child is working" from "child is wedged and unreachable" from any field the API returns.

Is that the intended delivery boundary? If so, an elapsed-time field on the summary is the only thing that would let a parent detect the condition.

## Suggested fix

1. In `packages/agent`, retain the start time (and the already-available tool name) when a call begins. Either add a parallel `pendingToolCallStartedAt: ReadonlyMap<string, number>` to `AgentState`, or widen `pendingToolCalls` to `ReadonlyMap<string, { toolName: string; startedAt: number }>`. The map form is strictly more useful; the parallel-map form is additive and avoids touching existing `.size` and iteration call sites.

2. Add to `AgentObserveAgentSummary` in `agent-observe.ts`:
   ```ts
   pendingToolCallCount: number;
   currentToolCallStartedAt?: number;   // epoch ms, oldest in-flight call
   currentToolCallElapsedMs?: number;   // Date.now() - currentToolCallStartedAt
   currentToolCallNames?: string[];     // optional; toolName is already on the event
   ```
   All additive, so existing consumers are unaffected.

3. Populate them in `createAgentObserveSummary`, taking the oldest in-flight call as "current" so that parallel tool calls still surface the longest-running one.

4. Document the new fields in `packages/coding-agent/skills/agent-observe/SKILL.md`, with an explicit note that `status: "tool"` alone is not a liveness signal and that the elapsed-time field is the one to threshold on. Clarifying the existing "pending count" wording against `queuedCount` at the same time would help.

5. Minor and adjacent, but it bites when `latestMessage` is used as a fallback progress signal: the summary's `latestMessage` preview is always clipped to a hardcoded `240` (`daemon-mode.ts:2980`). That includes the `agent` summary embedded in the `agent_observe.recent_messages` response, whose sibling `messages[]` entries *do* honour the caller's `max_chars` (`daemon-mode.ts:2933`). `list_agents` and `get_agent` accept no `max_chars` at all. If the asymmetry is deliberate, disregard; otherwise hoisting `240` into a named exported constant, or accepting `max_chars` on `get_agent`, would help.

If a change to `packages/agent` is undesirable, a narrower version can be derived entirely inside `createAgentObserveSummary` by walking back from the end of `session.messages` to the last assistant message and using its `timestamp` as the tool-call start. That is approximate — it measures time since the assistant message that requested the call, not since execution began — and it depends on that message carrying a timestamp, but it is enough to threshold on and requires no core state change.

## Impact

This is the load-bearing signal for any long-running multi-agent setup. An orchestrator supervising several children has no bounded way to answer "is this child working or stuck?" A wedged child holds its slot indefinitely, cannot be reached by `agent_message` (see above), and the parent cannot tell that it should escalate to a lifecycle operation such as `rlm.delete_subagent()`. The failure is silent and unbounded: the parent waits on a child that will never report back, and the only recovery is a human noticing.

The information needed is already present in the daemon process; it is discarded one line before it would be serialized.

Related open issues, all instances of "the parent cannot determine child state from the API": #758 (`repliedSinceTask` flag unreliable for determining subagent completion) and #759 (subagents can complete and write files without sending agent messages) cover the *completion* half of this problem; this issue covers the *liveness* half. #762 (`agent_message` and `agent_observe` skill API surface not discoverable) overlaps with the SKILL.md portion of the fix.

Hotragn:
Fixes #822.

`agent_observe` told an orchestrator *that* a child was executing a tool call, never *how long*. `AgentObserveAgentSummary` collapsed the whole in-flight set into `status: "tool"` — no count, no start time, no duration — so a child three seconds into a fast `bash` call and a child forty minutes into one blocked on a subprocess that never exits produced byte-identical summaries. Since `bash`'s `timeout` is optional with no default, an indefinitely blocking call is ordinary behaviour rather than an error path, and there was no way to write "treat a child stuck in one tool call for more than N minutes as wedged" against the API.

## The time dimension did not exist

`AgentState` tracked `pendingToolCalls` as a bare `Set<string>` of ids; nothing anywhere recorded when a call started, so this could not be fixed in the coding-agent alone.

`packages/agent` now keeps `AgentState.pendingToolCallStartedAt: ReadonlyMap<string, number>` alongside it — populated in the `tool_execution_start` reducer, removed in `tool_execution_end`, and cleared by `reset()` and `finishRun()` exactly where `pendingToolCalls` is. It is replaced rather than mutated in place, matching how the existing set is updated, so snapshot readers keep seeing a stable value. It joins the `Omit` lists that keep runtime-owned state out of `initialState`, so it cannot be seeded from the outside.

## What summaries now carry

Every `AgentObserveAgentSummary` gains:

- `pendingToolCallCount` — tool calls executing right now. Required, so a caller can always branch on it. A passive (non-resident) subagent has no live agent and reports `0`.
- `oldestPendingToolCallStartedAt` — epoch ms of the longest-running in-flight call. Absent when nothing is in flight.
- `pendingToolCallElapsedMs` — how long that call has been running, measured when the summary was built.

Both the timestamp and the duration are reported deliberately: elapsed is what a policy reads directly, while the timestamp lets a caller recompute against its own clock or diff across two polls without trusting the daemon's wall clock. `summarizeInFlightToolCalls` is a small exported pure function so all three fields have one definition and one test surface, and it clamps elapsed at `0` rather than reporting a negative duration if the clock moves backwards.

The Python `agent_observe` skill passes host payloads through unchanged, so the fields are available from the kernel with no skill change. `SKILL.md` documents them with the policy this issue asks for:

```python
snapshot = await agent_observe.get_agent(handle.name)
agent = snapshot["agent"]
stuck_for = agent.get("pendingToolCallElapsedMs") or 0
if agent["status"] == "tool" and stuck_for > 10 * 60 * 1000:
    await agent_message.send(
        "You have been in one tool call for over ten minutes. Report what it is waiting on.",
        receiver_role="child",
        receiver_name=handle.name,
    )
```

## Scope

`status` was already correct for a wedged child — it reports `"tool"`, and `latestMessage.toolCalls` names the tools — so nothing about status classification changed. This adds only the missing time and count.

No daemon schema revision: `AgentObserveAgentSummary` is carried inside `agent_observe.*` host-bridge results, not the daemon command/event wire types the schema id covers.

## Tests

`packages/coding-agent/test/suite/regressions/822-agent-observe-tool-elapsed.test.ts` (4 cases): the empty case reports a zero count and no timestamps; a three-call map picks the *oldest* start rather than the most recent and derives elapsed from it; a backwards clock yields `0` rather than a negative; and a live-session case runs two tools that block until released, asserting the tracked ids match `pendingToolCalls` exactly, that the recorded starts are not earlier than the moment the turn began, that both calls are counted concurrently, and that the map is empty again once the run finishes.

Three existing controller fixtures construct summaries by hand and were updated for the now-required count.

Verified: `npm run check` clean; `packages/agent` full suite 69/69; `agent-session-observe`, `agent-session-services`, and the new suite 9/9.

> [!NOTE]
> ### Report in-flight tool-call progress in `agent_observe` summaries
> - Adds `pendingToolCallStartedAt` (a `ReadonlyMap<string, number>`) to `AgentState` in [types.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/891/files#diff-a882de319ec56787d70c4545d2d565440cb96de1b96bac234f4605144969c825), recording epoch-ms start times for each active tool call alongside the existing `pendingToolCalls` set.
> - Introduces `summarizeInFlightToolCalls` in [agent-observe.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/891/files#diff-cbcb012ae0b9420e72441c880504a42316b1c46d476747795b63c9d4aa7574c1) which computes `pendingToolCallCount`, `oldestPendingToolCallStartedAt`, and `pendingToolCallElapsedMs` from the map.
> - Spreads those fields into live agent summaries returned by `agent_observe` list/summary calls; passive subagents report `pendingToolCallCount: 0`.
> - Adds a regression test suite in [822-agent-observe-tool-elapsed.test.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/891/files#diff-15e5eec0fa7fd10cea1dcd0320273f30793e0983f2977da43a52bc7fc71c8182) covering empty maps, clock skew, and integration with session state.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 8675661.</sup>
> 
>

### DIFF
--- a/packages/agent/CHANGELOG.md
+++ b/packages/agent/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [Unreleased]
 
 ## [0.7.2] - 2026-08-11
+- Added `AgentState.pendingToolCallStartedAt`, recording when each in-flight tool call started so observers can measure how long one has been running.
 
 ## [0.7.1] - 2026-08-07
 
--- a/packages/agent/src/agent.ts
+++ b/packages/agent/src/agent.ts
@@ -57,15 +57,24 @@ const DEFAULT_MODEL = {
 
 type QueueMode = "all" | "one-at-a-time";
 
-type MutableAgentState = Omit<AgentState, "isStreaming" | "streamingMessage" | "pendingToolCalls" | "errorMessage"> & {
+type MutableAgentState = Omit<
+	AgentState,
+	"isStreaming" | "streamingMessage" | "pendingToolCalls" | "pendingToolCallStartedAt" | "errorMessage"
+> & {
 	isStreaming: boolean;
 	streamingMessage?: AgentMessage;
 	pendingToolCalls: Set<string>;
+	pendingToolCallStartedAt: Map<string, number>;
 	errorMessage?: string;
 };
 
 function createMutableAgentState(
-	initialState?: Partial<Omit<AgentState, "pendingToolCalls" | "isStreaming" | "streamingMessage" | "errorMessage">>,
+	initialState?: Partial<
+		Omit<
+			AgentState,
+			"pendingToolCalls" | "pendingToolCallStartedAt" | "isStreaming" | "streamingMessage" | "errorMessage"
+		>
+	>,
 ): MutableAgentState {
 	let tools = initialState?.tools?.slice() ?? [];
 	let messages = initialState?.messages?.slice() ?? [];
@@ -90,13 +99,19 @@ function createMutableAgentState(
 		isStreaming: false,
 		streamingMessage: undefined,
 		pendingToolCalls: new Set<string>(),
+		pendingToolCallStartedAt: new Map<string, number>(),
 		errorMessage: undefined,
 	};
 }
 
 /** Options for constructing an {@link Agent}. */
 export interface AgentOptions {
-	initialState?: Partial<Omit<AgentState, "pendingToolCalls" | "isStreaming" | "streamingMessage" | "errorMessage">>;
+	initialState?: Partial<
+		Omit<
+			AgentState,
+			"pendingToolCalls" | "pendingToolCallStartedAt" | "isStreaming" | "streamingMessage" | "errorMessage"
+		>
+	>;
 	convertToLlm?: (messages: AgentMessage[]) => Message[] | Promise<Message[]>;
 	transformContext?: (messages: AgentMessage[], signal?: AbortSignal) => Promise<AgentMessage[]>;
 	streamFn?: StreamFn;
@@ -342,6 +357,7 @@ export class Agent {
 		this._state.isStreaming = false;
 		this._state.streamingMessage = undefined;
 		this._state.pendingToolCalls = new Set<string>();
+		this._state.pendingToolCallStartedAt = new Map<string, number>();
 		this._state.errorMessage = undefined;
 		this.clearFollowUpQueue();
 		this.clearSteeringQueue();
@@ -551,6 +567,7 @@ export class Agent {
 		this._state.isStreaming = false;
 		this._state.streamingMessage = undefined;
 		this._state.pendingToolCalls = new Set<string>();
+		this._state.pendingToolCallStartedAt = new Map<string, number>();
 		this.activeRun?.resolve();
 		this.activeRun = undefined;
 	}
@@ -581,6 +598,10 @@ export class Agent {
 				const pendingToolCalls = new Set(this._state.pendingToolCalls);
 				pendingToolCalls.add(event.toolCallId);
 				this._state.pendingToolCalls = pendingToolCalls;
+				// Start times let an observer distinguish a fast call from a wedged one.
+				const startedAt = new Map(this._state.pendingToolCallStartedAt);
+				startedAt.set(event.toolCallId, Date.now());
+				this._state.pendingToolCallStartedAt = startedAt;
 				break;
 			}
 
@@ -588,6 +609,9 @@ export class Agent {
 				const pendingToolCalls = new Set(this._state.pendingToolCalls);
 				pendingToolCalls.delete(event.toolCallId);
 				this._state.pendingToolCalls = pendingToolCalls;
+				const startedAt = new Map(this._state.pendingToolCallStartedAt);
+				startedAt.delete(event.toolCallId);
+				this._state.pendingToolCallStartedAt = startedAt;
 				break;
 			}
 
--- a/packages/agent/src/types.ts
+++ b/packages/agent/src/types.ts
@@ -340,6 +340,12 @@ export interface AgentState {
 	readonly streamingMessage?: AgentMessage;
 	/** Tool call ids currently executing. */
 	readonly pendingToolCalls: ReadonlySet<string>;
+	/**
+	 * Epoch milliseconds at which each in-flight tool call started executing.
+	 * Keyed by the same ids as {@link pendingToolCalls}, so an observer can tell a
+	 * call that started three seconds ago from one that has been blocked for an hour.
+	 */
+	readonly pendingToolCallStartedAt: ReadonlyMap<string, number>;
 	/** Error message from the most recent failed or aborted assistant turn, if any. */
 	readonly errorMessage?: string;
 }
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -20,6 +20,7 @@
 - Fixed workers with no live connection reporting as `ready`; stopping workers now report a `stopping` state, are hidden from live sessions, and no longer receive daemon-wide commands ([#850](https://github.com/PrimeIntellect-ai/prime-agent/pull/850)).
 - Fixed timed-out worker stops stranding dead-but-registered workers ("Session worker is not connected"); stops now finalize in the background once the process exits, and zombie processes are no longer counted as alive ([#851](https://github.com/PrimeIntellect-ai/prime-agent/pull/851)).
 - Fixed sessions becoming permanently unopenable after a stale worker registration was left behind; open/resume now self-heals by finishing the old cleanup and starting a fresh worker ([#852](https://github.com/PrimeIntellect-ai/prime-agent/pull/852)).
+- Added in-flight tool-call count and elapsed time to `agent_observe` summaries, so an orchestrator can tell a child in a fast tool call from one wedged in a blocking call ([#822](https://github.com/PrimeIntellect-ai/prime-agent/issues/822))
 
 ## [0.7.1] - 2026-08-07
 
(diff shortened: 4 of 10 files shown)


kevinjosethomas:

PR #892, opened 2026-08-07
fix(coding-agent): keep retried agent messages on one supervisor journal key

### LINKED ISSUE #821: `sendRemoteAgentSessionMessage` retries with a fresh `DaemonClient` per attempt, bypassing the supervisor command journal and duplicating steering prompts
## Summary

The supervisor already implements durable idempotency for mutating daemon commands: `CommandRecoveryJournal` keys completed results on `createCommandIdempotencyKey(clientId, commandId)` and replays the stored response instead of re-executing. `send_message` is a mutating command, so it is covered by this mechanism in principle.

The problem is that the one retry loop in the tree that re-sends `send_message` across the accept window — `DaemonMode.sendRemoteAgentSessionMessage` (the worker to supervisor hop) — constructs a **new `DaemonClient` on every attempt**. `protocolClientId` is a fresh `randomUUID()` per client instance and `requestId` restarts at 0, so each retry presents a **different journal key for the same logical message**. The journal cannot recognize the retry as a duplicate, `acceptAgentSessionMessage` performs no dedupe of its own, and the result is N independent steering prompts under N distinct `agentmsg_` ids.

Client-side, `DaemonClient.request` times out locally at 30s and sends no cancellation, so a timeout does not mean the supervisor did not accept and enqueue the message. That is exactly the window the journal was built to cover, and exactly the window this retry loop steps outside of.

## Environment

- prime-agent 0.7.0, installed from npm as `prime-agent` (in-repo source package `packages/coding-agent`, whose `package.json` name is `@earendil-works/pi-coding-agent`)
- Component: daemon mode, agent-messaging path (worker to supervisor `send_message` hop; also `daemon send` and any `DaemonClient` consumer of `send_message`)
- Verified against `main` at b9a4461149419156599d60174dddf15458e2b9ee

## What already works

`packages/coding-agent/src/modes/daemon/command-recovery-journal.ts` L44-46:

```ts
export function createCommandIdempotencyKey(clientId: DaemonClientId, commandId: DaemonCommandId): string {
	return JSON.stringify([clientId, commandId]);
}
```

`packages/coding-agent/src/modes/daemon/daemon-supervisor.ts` L1288-1298 short-circuits a duplicate of a completed command by replaying its recorded response:

```ts
const mutation = isDaemonMutatingCommand(command);
const journalIdentity =
    envelopeClientId && command.id && mutation ? { clientId: envelopeClientId, commandId: command.id } : undefined;
const existing = journalIdentity
    ? this.commandJournal.lookup(journalIdentity.clientId, journalIdentity.commandId)
    : undefined;
if (existing?.status === "complete") {
    if (parsedAdmission) this.deletePromptAdmission(parsedAdmission);
    this.write(client, existing.response);
    return;
}
```

`send_message` is not in `READ_ONLY_DAEMON_COMMANDS` (`daemon-protocol.ts` L1006-1039), so `isDaemonMutatingCommand` returns true and the journal path is active for it. The record is fsync'd before dispatch, so a supervisor crash leaves the entry `pending` and yields `command_result_uncertain` rather than a silent re-execution (L1299-1309). The in-instance reconnect replay in `DaemonClient` is correspondingly safe: L438-460 re-writes `pending.wireData` under the **same** request id, so the journal recognizes it.

## The gap

`packages/coding-agent/src/modes/daemon/daemon-client.ts` L113-114 — both halves of the journal key are per-instance:

```ts
private requestId = 0;
private readonly protocolClientId = `daemon-client:${randomUUID()}`;
```

`packages/coding-agent/src/modes/daemon/daemon-mode.ts` L5452-5488, `DaemonMode.sendRemoteAgentSessionMessage` — a new client per attempt, inside the retry loop:

```ts
const deadline = Date.now() + 30_000;
let lastError: unknown;
while (Date.now() < deadline && !this.shuttingDown) {
    const client = new DaemonClient(supervisorSocketPath);   // <-- new protocolClientId every pass
    let receivedResponse = false;
    try {
        await client.connect(1000);
        await client.waitForHello(1000);
        const response = await client.request({ type: "send_message", /* ... */ }, 30_000);
        receivedResponse = true;
```

```ts
    } catch (error) {
        lastError = error;
        if (receivedResponse) {
            throw error;
        }
    } finally {
        client.close();
    }
    await new Promise((resolveDelay) => setTimeout(resolveDelay, 250));
}
```

The `receivedResponse` guard correctly refuses to retry once a response was parsed. It does not cover the duplicate window: any transport-level failure *after* the supervisor accepted and enqueued but before the response was read — socket drop mid-response, supervisor restart, or `connect`/`waitForHello` succeeding and then `request` timing out. In that window the loop re-sends every 250ms until the 30s deadline, each pass under a fresh journal key.

Supporting facts:

1. **The client timeout is local-only and cancels nothing** — `daemon-client.ts` L293-295 (`request`, default `timeoutMs = 30000`) and L388-397:

```ts
private armPendingRequestTimeout(id: string, pending: PendingDaemonRequest): void {
    pending.timeout = setTimeout(() => {
        this.pendingRequests.delete(id);
        pending.reject(new Error(`Timed out after ${pending.timeoutMs}ms waiting for the Prime Agent daemon response to "${pending.commandType}". ...`));
    }, pending.timeoutMs);
}
```

It deletes the local pending entry and rejects. No cancellation is sent, and supervisor-side handling completes normally.

2. **The message id is minted server-side, per attempt** — `daemon-mode.ts` L5407-5416 in `sendAgentSessionMessage`:

```ts
const payload: AgentSessionMessagePayload = {
    id: createAgentSessionMessageId(),
    source: AGENT_MESSAGE_SOURCE,
    message,
```

`createAgentSessionMessageId` — `packages/coding-agent/src/core/agent-messages.ts` L330-332:

```ts
export function createAgentSessionMessageId(): string {
	return `agentmsg_${randomUUID()}`;
}
```

So there is no message-level identity a retry could reuse either; the only dedupe available is the journal's, and the retry loop moves out from under it.

3. **No dedupe on the receiving side** — `daemon-mode.ts` L5491-5526, `acceptAgentSessionMessage`. It checks queue capacity, computes `shouldQueue`, then queues:

```ts
if (shouldQueue) {
    const didQueue = await session.queueAgentMessagePrompt(prompt, streamingBehavior, message);
```

There is no lookup of `payload.id` against anything already queued or delivered.

4. **The rate limiter is charged per attempt** — same file, L5402 `const rateLimit = this.agentMessageRateLimiter.tryConsume(rateLimitKey);` and L5435 `this.agentMessageRateLimiter.refund(rateLimitKey);`. The refund only runs on the throw path, so an accepted-but-unacknowledged attempt keeps its budget and the retry consumes more.

5. **The wire command carries no caller-supplied key** — `daemon-protocol.ts` L459-468:

```ts
| {
        id?: string;
        type: "send_message";
        targetActiveSessionId: string;
        message: string;
        fromActiveSessionId?: string;
        /** Internal worker-origin marker; public clients remain unrestricted. */
        agentOrigin?: boolean;
        deliveryMode?: AgentSessionMessageDeliveryMode;
  }
```

`id?` is the request correlation id, overwritten per attempt by `requestWire` (`daemon-client.ts` L356-357). It doubles as half the journal key, which is why a stable client identity is what matters. An external retry wrapper around `daemon send` has no way to declare attempt #2 a retry of attempt #1, because each CLI invocation is a separate process with its own `protocolClientId`.

6. **The receipt gives the sender nothing to reconcile against** — `packages/coding-agent/skills/agent-message/SKILL.md` L45-47:

> `deliveryStatus` field: `"delivered"` means the message reached an idle target's context; `"queued"` means a steering message was accepted and will deliver when the target's current work allows (`send` does not block waiting for that).

This is deliberate and correct — `acceptAgentSessionMessage` L5522-5524 documents that awaiting delivery would deadlock mutual sends — but it means even a successful send yields no completion signal, so a sender cannot detect after the fact that it double-sent.

## Reproduction

The duplicate requires the transport to fail after acceptance, so the deterministic part below only establishes the missing caller-facing key; the duplicate itself is observed under load.

**Part A — no caller-facing way to mark a retry (deterministic).**

1. Start the daemon and two sessions; note the target's active session id (`prime-agent daemon list --json`).
2. Put the target into a long-running turn so subsequent sends queue rather than deliver.
3. Issue the same instruction twice, as an external retry wrapper would after a timeout:

   ```
   prime-agent daemon send --json <TARGET> --message "apply fix X"
   prime-agent daemon send --json <TARGET> --message "apply fix X"
   ```

4. Each receipt carries a different `id` (`agentmsg_<uuid-1>`, `agentmsg_<uuid-2>`), both `"deliveryStatus":"queued"`, and both are independent steering prompts.

   Two deliberate invocations *should* produce two messages, so this is not itself the defect — it establishes that the wire command exposes no key by which call #2 could declare itself a retry of call #1.

**Part B — the duplicate (requires the accept-then-fail window).**

5. Put sustained load on the supervisor so a `send_message` round trip can exceed the 30s client budget, and issue a send concurrently.
6. The client fails with:

   ```
   Timed out after 30000ms waiting for the Prime Agent daemon response to "send_message".
   ```

7. Inspect the target's transcript: the message from the "failed" attempt is present. The timeout was client-side only.
8. Let `sendRemoteAgentSessionMessage` retry, or retry manually. Measured on this path: 5 attempts, each reported to the caller as a timeout, produced 5 identical deliveries under 5 distinct `agentmsg_` ids.

## Expected

A send the supervisor has already accepted and enqueued is not re-enqueued by a retry — the same guarantee the command journal already provides for in-instance reconnect replay. N retries of one logical instruction result in exactly one delivery.

## Actual

The retry loop presents a fresh journal key per attempt, so the journal admits each retry as a new command and each becomes an additional steering prompt.

## Suggested fix

The mechanism already exists; the fix is to stop discarding the identity it keys on.

1. In `sendRemoteAgentSessionMessage` (`daemon-mode.ts` L5443), hoist the `DaemonClient` construction **out** of the `while` loop, or otherwise hold the client identity stable across attempts, so every retry of one logical message presents the same `(clientId, commandId)` and the journal replays the first result. This looks like the smallest correct change and does not alter the loop's timing.
   - If reconnecting per attempt is required, an alternative is to let a caller pin `protocolClientId` (and the request id) via a `DaemonClient` option so a retry can reconstruct the same key.
   - Worth checking whether `enableAutoReconnect` on a single long-lived client would let this loop be removed entirely, since the reconnect replay path is already journal-safe.
2. Handle `command_result_uncertain` explicitly in this loop. Once the key is stable, a retry racing an in-flight first attempt gets that code rather than a duplicate; today the loop has no branch for it.
3. Optionally, add an explicit `idempotencyKey?: string` to the `send_message` wire command so external retry wrappers around `daemon send` — separate processes that cannot share a `protocolClientId` — can also collapse retries. Strictly additive: absent a key, behavior is unchanged.
4. On a dedupe hit, ensure `agentMessageRateLimiter.tryConsume` is not charged again (fix 1 gets this for free, since the journal replays before dispatch).

## Impact

This affects long-running multi-agent setups, where worker-to-worker sends traverse the supervisor hop and the supervisor is loaded enough for a round trip to approach the 30s budget. The failure mode is worse than a dropped message: a duplicated *steering* prompt is an instruction the recipient executes again. Duplicate "apply fix X" / "start the deploy" / "open the PR" prompts arrive under different ids, so neither the recipient nor a human reading the transcript can distinguish them from a genuine repeat request, and the recipient burns a turn per copy. Combined with the documented `"queued"` semantics, the sender gets no completion signal either, so today the duplicate is undetectable at every layer *except* the journal — which this path bypasses.

## Open questions

- What actually drives a `send_message` round trip past 30s? The supervisor serializes mutating commands (`mutationDrain`) and awaits `waitForPromptAdmission` and an `idleEvictionFence` before dispatch; we have not isolated which of these dominates, and we may be misreading the cause. Guidance here would help us characterize the trigger precisely.
- Is the per-instance `protocolClientId` deliberate — i.e. is a fresh `DaemonClient` intended to be a distinct idempotency domain? If so, fix 1 needs to be a pinned-identity option rather than a hoisted client.
- Should `sendRemoteAgentSessionMessage` retry at all given the journal, or should it surface `command_result_uncertain` to the caller?

## Scoping note

The Python kernel layer does not contribute. `prime-agent-runtime/src/rlm/__init__.py` `host_request` (L84-140) opens a Jupyter comm and ends with a bare `return await future` — no timeout, no retry — so `await agent_message.send(...)` blocks indefinitely and never re-sends. The duplication originates at the daemon socket hop.

Adjacent: #758 and #759 cover the sender's inability to tell when a child finished. This is the same reconciliation gap on the send side.

Hotragn:
Fixes #821.

The supervisor already has durable idempotency for mutating daemon commands: `CommandRecoveryJournal` keys completed results on `createCommandIdempotencyKey(clientId, commandId)` and replays the stored response instead of re-executing. `send_message` is mutating, so it is covered in principle.

`DaemonMode.sendRemoteAgentSessionMessage` — the one loop in the tree that re-sends `send_message` across the accept window — steps outside it. It constructs a **new `DaemonClient` on every attempt**, and both halves of the journal key are per-instance:

```ts
private requestId = 0;
private readonly protocolClientId = `daemon-client:${randomUUID()}`;
```

So each retry presented a different key for the same logical message. The journal could not recognize the duplicate, `acceptAgentSessionMessage` performs no dedupe of its own, and the target received N independent steering prompts under N distinct `agentmsg_` ids.

The window is real rather than theoretical. `DaemonClient.request` times out **locally** and sends no cancellation, so a timeout does not mean the supervisor declined the message — it may already have accepted and enqueued it. A socket drop mid-response or a supervisor restart has the same shape. The existing `receivedResponse` guard correctly refuses to retry once a response was *parsed*, but that is precisely the case this window excludes.

## Change

Let a caller pin the identity, and have the one loop that needs it do so:

- `DaemonClient` takes an optional `clientId`, used in place of the per-instance uuid.
- `DaemonClientRequestOptions` takes an optional `commandId`, used in place of `daemon_${++this.requestId}`.
- `sendRemoteAgentSessionMessage` mints one `clientId` + `commandId` per logical send, before the retry loop, and passes them to every attempt.

Behaviour after the change:

- Retry after the supervisor **completed** the command: the journal reports `complete` and replays the recorded receipt. Exactly one message is enqueued and the caller gets the real receipt rather than an error.
- Retry that races an attempt **still in flight**: the journal reports `pending` and the supervisor answers `command_result_uncertain`. That is a definite answer instead of a silent duplicate, which is what the journal was built to provide.
- Retry where nothing reached the supervisor (`connect`/`waitForHello` failed, the common restart case): no journal entry exists, so the attempt proceeds normally. Unchanged.
- The journal file is append-only and reloaded on construction, so the key also survives a supervisor restart.

Callers that pass neither option keep a fresh identity per client, so every other `DaemonClient` consumer is unchanged — this is opt-in, not a global behaviour change. Reusing a command id while it is still in flight on the *same* client would clobber that client's pending-request map, so it is rejected with a clear error.

No protocol or schema change: `clientId` and `id` are existing envelope fields, and this only controls what is put in them.

## Tests

`packages/coding-agent/test/suite/regressions/821-agent-message-retry-idempotency.test.ts` (4 cases), using the same mocked-socket harness as `daemon-client.test.ts` so it needs no real daemon:

- Two separate `DaemonClient` instances with a pinned identity write envelopes whose `clientId` and `id` are identical, and `createCommandIdempotencyKey` over the two envelopes produces the same key — the defect stated directly.
- Two unpinned clients still produce different keys, proving the default is untouched.
- Driving the real `CommandRecoveryJournal` with the two captured identities: the first `begin` is `new`, the recorded receipt is stored, and the retry's `begin` returns `complete` with that same receipt instead of admitting a second command. The same test then shows an unpinned identity would have been admitted as `new` — the duplicate steering prompt.
- Reusing an in-flight command id on one client rejects.

Verified: `npm run check` clean; the new suite plus `daemon-client.test.ts` pass (33). `command-recovery-journal.test.ts` has one pre-existing Windows-only failure (`EPERM` on directory `fsync`, tracked as #666) that does not touch this path.

> [!NOTE]
> ### Fix duplicate steering prompts on retried agent messages by pinning supervisor journal keys
> - When a `send_message` transport call fails after the supervisor has already accepted the message, retries previously enqueued duplicate steering prompts. This fix pins a stable `clientId` and `commandId` for the duration of each `send_message` operation so all retry attempts share the same supervisor journal key and replay the recorded receipt instead.
> - `DaemonClient` constructor now accepts an optional `clientId` to pin the protocol client identity, and `DaemonClient.request` accepts an optional `commandId` to pin the wire command id.
> - The client rejects a request if the chosen `commandId` is already in-flight on the same client instance, preventing accidental duplicate submissions.
> - A regression test suite in [`821-agent-message-retry-idempotency.test.ts`](https://github.com/PrimeIntellect-ai/prime-agent/pull/892/files#diff-c1fdb7c44b5eb3090716b4c88590d49d6fbfdd61d589459ca0910939202b805f) covers pinned and default identity behavior, supervisor journal replay, and in-flight duplicate rejection.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized d5cc2c1.</sup>
> 
>

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -20,6 +20,7 @@
 - Fixed workers with no live connection reporting as `ready`; stopping workers now report a `stopping` state, are hidden from live sessions, and no longer receive daemon-wide commands ([#850](https://github.com/PrimeIntellect-ai/prime-agent/pull/850)).
 - Fixed timed-out worker stops stranding dead-but-registered workers ("Session worker is not connected"); stops now finalize in the background once the process exits, and zombie processes are no longer counted as alive ([#851](https://github.com/PrimeIntellect-ai/prime-agent/pull/851)).
 - Fixed sessions becoming permanently unopenable after a stale worker registration was left behind; open/resume now self-heals by finishing the old cleanup and starting a fresh worker ([#852](https://github.com/PrimeIntellect-ai/prime-agent/pull/852)).
+- Fixed a retried agent message delivering duplicate steering prompts when the transport failed after the supervisor had already accepted it ([#821](https://github.com/PrimeIntellect-ai/prime-agent/issues/821))
 
 ## [0.7.1] - 2026-08-07
 
--- a/packages/coding-agent/src/modes/daemon/daemon-client.ts
+++ b/packages/coding-agent/src/modes/daemon/daemon-client.ts
@@ -6,10 +6,12 @@ import {
 	createDaemonCommandEnvelope,
 	DAEMON_COMMAND_ENVELOPE_MIN_PROTOCOL_VERSION,
 	DAEMON_PROTOCOL_VERSION,
+	type DaemonClientId,
 	type DaemonClosingReason,
 	type DaemonCommand,
 	type DaemonCommandCompatibility,
 	type DaemonCommandEnvelope,
+	type DaemonCommandId,
 	type DaemonOutbound,
 	type DaemonProtocolVersion,
 	type DaemonRequestProgress,
@@ -34,6 +36,24 @@ export type DaemonClientProgressListener = (message: DaemonRequestProgress) => v
 
 export interface DaemonClientRequestOptions {
 	onProgress?: DaemonClientProgressListener;
+	/**
+	 * Wire id for this command, overriding the per-instance counter.
+	 *
+	 * The supervisor keys its command journal on [clientId, commandId], so a caller
+	 * that retries a mutating command across separate client instances must present
+	 * the same pair on every attempt for the retry to be recognized as a duplicate
+	 * rather than executed again.
+	 */
+	commandId?: DaemonCommandId;
+}
+
+export interface DaemonClientOptions {
+	/**
+	 * Protocol client id to present instead of a fresh per-instance one. Pair with
+	 * {@link DaemonClientRequestOptions.commandId} to keep a retried mutating command
+	 * on one journal key.
+	 */
+	clientId?: DaemonClientId;
 }
 
 interface PendingDaemonRequest {
@@ -111,7 +131,7 @@ export class DaemonClient {
 	private readonly closeListeners = new Set<DaemonClientCloseListener>();
 	private readonly pendingRequests = new Map<string, PendingDaemonRequest>();
 	private requestId = 0;
-	private readonly protocolClientId = `daemon-client:${randomUUID()}`;
+	private readonly protocolClientId: DaemonClientId;
 	private requestRecoveryEnabled = false;
 	private reconnectOptions?: DaemonClientReconnectOptions;
 	private autoReconnectPromise?: Promise<void>;
@@ -125,7 +145,12 @@ export class DaemonClient {
 		timeout: ReturnType<typeof setTimeout>;
 	}>();
 
-	constructor(private readonly socketPath: string) {}
+	constructor(
+		private readonly socketPath: string,
+		options: DaemonClientOptions = {},
+	) {
+		this.protocolClientId = options.clientId ?? `daemon-client:${randomUUID()}`;
+	}
 
 	get hello(): DaemonHello | undefined {
 		return this.helloMessage;
@@ -353,7 +378,10 @@ export class DaemonClient {
 			);
 		}
 
-		const id = `daemon_${++this.requestId}`;
+		const id = options.commandId ?? `daemon_${++this.requestId}`;
+		if (this.pendingRequests.has(id)) {
+			throw new Error(`Daemon command id "${id}" is already in flight on this client`);
+		}
 		const fullCommand = { ...command, id } as DaemonCommand | DaemonWorkerCommand;
 		const wireCommand: DaemonCommand | DaemonWorkerCommand | DaemonCommandEnvelope = publicEnvelopeProtocolVersion
 			? createDaemonCommandEnvelope(
--- a/packages/coding-agent/src/modes/daemon/daemon-mode.ts
+++ b/packages/coding-agent/src/modes/daemon/daemon-mode.ts
@@ -5463,8 +5463,18 @@ export class AgentDaemon {
 		}
 		const deadline = Date.now() + 30_000;
 		let lastError: unknown;
+		// The supervisor journals mutating commands under [clientId, commandId] and
+		// replays a completed result instead of re-executing. A fresh DaemonClient
+		// mints a new protocol client id and restarts its request counter, so retrying
+		// on a new client presented a new key for the same logical message and the
+		// supervisor enqueued another steering prompt. Pin both halves of the key for
+		// the lifetime of this send so every attempt is recognized as the same command.
+		const idempotency = {
+			clientId: `daemon-client:${randomUUID()}`,
+			commandId: `send_message_${randomUUID()}`,
+		};
 		while (Date.now() < deadline && !this.shuttingDown) {
-			const client = new DaemonClient(supervisorSocketPath);
+			const client = new DaemonClient(supervisorSocketPath, { clientId: idempotency.clientId });
 			let receivedResponse = false;
 			try {
 				await client.connect(1000);
@@ -5478,6 +5488,7 @@ export class AgentDaemon {
 						agentOrigin: true,
 					},
 					30_000,
+					{ commandId: idempotency.commandId },
 				);
 				receivedResponse = true;
 				if (!response.success) {
--- a/packages/coding-agent/test/suite/regressions/821-agent-message-retry-idempotency.test.ts
+++ b/packages/coding-agent/test/suite/regressions/821-agent-message-retry-idempotency.test.ts
@@ -1,0 +1,237 @@
+import { mkdtempSync, rmSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import {
+	CommandRecoveryJournal,
+	createCommandIdempotencyKey,
+} from "../../../src/modes/daemon/command-recovery-journal.js";
+import { DaemonClient } from "../../../src/modes/daemon/daemon-client.js";
+import {
+	DAEMON_PROTOCOL_VERSION,
+	type DaemonCommandEnvelope,
+	type DaemonResponse,
+} from "../../../src/modes/daemon/daemon-protocol.js";
+
+const netMock = vi.hoisted(() => {
+	type Listener = (...args: unknown[]) => void;
+	type TrackedListener = Listener & { originalListener?: Listener };
+
+	class MockSocket {
+		private readonly listeners = new Map<string, Set<TrackedListener>>();
+		readonly writes: string[] = [];
+		destroyed = false;
+
+		constructor(readonly path: string) {}
+
+		on(event: string, listener: Listener): this {
+			const listeners = this.listeners.get(event) ?? new Set<TrackedListener>();
+			listeners.add(listener as TrackedListener);
+			this.listeners.set(event, listeners);
+			return this;
+		}
+
+		once(event: string, listener: Listener): this {
+			const onceListener: TrackedListener = (...args) => {
+				this.off(event, onceListener);
+				listener(...args);
+			};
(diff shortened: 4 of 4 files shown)


kevinjosethomas:

PR #909, opened 2026-08-08
fix(coding-agent): sanitize side-question tool history

### LINKED ISSUE #886: `/btw` and `/side` fail on Bedrock provider: missing toolConfig when message history contains tool blocks
## Summary

`/btw` and `/side` crash with a Bedrock validation error in any session where the agent has used tools (ipython, bash, etc.). The side-question agent copies the full parent message history — including `toolUse`/`toolResult` content blocks — but sends `tools: []`, so Bedrock's Converse API rejects the request.

**Version:** 0.7.x · **Provider:** amazon-bedrock · **Model:** us.anthropic.claude-opus-4-6-v1

## Error

```
Validation error: The toolConfig field must be defined when using toolUse and toolResult content blocks.
```

Happens on both `/btw <question>` and `/side <question>`. Works fine in a fresh session before any tool calls are made.

## Cause

`dist/core/side-question.js` — `startSideQuestion` initializes the side agent with:

```js
const sideAgent = new Agent({
    initialState: {
        ...
        messages: [...structuredClone(parent.state.messages), ...previousTurnMessages],
        tools: [],
        ...
    },
    ...
});
```

The cloned `parent.state.messages` contains `toolUse`/`toolResult` blocks from prior turns. Amazon Bedrock's Converse API requires `toolConfig` in the request whenever tool content blocks appear in the message array — regardless of whether the model is expected to call tools. The direct Anthropic API is lenient here, so this only surfaces on Bedrock.

## Suggested fix

Cleanest option: strip `toolUse` and `toolResult` content blocks (and their paired messages) from the cloned messages before passing to the side agent. The side agent's system prompt already says "Do not use tools," so tool-call history adds no value to the side conversation and just bloats context.

Alternative: forward the parent's tool definitions into the side agent's `tools` array so the serializer includes `toolConfig` in the Bedrock payload.

---

<details>
<summary>AI usage disclosure</summary>

This bug was discovered by a human user (hitting the error in normal usage). Root-cause analysis and source tracing was done with prime-agent itself — the agent read `dist/core/side-question.js`, identified the `tools: []` + cloned messages mismatch, and confirmed no existing issue covered it.
</details>

chai1any:
## Summary

- remove tool-call blocks and paired tool-result messages from side-question LLM context
- coalesce adjacent same-role messages after filtering so the Bedrock conversation still alternates roles
- keep `/btw` and `/side` tool-free while preserving ordinary user and assistant conversation content
- cover mixed assistant content, tool results, role alternation, and the empty `tools` invariant in regression tests

Amazon Bedrock requires `toolConfig` whenever history contains `toolUse` or `toolResult` blocks and requires conversation roles to alternate. Side questions intentionally provide no tools, so forwarding the main thread's tool protocol history caused the request to fail after any tool use.

## Verification

- `npx tsx ../../node_modules/vitest/dist/cli.js --run test/suite/regressions/4509-side-questions.test.ts test/suite/regressions/886-bedrock-side-questions.test.ts` (39 tests passed)
- `npm run check`

Fixes #886

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -5,6 +5,7 @@
 - Added privacy-safe pseudonymous product analytics for onboarding, command use, execution modes, run outcomes, TTFT, latency, usage, tools, retries, and compactions, with disclosure and opt-out controls ([ENG-4682](https://linear.app/primeintellect/issue/ENG-4682/add-privacy-safe-posthog-analytics-to-prime-agent)).
 - Changed sent agent messages in the IPython cell UI to show only the message text with a `╰─` gutter when expanded, matching received messages, and hid the raw `agent_message.send` receipt dictionary.
 - Fixed Homebrew installs attempting to self-update their versioned Cellar keg instead of directing users to `brew upgrade prime-agent` ([#844](https://github.com/PrimeIntellect-ai/prime-agent/issues/844))
+- Fixed `/btw` and `/side` failing on Amazon Bedrock after the main conversation used tools ([#886](https://github.com/PrimeIntellect-ai/prime-agent/issues/886)).
 
 ## [0.7.1] - 2026-08-07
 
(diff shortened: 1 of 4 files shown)


kevinjosethomas:

PR #910, opened 2026-08-08
fix(coding-agent): reclaim reused Termux lease pids

### LINKED ISSUE #868: Session lease liveness check fails on Android/Termux: dead pid returns EPERM → permanent SessionAlreadyActiveError
## Environment

- Android aarch64, Termux — a supported platform with official docs (`packages/coding-agent/docs/termux.md`; Termux is listed in the README's Platform notes)
- Source install, Node v24, daemon mode

## Symptom

Resuming a session whose worker has been stopped (e.g. by idle eviction) always fails with:

```
SessionAlreadyActiveError: Session is already active in <activeSessionId>: .../<session>.jsonl
```

even though the worker process that owned the session lease is long dead (SIGKILLed, descriptor cleaned up, socket gone).

## Root cause

Three layers combine to make the lease liveness check permanently report "alive":

1. `SessionLease` (src/core/session-lease.ts) is released only on graceful shutdown (`release()`). A worker killed with SIGKILL (e.g. idle eviction that did not stop cleanly, later adopted with SIGKILL on daemon restart) leaves `~/.prime/agent/session-leases/<hash>.lock/owner.json` behind.
2. `isProcessAlive` (session-lease.ts:104) uses `process.kill(pid, 0)`. On Android/Termux a **dead pid that has been reused by another (SELinux-protected) process returns EPERM, not ESRCH**, and the code conservatively treats EPERM as alive:

   ```ts
   function isProcessAlive(pid: number): boolean {
       try { process.kill(pid, 0); return true; }
       catch (error) { return (error as NodeJS.ErrnoException).code === "EPERM"; }
   }
   ```
3. The second guard `isLeaseOwnerAlive` (session-lease.ts:177) compares `processStartId` read from `/proc/<pid>/stat`. On Termux the reused pid's `/proc/<pid>/stat` is invisible (ENOENT), the `ps` fallback also returns nothing, so `getProcessStartId` returns `undefined` — and `undefined` is treated as alive:

   ```ts
   const currentStartId = getProcessStartId(owner.pid);
   return currentStartId === undefined || currentStartId === owner.processStartId;
   ```

Result: `acquireSessionLease` never reaches `reclaimStaleLease` and the session can never be resumed without manually deleting the lease directory.

## Verified on device

```
pid 99999          (never existed)        -> kill(0) ESRCH      /proc/<pid>/stat ENOENT
pid 26617          (dead, pid reused)     -> kill(0) EPERM ❌   /proc/<pid>/stat ENOENT ✅
pid 29551/30447    (alive, same uid)      -> kill(0) OK         /proc/<pid>/stat OK
```

So `/proc/<pid>/stat`'s ENOENT semantics are reliable for same-uid processes where `kill(0)` EPERM is not.

## Suggested fixes

**Minimal (3 lines):** in `isProcessAlive`, on EPERM also check `/proc/<pid>/stat`; if it is ENOENT, treat the process as dead (falling back to the current conservative behavior only when the proc entry is readable).

**Robust (recommended):** make the lease an ownership lock held with `flock`/`fcntl` (e.g. keep the lock file open for the lease duration). The kernel guarantees the lock is released when the owning process dies — including SIGKILL — with no dependency on pid semantics, `/proc` visibility, or `kill(0)` behavior. Verified working on Termux: a SIGKILLed flock holder's lock is released immediately and a new acquirer succeeds.

## Related issues

Same stale-session-lease failure family, different platform-specific fault point:

- **#667** / **#841** (Windows): `renameSync` onto an existing lease directory returns `EPERM`/`EACCES` on Windows instead of POSIX `EEXIST`/`ENOTEMPTY`, so `reclaimStaleLease` is never reached. PR **#727** fixes that with `isRenameTargetExistsError`.
- **This issue (#868)** (Android/Termux): the rename succeeds (POSIX semantics), but `isLeaseOwnerAlive` misjudges a dead owner as alive — `kill(pid, 0)` returns `EPERM` for a pid reused by a SELinux-protected process, and the `/proc` starttime fallback returns `undefined`, which is also treated as alive. So `reclaimStaleLease` is again never reached, and **PR #727 does not cover this case** (the failing check here is in `isLeaseOwnerAlive`, not in the rename error handling).

A robust fix that works everywhere would be to hold the lease as an `flock`/`fcntl` ownership lock (kernel releases it automatically when the owning process dies, independent of pid semantics and `/proc` visibility).

chai1any:
## Summary

- refine `kill(pid, 0)` `EPERM` handling on Linux and Android with a `/proc/<pid>/stat` probe
- treat a missing proc entry as a stale or reused lease-owner PID so the session lease can be reclaimed
- retain conservative alive behavior for readable proc entries, other procfs errors, and non-procfs platforms
- add platform-specific liveness regression coverage

On Termux, a dead worker PID can be reused by a SELinux-protected process. The signal probe then returns `EPERM` while the proc entry is hidden as `ENOENT`; previously both conditions were treated as proof that the original lease owner was still alive.

## Verification

- `npx tsx ../../node_modules/vitest/dist/cli.js --run test/session-lease.test.ts` (13 tests passed)
- `npm run check`

Fixes #868

> [!NOTE]
> ### Fix stale session lease detection for reused PIDs on Android/Termux
> On Linux/Android, `isProcessAlive` previously treated `EPERM` from `process.kill(pid, 0)` as alive, which caused stale leases to become permanent when a dead worker's PID was reused by a protected process.
>
> - `isProcessAlive` in [`session-lease.ts`](https://github.com/PrimeIntellect-ai/prime-agent/pull/910/files#diff-74865e03a1e2d7f4025a576ca8f4c09dbc1ec7c9e2776b5d93f2691ee29e6cda) now falls back to reading `/proc/<pid>/stat` when `EPERM` is received on Linux/Android; `ENOENT` in procfs means the process is dead.
> - On non-procfs platforms (e.g. Darwin), `EPERM` continues to be treated as alive.
> - The function is now exported and accepts injectable `ProcessProbe` dependencies for testing.
> - Behavioral Change: PIDs returning `EPERM` on Linux/Android are now considered dead if they have no `/proc` entry, changing the previous unconditional alive result.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized c30502b.</sup>
> 
>

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -5,6 +5,7 @@
 - Added privacy-safe pseudonymous product analytics for onboarding, command use, execution modes, run outcomes, TTFT, latency, usage, tools, retries, and compactions, with disclosure and opt-out controls ([ENG-4682](https://linear.app/primeintellect/issue/ENG-4682/add-privacy-safe-posthog-analytics-to-prime-agent)).
 - Changed sent agent messages in the IPython cell UI to show only the message text with a `╰─` gutter when expanded, matching received messages, and hid the raw `agent_message.send` receipt dictionary.
 - Fixed Homebrew installs attempting to self-update their versioned Cellar keg instead of directing users to `brew upgrade prime-agent` ([#844](https://github.com/PrimeIntellect-ai/prime-agent/issues/844))
+- Fixed stale session leases becoming permanent on Android/Termux when a dead worker PID is reused by a protected process ([#868](https://github.com/PrimeIntellect-ai/prime-agent/issues/868)).
 
 ## [0.7.1] - 2026-08-07
 
--- a/packages/coding-agent/src/core/session-lease.ts
+++ b/packages/coding-agent/src/core/session-lease.ts
@@ -101,12 +101,38 @@ function readLeaseOwner(directory: string): SessionLeaseOwner | undefined {
 	}
 }
 
-function isProcessAlive(pid: number): boolean {
+type ProcessProbe = (pid: number) => void;
+
+const probeProcessSignal: ProcessProbe = (pid) => {
+	process.kill(pid, 0);
+};
+
+const probeProcStat: ProcessProbe = (pid) => {
+	readFileSync(`/proc/${pid}/stat`);
+};
+
+export function isProcessAlive(
+	pid: number,
+	signalProbe: ProcessProbe = probeProcessSignal,
+	procStatProbe: ProcessProbe = probeProcStat,
+	platform: NodeJS.Platform = process.platform,
+): boolean {
 	try {
-		process.kill(pid, 0);
+		signalProbe(pid);
 		return true;
 	} catch (error) {
-		return (error as NodeJS.ErrnoException).code === "EPERM";
+		if ((error as NodeJS.ErrnoException).code !== "EPERM") {
+			return false;
+		}
+	}
+	if (platform !== "linux" && platform !== "android") {
+		return true;
+	}
+	try {
+		procStatProbe(pid);
+		return true;
+	} catch (error) {
+		return (error as NodeJS.ErrnoException).code !== "ENOENT";
 	}
 }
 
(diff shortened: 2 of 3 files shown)


kevinjosethomas:

PR #924, opened 2026-08-08
feat(ai): add fast mode support for claude-opus-5 and claude-opus-4-8

### LINKED ISSUE #867: Support Opus 5 and Opus 4.8 fast on Anthropic API
These don't currently work as they're not individual models but use a speed and anthropic-beta header.

Xsidz:
## Summary

Adds fast mode support for \`claude-opus-5\` and \`claude-opus-4-8\` on the Anthropic API (Claude API only — not Bedrock, Vertex, or Foundry).

Fast mode enables up to 2.5× higher output tokens/second at premium pricing. Per the Anthropic docs, three things are required:

1. Use the beta messages endpoint (\`client.beta.messages.create\`)
2. Pass the \`fast-mode-2026-02-01\` beta header
3. Set \`speed: "fast"\` as a top-level request body parameter

**Changes:**

- \`packages/ai/src/models.ts\` — extended \`supportsFastMode\` to return \`true\` for \`claude-opus-5\` and \`claude-opus-4-8\` on the \`anthropic\` provider
- \`packages/ai/src/providers/anthropic.ts\`:
  - Added \`FAST_MODE_BETA = "fast-mode-2026-02-01"\` constant
  - Added \`speed?: "fast"\` to \`AnthropicOptions\`
  - Added \`useFastMode\` param to \`createClient\`; pushes the beta to \`betaFeatures\` when set
  - Updated the \`createClient\` call site to pass \`options?.speed === "fast"\`
  - Sets \`speed: "fast"\` on the params in \`buildParams\` when fast mode is active
  - Routes through \`client.beta.messages.create\` instead of \`client.messages.create\` when fast mode is active
- \`packages/ai/test/fast-mode.test.ts\` — added 4 new test cases covering supported models, unsupported models, and API mismatch

## Test plan

- [x] \`npm run check\` passes (biome, tsgo, installer render, browser smoke)
- [x] \`test/fast-mode.test.ts\` — 9/9 tests pass including 4 new Anthropic cases
- [x] All Anthropic unit tests pass: \`anthropic-sse-parsing\`, \`anthropic-eager-tool-input-compat\`, \`anthropic-thinking-disable\`, \`anthropic-tool-name-normalization\`, \`xhigh\`, \`supports-xhigh\`, \`interleaved-thinking\` — 50 pass, 0 fail
- [x] Broader unit suite — 87 pass, 0 fail

Fixes #867

> [!NOTE]
> ### Add fast mode support for `claude-opus-5` and `claude-opus-4-8` on Anthropic API
> - Adds a `speed: "fast"` option to `AnthropicOptions` that routes requests through the `fast-mode-2026-02-01` beta endpoint for `claude-opus-5` and `claude-opus-4-8`.
> - Fast mode is also activated when `serviceTier` is `"priority"` on supported models; costs are doubled to reflect 2× pricing.
> - Adds `supportsAnthropicFastMode(model)` in [models.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/924/files#diff-9d03087de13641e60bdb725ae786441aa71caf6626582728660f2a30328138b7) so callers and agent sessions can detect fast-mode-capable models.
> - Updates `AgentSession` and `createAgentSession` in the coding-agent package to preserve the `priority` service tier for Anthropic fast-mode models instead of demoting to `default`.
> - Auth errors during fast mode include a hint that preview access is required.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized d808690.</sup>
> 
>

### DIFF
--- a/packages/ai/CHANGELOG.md
+++ b/packages/ai/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [Unreleased]
 
 ## [0.7.2] - 2026-08-11
+- Added fast mode support (`speed: "fast"`) for `claude-opus-5` and `claude-opus-4-8` on the Anthropic API, enabling up to 2.5× higher output tokens/second via the `fast-mode-2026-02-01` beta ([#867](https://github.com/PrimeIntellect-ai/prime-agent/issues/867)).
 
 ## [0.7.1] - 2026-08-07
 
--- a/packages/ai/src/models.ts
+++ b/packages/ai/src/models.ts
@@ -37,15 +37,20 @@ export function getModels<TProvider extends KnownProvider>(
 }
 
 export function supportsFastMode<TApi extends Api>(model: Model<TApi>): boolean {
-	if (model.provider === "openai-codex" && model.api === "openai-codex-responses") {
-		return (
-			model.id === "gpt-5.4" || model.id === "gpt-5.5" || model.id === "gpt-5.6" || model.id.startsWith("gpt-5.6-")
-		);
-	}
-	if (model.provider === "anthropic" && model.api === "anthropic-messages") {
-		return model.id === "claude-opus-5" || model.id === "claude-opus-4-8";
-	}
-	return false;
+	return (
+		model.provider === "openai-codex" &&
+		model.api === "openai-codex-responses" &&
+		(model.id === "gpt-5.4" || model.id === "gpt-5.5" || model.id === "gpt-5.6" || model.id.startsWith("gpt-5.6-"))
+	);
+}
+
+/** Whether the model supports Anthropic fast mode (speed="fast" + fast-mode-2026-02-01 beta). */
+export function supportsAnthropicFastMode<TApi extends Api>(model: Model<TApi>): boolean {
+	return (
+		model.provider === "anthropic" &&
+		model.api === "anthropic-messages" &&
+		(model.id === "claude-opus-5" || model.id === "claude-opus-4-8")
+	);
 }
 
 export interface CostOverrides {
--- a/packages/ai/src/models.ts
+++ b/packages/ai/src/models.ts
@@ -37,11 +37,15 @@ export function getModels<TProvider extends KnownProvider>(
 }
 
 export function supportsFastMode<TApi extends Api>(model: Model<TApi>): boolean {
-	return (
-		model.provider === "openai-codex" &&
-		model.api === "openai-codex-responses" &&
-		(model.id === "gpt-5.4" || model.id === "gpt-5.5" || model.id === "gpt-5.6" || model.id.startsWith("gpt-5.6-"))
-	);
+	if (model.provider === "openai-codex" && model.api === "openai-codex-responses") {
+		return (
+			model.id === "gpt-5.4" || model.id === "gpt-5.5" || model.id === "gpt-5.6" || model.id.startsWith("gpt-5.6-")
+		);
+	}
+	if (model.provider === "anthropic" && model.api === "anthropic-messages") {
+		return model.id === "claude-opus-5" || model.id === "claude-opus-4-8";
+	}
+	return false;
 }
 
 export interface CostOverrides {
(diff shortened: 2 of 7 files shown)


Xsidz (commented):
Fixed in `af98c34d`:
1. Added `supportsAnthropicFastMode()` as a separate function from `supportsFastMode()` — Anthropic fast mode (`speed="fast"`) and the Codex priority service tier (`serviceTier="priority"`) are unrelated features. The existing `supportsFastMode` gate for `serviceTier` is unchanged.
2. All three fast-mode paths in `anthropic.ts` (beta header, `speed` body param, `client.beta.messages` routing) are now guarded with `supportsAnthropicFastMode(model)`, so unsupported providers and models are never affected.
3. Updated the coding-agent CHANGELOG to clarify this is an AI SDK addition.

BILLKISHORE:
Two things worth checking before this merges.

**Pricing.** Fast mode bills at $10/MTok input and $50/MTok output, against Opus 5's catalog entry of `cost: { input: 5, output: 25 }`. This diff changes no cost path, so a fast request is costed at standard rates and fast usage is under-reported by half. `openai-codex-responses` covers the equivalent case in `applyServiceTierPricing`, scaling all four cost components after `calculateCost`.

**Reachability.** The diff touches no coding-agent source, so nothing sets `options.speed`. The agent gates its fast toggle on `supportsFastMode` (`interactive-mode.ts:7679`, `agent-session.ts:6820`, `sdk.ts:249`), and this PR deliberately keeps that false for Anthropic, so selecting fast on an Opus model still sends a standard request. The coding-agent CHANGELOG entry describes a user-facing change that no code path reaches yet.

The model set looks right. Opus 4.7 returns an error for `speed: "fast"` and Opus 4.6 silently runs at standard speed while billing standard rates, so excluding both is correct.

I had a branch for #867 going before I saw this one. Not opening it; happy to send the pricing multiplier and the wiring as a follow-up if that is useful.

Xsidz:
Good catches, both of them. Fixed in f9373239.

Pricing: added a 2x multiplier after both calculateCost call sites in anthropic.ts, the same pattern applyServiceTierPricing uses for Codex. The catalog entry stays at $5/$25 since that's the standard rate; the multiplier only fires when fast mode is actually active.

Reachability: serviceTier="priority" on claude-opus-5 and claude-opus-4-8 now passes through all four supportsFastMode gates in agent-session.ts and sdk.ts by also checking supportsAnthropicFastMode, so the existing priority toggle in the UI activates fast mode without needing a new option. The CHANGELOG now describes the user-visible behavior rather than the internal plumbing.

BILLKISHORE:
Both fixes came through the restructure intact, and splitting `supportsAnthropicFastMode` from `supportsFastMode` is a better shape than widening the Codex gate.

One thing worth thinking about before this lands: fast mode is a gated research preview. The docs say access is granted per account ("contact your account manager", otherwise a waitlist), so most accounts cannot use it today.

On current `main`, `serviceTier` is never read in `anthropic.ts`, so toggling fast mode on an Anthropic model is a harmless no-op. After this PR the same toggle sends `speed: "fast"` with the beta header, and there is no fallback in the diff. For an account without preview access that turns a no-op into a failed request, on a toggle that `/fast` presents as a plain on/off.

The existing degradation at `agent-session.ts:9027` does not cover it, since it tests model capability and access here is an account property that no model check can see.

A couple of options depending on how defensive you want to be. Catching the access error once and retrying without `speed` would keep the toggle harmless for accounts that lack it, at the cost of one wasted round trip. Alternatively, leaving the request to fail but surfacing a message that names fast mode as the reason would at least make the failure legible, since the current path gives the user a bare API error from a toggle they just flipped.

One smaller naming point: the Anthropic docs state fast mode is not available alongside a Priority Tier commitment. Prime Agent's `serviceTier: "priority"` is a local concept and never reaches Anthropic on the wire, so there is no actual conflict, but an account that has both a Priority Tier commitment and fast-mode access would find "priority" meaning the opposite of what Anthropic means by it. Probably worth a sentence in the docs.

Xsidz:
Fixed in 7e63a1e3.

When a fast-mode request fails with an auth/permission error, the error message is now prefixed with "Fast mode requires preview access for your account (contact your Anthropic account manager)." so the user sees a clear explanation rather than a bare API error.

On the naming point: agreed that `serviceTier: "priority"` is a local Prime Agent concept and never reaches Anthropic on the wire, so there's no real conflict. A docs note would be a good follow-up — leaving that out of this PR.

kevinjosethomas:

Xsidz:
Got it. Will open a Discussion for this.

PR #966, opened 2026-08-08
fix(coding-agent): invalidate snapshot cache when worker recovery drops operations

avion23:
## What changed

- Invalidate the supervisor snapshot cache for every unique active session with uncertain operations dropped during worker recovery, including child records that do not have a `sessionFile`.
- Keep unexpected duplicate-snapshot byte or protocol mismatches fatal to the worker channel; recovery-time invalidation removes the known stale-cache cause before the worker reconnects.
- Add recovery scenario coverage for root and child sessions, pending snapshot readers, transcript disposal, and isolation of unrelated cached snapshots.

## Why

Snapshot IDs are derived from event counters rather than transcript content. When recovery drops an in-flight operation, the recovered transcript can have different bytes under the same snapshot ID. A stale supervisor cache then treats the transfer as a duplicate and rejects it, which can leave attached clients unable to re-sync.

Recovery journal entries can omit `sessionFile` for non-root sessions. Catalog interruption marking still needs a session file, but cache invalidation does not. Separating those sets ensures every uncertain active session is invalidated without weakening duplicate-stream containment.

## Verification

- `npx tsx ../../node_modules/vitest/dist/cli.js --run test/daemon-supervisor-monitor.test.ts test/suite/regressions/4602-snapshot-transfer-idempotency.test.ts test/worker-recovery-journal.test.ts` — 56 passed
- `npm run check` — passed (`biome`, `tsgo`, installer render, browser smoke)

> [!NOTE]
> ### Invalidate snapshot cache for uncertain sessions during worker recovery in `DaemonSupervisor`
> When `DaemonSupervisor.recoverUncertainWorkerOperations` drops uncertain operations, it now collects the affected active session IDs and calls `failWorkerSnapshotCache` for each one, preventing stale cached snapshots from being used after recovery.
>
> - Uncertain sessions have their snapshot cache and transcript caches disposed; unrelated sessions are unaffected.
> - A new test verifies that each uncertain session's cache reader rejects with an invalidation error and that unrelated session caches remain valid.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 345c0e6.</sup>
> 
>

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [Unreleased]
 
+- Fixed the daemon supervisor leaving a stale snapshot cache after a worker recovery that dropped uncertain operations, which could make an attached client fail to re-sync its snapshot ([#966](https://github.com/PrimeIntellect-ai/prime-agent/pull/966) by [@avion23](https://github.com/avion23)).
 - Added privacy-safe pseudonymous product analytics for onboarding, command use, execution modes, run outcomes, TTFT, latency, usage, tools, retries, and compactions, with disclosure and opt-out controls ([ENG-4682](https://linear.app/primeintellect/issue/ENG-4682/add-privacy-safe-posthog-analytics-to-prime-agent)).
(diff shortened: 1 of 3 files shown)


kevinjosethomas:

PR #989, opened 2026-08-08
fix(coding-agent): write settings and credentials atomically and migrate auth safely

### LINKED ISSUE #983: Settings and credential files can be silently lost: non-atomic writes and unsafe auth migration order
Three related problems in how `settings.json` and `auth.json` are written. A crash or a second concurrent process at the wrong moment can silently lose all user settings or all stored credentials.

## 1. `FileSettingsStorage.withLock` reads before locking and writes in place

`packages/coding-agent/src/core/settings-manager.ts:256-284`. When the file does not exist yet, no lock is taken. `fn(current)` computes the next state from `undefined`, and only then is the lock acquired and `writeFileSync(path, next)` called directly on the target. Two processes creating `settings.json` at the same time both compute from empty state and the later write wins. A crash mid-write leaves a truncated file, and the loader then silently falls back to `{}`, dropping all settings.

`writePrimeCliConfigData` in `prime-inference-auth.ts` already does this correctly (temp file, chmod, rename), so there is a pattern to copy.

## 2. `FileAuthStorageBackend` writes `auth.json` in place

`packages/coding-agent/src/core/auth-storage.ts:206-208`, plus the sync variant. Same non-atomic write, but this file holds every OAuth refresh token and API key. A crash mid-write corrupts all of them at once.

## 3. The auth migration deletes the old data before writing the new

`packages/coding-agent/src/migrations.ts`. `migrateAuthToAuthJson` renames `oauth.json` to `oauth.json.migrated` (line 53) and deletes `apiKeys` from `settings.json` (lines 71-72) before `auth.json` is written (line 81). A crash in between leaves nothing to migrate on the next run, and the `.migrated` file is never read back. All provider credentials are effectively gone.

Writing `auth.json` first (temp + rename + chmod) and only then cleaning up the old locations would close the window.

Adolanium:
Fixes #983.

## What was broken

Three ways to silently lose settings or credentials:

1. `FileSettingsStorage.withLock` computed the next state before taking the lock when the file did not exist yet, so two concurrent first-time creators both merged from empty state and the later write won. It also wrote directly to the target, so a crash mid-write left a truncated `settings.json` that the loader silently replaced with defaults.
2. `FileAuthStorageBackend` wrote `auth.json` in place on both the sync and async paths. A crash mid-write corrupted every stored OAuth refresh token and API key at once.
3. `migrateAuthToAuthJson` renamed `oauth.json` to `.migrated` and deleted `apiKeys` from `settings.json` before `auth.json` was written. A crash in that window left nothing to migrate on the next run.

## The fix

- Settings writes now re-read under the lock before a first-time write and go through temp file + rename.
- Both auth write paths go through temp file (`0o600` at creation) + chmod + rename, mirroring the existing `writePrimeCliConfigData` pattern.
- The migration writes `auth.json` first and only then rewrites `settings.json` and renames `oauth.json`.

Locking structure, the sync retry API, and the OAuth refresh flow are unchanged.

## Testing

- New regression file `test/suite/regressions/983-credential-file-durability.test.ts` (5 tests): migration ordering asserted via recorded fs events, re-read under lock on first creation, temp+rename with no leftovers on the settings and auth paths.
- `npx tsx ../../node_modules/vitest/dist/cli.js --run` on the new file plus the existing auth-storage, settings-manager, and migrations suites: all green. (Three failures in `auth-storage.test.ts` / `auth-flows.test.ts` / `4620-fast-mode-settings` reproduce on pristine main and are unrelated, verified in a clean worktree.)
- `npm run check` clean.

> [!NOTE]
> ### Write auth.json and settings.json atomically and harden credential migration
> - Auth writes in [`FileAuthStorageBackend`](https://github.com/PrimeIntellect-ai/prime-agent/pull/989/files#diff-c1bbb0577045e4d9c03e250fa121993eb80a6129ee047c6097583a07d24df74b) now go through a new `writeAtomically` helper: content is written to a uniquely-named `0600` temp file in the same directory, then renamed over the target, avoiding truncation on crash.
> - Settings writes in [`FileSettingsStorage`](https://github.com/PrimeIntellect-ai/prime-agent/pull/989/files#diff-6e8b43ea9be5845c1f130f3a12344386eb8e1e03ca381c3577a40713e6a3d8ab) use a new `writeSettingsFileAtomically` helper with the same temp-file-and-rename approach; first-time creation re-reads under the lock to avoid losing concurrent writes.
> - [`migrateAuthToAuthJson`](https://github.com/PrimeIntellect-ai/prime-agent/pull/989/files#diff-0a96b36fc418ab60e6abdab862d571a29c56ab14753b336d965a8f9fe3320420) now writes `auth.json` durably (via atomic helper, `0600`) before removing `apiKeys` from `settings.json` or renaming `oauth.json`, reducing credential loss risk on crash mid-migration.
> - A new regression test suite in [`983-credential-file-durability.test.ts`](https://github.com/PrimeIntellect-ai/prime-agent/pull/989/files#diff-7ec3cda8008df1deb7f737282275a535a803c86ec7af09e2aa193842f907189f) verifies atomic rename behavior, lock re-read semantics, and restrictive file permissions.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 99a136c.</sup>
> 
>

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -5,6 +5,9 @@
 - Added privacy-safe pseudonymous product analytics for onboarding, command use, execution modes, run outcomes, TTFT, latency, usage, tools, retries, and compactions, with disclosure and opt-out controls ([ENG-4682](https://linear.app/primeintellect/issue/ENG-4682/add-privacy-safe-posthog-analytics-to-prime-agent)).
 - Changed sent agent messages in the IPython cell UI to show only the message text with a `╰─` gutter when expanded, matching received messages, and hid the raw `agent_message.send` receipt dictionary.
 - Fixed Homebrew installs attempting to self-update their versioned Cellar keg instead of directing users to `brew upgrade prime-agent` ([#844](https://github.com/PrimeIntellect-ai/prime-agent/issues/844))
+- Fixed concurrent first-time `settings.json` writes losing each other's changes, and crash-interrupted writes leaving a truncated file that silently reset all settings; updates now hold the file lock before reading and write through a temp file and rename ([#983](https://github.com/PrimeIntellect-ai/prime-agent/issues/983))
+- Fixed crash-interrupted `auth.json` writes risking loss of all stored API keys and OAuth tokens; writes now go through a `0600` temp file and atomic rename ([#983](https://github.com/PrimeIntellect-ai/prime-agent/issues/983))
+- Fixed the legacy credential migration removing `oauth.json` and `settings.json` apiKeys before `auth.json` was safely written, which let a crash mid-migration destroy all credentials ([#983](https://github.com/PrimeIntellect-ai/prime-agent/issues/983))
 
 ## [0.7.1] - 2026-08-07
 
--- a/packages/coding-agent/src/core/auth-storage.ts
+++ b/packages/coding-agent/src/core/auth-storage.ts
@@ -15,8 +15,18 @@ import {
 	type OAuthProviderId,
 } from "@earendil-works/pi-ai";
 import { getOAuthApiKey, getOAuthProvider, getOAuthProviders } from "@earendil-works/pi-ai/oauth";
-import { chmodSync, existsSync, mkdirSync, readFileSync, writeFileSync } from "fs";
-import { dirname, join } from "path";
+import {
+	chmodSync,
+	closeSync,
+	existsSync,
+	mkdirSync,
+	openSync,
+	readFileSync,
+	renameSync,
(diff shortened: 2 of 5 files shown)


sethkarten:
Thank you for the report and proposed work. This root cause is now covered by maintainer-owned stacked PR #1160, authored independently from `upstream/main`.

We did not inspect or reuse this PR's diff, branch, commits, implementation code, or tests; its public description/comments were used only as a bug report. To keep one review surface, this PR is superseded by #1160 and is being closed.

PR #996, opened 2026-08-08
fix(ai): flush trailing SSE events and keep unparented text deltas

### LINKED ISSUE #995: Responses-family stream parsers silently drop the terminal event and unparented text deltas
Two silent data-loss bugs in the OpenAI Responses-family stream parsers in `packages/ai`.

## 1. Codex `parseSSE` drops a trailing unterminated event

`packages/ai/src/providers/openai-codex-responses.ts`, `parseSSE` (~535-586). Events are only extracted when a blank-line delimiter arrives inside the read loop. If a server closes the connection right after the final `data:` line without a trailing blank line, the terminal `response.completed` event is dropped and the final usage and stop reason are lost. The Anthropic provider's `iterateSseMessages` already handles this by draining the decoder and flushing the remaining buffer, the Codex one does not. CRLF line endings are also not normalized.

## 2. `output_text.delta` is silently dropped when `content_part.added` never arrives

`packages/ai/src/providers/openai-responses-shared.ts` (~376-392, same pattern for `refusal.delta` at ~393-409). The delta handler only appends when a matching content part already exists, and silently continues otherwise. Responses-compatible endpoints that are not fully spec-compliant (Azure, OpenRouter, various proxies) can stream `response.output_text.delta` without a preceding `response.content_part.added`, and the assistant text just vanishes with no error.

Fix for both: flush the decoder and any remaining buffered event after the read loop, normalize line endings, and lazily create the missing content part before appending a delta.

Adolanium:
Fixes #995.

## What was broken

Two silent data-loss bugs in the Responses-family stream parsers:

1. The Codex `parseSSE` only processed events delimited by a blank line inside the read loop. A server that closes the connection right after the final `data:` line (no trailing blank line) lost its terminal `response.completed` event, taking the final usage and stop reason with it. CRLF line endings were also not tolerated.
2. The shared Responses handler discarded `output_text.delta` and `refusal.delta` events when no `content_part.added` had created the content part first. Non-compliant endpoints (Azure, OpenRouter, proxies) do this, and the streamed text vanished without any error.

## The fix

- `parseSSE` drains the `TextDecoder` after the read loop, flushes any remaining buffered event, and normalizes CRLF before splitting. Same approach as the Anthropic provider's `iterateSseMessages`.
- The delta handlers lazily create the missing content part before appending, instead of dropping the delta.

## Testing

- New regression coverage: a Codex stream ending without a trailing blank line still delivers `response.completed` with text, stop reason, and usage; deltas without a preceding `content_part.added` still emit text.
- `npx tsx ../../node_modules/vitest/dist/cli.js --run` on the new/changed files plus the neighboring Responses and Codex suites: 48 passed, 4 skipped (live-key tests skip offline).
- `npm run check` clean.

> [!NOTE]
> ### Fix SSE parser to flush trailing events and handle unparented text deltas
> - The `parseSSE` generator in [openai-codex-responses.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/996/files#diff-69623a8c4e616b30339182296d1b69ed62a9d7fcce14b744ed6e638ecf357abc) now flushes any buffered data after the read loop ends, yielding the final SSE event even when the stream has no trailing blank line. CRLF line endings are also normalized to LF.
> - `processResponsesStream` in [openai-responses-shared.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/996/files#diff-75a92c85369ffe30444d6da4b398bb4fd5b852ff45a5eea02f95b3b2fa48df48) no longer drops `output_text.delta` and `refusal.delta` events that arrive without a preceding `response.content_part.added` event — it now creates the missing content part on demand and emits the delta normally.
> - Behavioral Change: streams that previously silently discarded the terminal SSE event or lost assistant text will now surface that content.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized b2c73f6.</sup>
> 
>

### DIFF
--- a/packages/ai/CHANGELOG.md
+++ b/packages/ai/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## [Unreleased]
 
+- Fixed the Codex SSE parser dropping the terminal event when a stream ends without a trailing blank line, losing final usage ([#995](https://github.com/PrimeIntellect-ai/prime-agent/issues/995))
+- Fixed Responses streams losing assistant text when a text delta arrives without a preceding content part event ([#995](https://github.com/PrimeIntellect-ai/prime-agent/issues/995))
+
 ## [0.7.1] - 2026-08-07
 
 ## [0.7.0] - 2026-08-05
--- a/packages/ai/src/providers/openai-codex-responses.ts
+++ b/packages/ai/src/providers/openai-codex-responses.ts
@@ -543,32 +543,25 @@ async function* parseSSE(response: Response): AsyncGenerator<Record<string, unkn
 		while (true) {
 			const { done, value } = await reader.read();
 			if (done) break;
-			buffer += decoder.decode(value, { stream: true });
+			buffer += decoder.decode(value, { stream: true }).replace(/\r\n/g, "\n");
 
 			let idx = buffer.indexOf("\n\n");
 			while (idx !== -1) {
 				const chunk = buffer.slice(0, idx);
 				buffer = buffer.slice(idx + 2);
 
-				const dataLines = chunk
-					.split("\n")
-					.filter((l) => l.startsWith("data:"))
-					.map((l) => l.slice(5).trim());
-				if (dataLines.length > 0) {
-					const data = dataLines.join("\n").trim();
-					if (data && data !== "[DONE]") {
-						try {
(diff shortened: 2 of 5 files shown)


kevinjosethomas:

PR #998, opened 2026-08-08
fix(ai): reprice Anthropic cache writes from final stream usage

### LINKED ISSUE #997: Anthropic provider reports stale cache-write cost when message_delta carries updated cache creation
In `packages/ai/src/providers/anthropic.ts`, the cache-write per-token rate is computed once from the `cache_creation` TTL breakdown in `message_start` (~550-556) and captured in a closure. The `message_delta` handler (~682-711) re-reads the latest token counts but keeps passing the stale rate into `calculateCost`.

When `message_delta` carries different cache-creation numbers than `message_start`, reported `cost.cacheWrite` and `cost.total` drift from actual billing. This happens behind proxies like LiteLLM that re-aggregate usage, and the 1h vs 5m rate gap under `cacheRetention: "long"` amplifies the error. The SDK's `MessageDeltaUsage` type does not declare the `cache_creation` breakdown (only the scalar), so the current code never reads it from `message_delta` at all, but proxies do include it at runtime.

Fix: reprice inside the `message_delta` handler from the latest `event.usage.cache_creation` breakdown when present, and fall back to the `message_start` rate otherwise.

Adolanium:
Fixes #997.

## What was broken

The Anthropic provider computes the cache-write per-token rate once from the `cache_creation` TTL breakdown in `message_start` and reuses that value for every later `calculateCost` call. When `message_delta` carries different cache-creation numbers, which proxies like LiteLLM produce when they re-aggregate usage, the reported `cost.cacheWrite` and `cost.total` drift from actual billing. The 1h vs 5m rate gap under `cacheRetention: "long"` makes the drift larger. The SDK type does not declare the breakdown on `MessageDeltaUsage`, but proxies include it at runtime and the raw SSE parser does not validate shapes.

## The fix

The `message_delta` handler reprices cache writes from the latest `event.usage.cache_creation` breakdown when it is present, and keeps the `message_start` rate when it is absent. No event shapes or public API changed.

## Testing

- New regression test in `test/anthropic-sse-parsing.test.ts`: `message_start` reports only 5m writes, `message_delta` reports only 1h writes under `cacheRetention: "long"`, and the final cost must use the 1h rate. Verified the test fails with the fix reverted.
- `npx tsx ../../node_modules/vitest/dist/cli.js --run` on the anthropic test files and neighbors (cache-retention, thinking, oauth, tool-name, copilot): 45 passed, 9 skipped (live-key suites), 0 failed.
- `npm run check` clean.

> [!NOTE]
> ### Fix Anthropic cache write repricing using final stream `cache_creation` breakdown
> Anthropic streams can return an updated `cache_creation` breakdown in `message_delta` events that differs from the initial `message_start` rate. Previously, `streamAnthropic` used only the `message_start` cache write cost, leading to stale pricing. Now, when a `cache_creation` breakdown is present in `message_delta`, the cache write cost is recomputed using `getAnthropicCacheWriteCost` with the latest TTL mapping before the final `calculateCost` call.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 4b7d7c3.</sup>
> 
>

### DIFF
--- a/packages/ai/CHANGELOG.md
+++ b/packages/ai/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## [Unreleased]
 
+- Fixed Anthropic cost reporting using a stale cache-write rate when the final stream usage carries updated cache-creation numbers ([#997](https://github.com/PrimeIntellect-ai/prime-agent/issues/997))
+
 ## [0.7.1] - 2026-08-07
 
 ## [0.7.0] - 2026-08-05
--- a/packages/ai/src/providers/anthropic.ts
+++ b/packages/ai/src/providers/anthropic.ts
@@ -6,7 +6,11 @@ import type {
 	MessageParam,
 	RawMessageStreamEvent,
 } from "@anthropic-ai/sdk/resources/messages.js";
-import { getAnthropicCacheWriteCost, hasStandardAnthropicCachePricing } from "../cache-pricing.js";
+import {
+	type AnthropicCacheCreationUsage,
+	getAnthropicCacheWriteCost,
+	hasStandardAnthropicCachePricing,
+} from "../cache-pricing.js";
 import { getEnvApiKey } from "../env-api-keys.js";
 import { calculateCost, clampThinkingLevel } from "../models.js";
 import type {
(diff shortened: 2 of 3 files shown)


kevinjosethomas:

PR #1001, opened 2026-08-08
fix(coding-agent): run OAuth refresh outside the auth.json file lock

### LINKED ISSUE #999: OAuth token refresh holds the auth.json file lock across the network call
`refreshOAuthTokenWithLock` in `packages/coding-agent/src/core/auth-storage.ts` (~764-808) awaits the OAuth provider's token refresh inside the `withLockAsync` callback. That refresh is an HTTP round-trip, and the `proper-lockfile` lock on `auth.json` (stale threshold 30s) is held for the whole call.

A slow or hanging provider then blocks every other process trying to read or write `auth.json` for up to 30s. The sync lock retry path gives up after roughly 200ms and throws, so concurrent processes can fail outright just because a refresh was in flight.

Fix: only hold the lock to read and re-check the stored credentials and later to persist the result. Run the network refresh with no lock held, and when re-acquiring, skip the write if the stored credentials changed in the meantime (another process refreshed first). In that case return the newer stored credentials and discard our result, which matters because refresh tokens rotate on use.

Adolanium:
Fixes #999.

## What was broken

`refreshOAuthTokenWithLock` awaited the provider's token refresh (an HTTP round-trip) inside the `withLockAsync` callback, holding the `auth.json` lock the whole time. A slow or hanging provider blocked every other process reading or writing `auth.json` for up to the 30s stale threshold, and the sync retry path gave up after about 200ms and threw.

## The fix

The lock is now held only for the read and the write:

1. Acquire, read the stored credentials, re-check expiry, release.
2. Refresh over the network with no lock held.
3. Re-acquire and persist only if the stored credentials still match the earlier snapshot. If another process refreshed first, the newer stored credentials win and our result is dropped unwritten, which is required because refresh tokens rotate on use.

Public API, `withLockAsync`, and the `onCompromised` behavior are unchanged.

## Testing

- New regression file `test/suite/regressions/999-oauth-refresh-lock.test.ts`: a refresh stub takes `lockfile.lockSync` on `auth.json` mid-refresh to prove the lock is free while the network call is in flight, and a fresher credential written during the refresh is returned instead of being overwritten.
- `npx tsx ../../node_modules/vitest/dist/cli.js --run` on the new file plus `test/auth-storage.test.ts` and `test/auth-flows.test.ts`: 55 passed, 2 failed. Both failures are pre-existing environment artifacts (a Windows `stat` mode check and a test that reads the machine's real `~/.prime/config.json`), in code paths this change does not touch.
- `npm run check` clean.

> [!NOTE]
> ### Fix OAuth token refresh in `AuthStorage` to release auth.json lock before network call
> - `refreshOAuthTokenWithLock` previously held the auth.json file lock during the provider's `refreshToken` network call, risking ~30s blocking on slow providers.
> - The method now reads and snapshots expired credentials, releases the lock, performs the refresh, then reacquires the lock to write results.
> - If another process updated the credentials concurrently, the write is skipped and the fresher stored credentials are returned instead.
> - A regression test in [999-oauth-refresh-lock.test.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/1001/files#diff-ad52fa710df5cf31664125d1967204195cd1570c4b90b97c6c185b295475d1b4) verifies both the lock-release behavior and the stale-write protection.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 846b2a1.</sup>
> 
>

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -5,6 +5,7 @@
 - Added privacy-safe pseudonymous product analytics for onboarding, command use, execution modes, run outcomes, TTFT, latency, usage, tools, retries, and compactions, with disclosure and opt-out controls ([ENG-4682](https://linear.app/primeintellect/issue/ENG-4682/add-privacy-safe-posthog-analytics-to-prime-agent)).
 - Changed sent agent messages in the IPython cell UI to show only the message text with a `╰─` gutter when expanded, matching received messages, and hid the raw `agent_message.send` receipt dictionary.
 - Fixed Homebrew installs attempting to self-update their versioned Cellar keg instead of directing users to `brew upgrade prime-agent` ([#844](https://github.com/PrimeIntellect-ai/prime-agent/issues/844))
+- Fixed the OAuth token refresh holding the `auth.json` file lock across the network call, blocking other processes for up to 30s on a slow provider ([#999](https://github.com/PrimeIntellect-ai/prime-agent/issues/999))
 
 ## [0.7.1] - 2026-08-07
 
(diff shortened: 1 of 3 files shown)


kevinjosethomas:

PR #1002, opened 2026-08-08
fix(windows): terminate orphaned process trees, not just their roots

### LINKED ISSUE #917: Windows: daemon recovery does not terminate unassigned orphan process trees
## Summary

On Windows, daemon recovery waits for journaled orphan processes to disappear but does not terminate process trees that were spawned outside the worker's kill-on-close Job Object. Autonomous child processes are tracked in the orphan journal, but they are not assigned to that Job Object.

After a worker crash, those processes can remain alive until they finish naturally. Recovery waits five seconds, then fails with `One or more orphaned process trees could not be terminated` instead of killing the stale process and replacing the worker.

## Expected behavior

Windows recovery should terminate each still-current orphan process tree with the existing `taskkill /F /T` path, wait for it to exit, and only fail recovery if the process remains alive after the timeout.

## Scope

This is separate from #841, which covers stale session-lease lock directories. It is related to the Windows process isolation work in #744; the review there identified the same gap for autonomous gates that are not assigned to the Job Object.

## Evidence

The Windows branch currently has a platform-specific `reapOrphanProcesses` path that only polls process identity. The Unix path calls `signalProcessTree`, but the Windows path never does. A crashed worker can therefore leave unassigned autonomous descendants behind.

noahsabaj:
## Problem

Windows has no process groups. `signalProcessGroupOrProcess` tries `process.kill(-pid)` first and falls back to `process.kill(pid)`; on Windows the first call always fails, so the fallback terminates exactly one process. Worker recovery inlined the same two-step kill for each journaled orphan, with the same result.

A crashed worker therefore leaves its detached descendants running — a Python kernel, an autonomous gate — until they finish on their own. They are recorded in the orphan journal, and recovery reads that journal, but the kill it issues cannot reach them.

## Fix

`taskkill /F /T` walks the real process tree. Both call sites now go through `signalProcessGroupOrProcess`, so recovery and teardown share one implementation instead of two copies that disagree by platform.

Unix is unchanged: it still signals the process group in a single call.

`windowsHide` on the taskkill spawn is not cosmetic. A daemon is spawned detached and owns no console, and a console child of a console-less parent gets a new console *with a visible window* that outlives it, so a teardown killing many processes would leave a desktop full of windows to close by hand.

## Why the test looks the way it does

Two details are load-bearing, and both are easy to get wrong:

- **The child is detached.** Windows puts an ordinary child in a job object owned by its parent, so it dies with the parent whether or not anything walks the tree. A test with a non-detached child passes with the fix reverted. Detached descendants are what a crashed worker actually leaves behind.
- **Windows only.** A detached child leaves its parent's process group on Unix too, so the group signal does not reach it there either. That is the existing Unix behavior and this change does not touch it.

Removing the Windows branch fails the test. I verified that in both directions rather than assuming it.

## Verification

- The new test passes on Windows 11 and fails with the branch reverted.
- `test/orphan-process-journal.test.ts` passes.
- `biome`, `tsgo --noEmit`, and the root `npm run check` pass.
- Note for reviewers: `test/daemon-supervisor-process.test.ts` fails 8 cases on Windows on this branch — and fails the same 8 on unmodified `main`, with "Supervisor exited before becoming ready". That is the un-ported Windows daemon state, not a regression from this change. I checked the baseline before writing this.

## Relationship to #744

A job object is the better containment primitive and I would rather see that land than this. It does not make the journal redundant: a job object only covers processes assigned to it, and this issue is precisely about the descendants that were not. A tree kill over the journal is the backstop for those, so the two compose.

Fixes #917.

---

> [!NOTE]
> ### Fix Windows orphaned process teardown to kill entire process trees, not just root PIDs
> - `signalProcessGroupOrProcess` in [`child-process.ts`](https://github.com/PrimeIntellect-ai/prime-agent/pull/1002/files#diff-a9f6b77c57772609afaa172c89bf3bfb5dd1e9d8ab775b567e40388a0f78fe96) now runs `taskkill /F /T /PID <pid>` on Windows to recursively terminate all descendants, falling back to `process.kill` if `taskkill` fails.
> - [`DaemonSupervisor`](https://github.com/PrimeIntellect-ai/prime-agent/pull/1002/files#diff-18f2d5ec391785bebc6ddf3c1bd3b795e4ed6e02d0aa3852f5bf247d61d012f8) worker teardown now routes orphan PIDs through `signalProcessGroupOrProcess` instead of calling `process.kill` directly, picking up tree-kill behavior.
> - A Windows-only integration test in [`child-process-tree-kill-windows.test.ts`](https://github.com/PrimeIntellect-ai/prime-agent/pull/1002/files#diff-a2834ffefa4687fab1835c616d3cf3754fb0fcbc2539fe0737f6529da57d4c50) verifies that both a detached parent and its detached child are killed.
> - Behavioral Change: on Windows, orphaned worker processes that previously survived teardown (detached descendants) will now be forcibly terminated.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 31435c2.</sup>
> 
>

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -12,6 +12,7 @@
 - Fixed Homebrew installs attempting to self-update their versioned Cellar keg instead of directing users to `brew upgrade prime-agent` ([#844](https://github.com/PrimeIntellect-ai/prime-agent/issues/844))
 - Fixed the agents view collapsing expanded subagent lists when returning from an opened agent ([ENG-5105](https://linear.app/primeintellect/issue/ENG-5105/keep-the-agents-view-state-persistent)).
 - Kept the subagent summary row visible and selectable while its list is expanded in the agents view, so pressing enter on it collapses the list again ([ENG-5105](https://linear.app/primeintellect/issue/ENG-5105/keep-the-agents-view-state-persistent)).
+- Fixed worker teardown on Windows leaving detached descendants running. Windows has no process groups, so the kill reached only the named process; it now walks the real process tree ([#917](https://github.com/PrimeIntellect-ai/prime-agent/issues/917)).
 
 ## [0.7.1] - 2026-08-07
 
--- a/packages/coding-agent/src/modes/daemon/daemon-supervisor.ts
+++ b/packages/coding-agent/src/modes/daemon/daemon-supervisor.ts
@@ -2897,16 +2897,10 @@ export class DaemonSupervisor {
 					if (!isOrphanProcessIdentityCurrent(orphan)) {
 						continue;
 					}
-					const { pid } = orphan;
-					try {
-						process.kill(-pid, "SIGKILL");
-					} catch {
-						try {
-							process.kill(pid, "SIGKILL");
-						} catch {
-							// The detached resource may already have exited.
-						}
-					}
+					// Tree kill: a detached resource can itself have children, and on
+					// Windows the inlined single-process fallback this replaces left
+					// every one of them running.
+					signalProcessGroupOrProcess(orphan.pid, "SIGKILL");
 				}
 				clearOrphanProcessJournal(orphanProcessJournalPath);
 			} catch (error) {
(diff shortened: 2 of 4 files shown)


kevinjosethomas:

PR #1012, opened 2026-08-08
fix(coding-agent): allow Luna children on empty Codex catalogs

### LINKED ISSUE #1011: RLM subagents cannot use GPT-5.6 Luna from a GPT-5.6 Sol parent session
## Summary

In a chat whose active model is `openai-codex/gpt-5.6-sol`, RLM children inherit Sol. There does not appear to be an accessible GPT-5.6 Luna selector for spawning a child from that session, so the only apparent workaround is to switch the entire parent chat to **GPT-5.6 Luna (max effort)** with `/model` and then spawn children without an explicit selector.

This makes Luna subagents effectively unavailable from a Sol chat and can also lead to accidentally launching Sol children when Luna was intended.

## Steps to reproduce

1. Start or use a parent chat running `openai-codex/gpt-5.6-sol`.
2. Attempt to spawn an RLM child intended to run GPT-5.6 Luna (max effort).
3. Observe that no usable explicit Luna model selector is available (or Luna cannot be selected for the child).
4. Spawn without an explicit selector.
5. Observe that the child inherits `openai-codex/gpt-5.6-sol`.

## Actual behavior

The children run Sol. In my session, two children intended for Luna inherited Sol and had to be cancelled immediately.

## Expected behavior

One of the following should work:

- GPT-5.6 Luna (including max effort) is exposed as a valid explicit model selector for `rlm(...)`, regardless of the parent model; or
- the UI/API provides a documented way to choose a different child model without switching the parent chat.

At minimum, the selected/resolved child model should be clearly confirmed before work begins so unintended-model runs can be avoided.

## Workaround

Switch the parent chat with `/model` to **GPT-5.6 Luna (max effort)**, then spawn children without an explicit selector so they inherit Luna.

## Environment / evidence

- Parent session reported: `openai-codex/gpt-5.6-sol`
- Affected feature: RLM child/subagent spawning and model inheritance
- Two unintended Sol children were cancelled immediately

cskwork:
## Summary

Allow a Sol parent to discover and explicitly spawn a GPT-5.6 Luna RLM child without pinning Prime Agent to an unrelated Codex CLI version.

The `/codex/models` endpoint currently returns an empty successful catalog when Prime Agent sends its own `client_version`. An empty catalog is not proof that the authenticated account has no usable models: direct Luna requests work with the same credentials.

This change:

- treats an empty successful Codex catalog as inconclusive and keeps locally known Codex models executable;
- keeps non-empty account catalogs authoritative, preserving their filtering;
- keeps authentication preflight unchanged;
- adds an end-to-end regression proving a Sol parent discovers, admits, and retains a Luna child when discovery returns an empty catalog.

Fixes #1011. This overlaps the empty-catalog policy in #696/#717 but adds the exact Sol-to-Luna RLM regression.

## Verification

- Focused tests: 13/13 passed
- `npm run check`: passed
- Real patched-source run against the live service:
  - Sol parent
  - `rlm.find_models("luna")` returned `openai-codex/gpt-5.6-luna`
  - spawn handle reported `openai-codex/gpt-5.6-luna`
  - real child returned `NO_PIN_LUNA_OK`

Please merge this fix (or fold its Sol-to-Luna regression into #717) so explicit Luna children work without maintaining a hard-coded external CLI version.

> [!NOTE]
> ### Fix Luna child model discovery when Codex model catalog returns empty
> - In [`model-registry.ts`](https://github.com/PrimeIntellect-ai/prime-agent/pull/1012/files#diff-2fd8c4a2672b4ce27e73d494479df612c8b30ebac3dda5a160e2d861c612f453), introduces a `filterByCatalog` helper that treats an empty catalog as non-authoritative, leaving all local openai-codex models available instead of filtering them all out.
> - Previously, an empty catalog response (e.g. due to a backend gate) would cause all models to be filtered, breaking Luna subagent discovery.
> - Adds a regression test in [`1011-luna-subagent-selector.test.ts`](https://github.com/PrimeIntellect-ai/prime-agent/pull/1012/files#diff-79289e73948c6f6fa67cbe21af83b508f8817af73ea140b692ef8c2ed2c2ad78) that stubs an empty catalog fetch and asserts Luna is discoverable and executes successfully.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized b72703c.</sup>
> 
>

### DIFF
--- a/packages/coding-agent/src/core/model-registry.ts
+++ b/packages/coding-agent/src/core/model-registry.ts
@@ -978,9 +978,16 @@ export class ModelRegistry {
 			return availableModels.filter((model) => model.provider !== "openai-codex");
 		}
 		const authFingerprint = createHash("sha256").update(auth.apiKey).digest("hex");
+		// An empty successful catalog can be caused by the backend's client-version
+		// gate and does not prove that an authenticated account has no models. Keep
+		// local Codex models executable and let the actual request be authoritative.
+		const filterByCatalog = (ids: Set<string>): Model<Api>[] =>
+			ids.size === 0
+				? availableModels
+				: availableModels.filter((model) => model.provider !== "openai-codex" || ids.has(model.id));
 		const cached = this.openAICodexModelsCache;
 		if (cached?.authFingerprint === authFingerprint && Date.now() - cached.refreshedAt < 300_000) {
-			return availableModels.filter((model) => model.provider !== "openai-codex" || cached.modelIds.has(model.id));
+			return filterByCatalog(cached.modelIds);
 		}
 
 		const accountId = readOpenAICodexAccountId(auth.apiKey);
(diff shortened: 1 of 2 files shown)


sethkarten:
Closed as superseded by #1153, an independent implementation written from `upstream/main` without importing outside commits, source code, or test files. #1153 resolves the Sol-to-Luna child-selection path with its own implementation in the existing regression suite.

PR #1038, opened 2026-08-09
fix(coding-agent): bound host requests and drain lost-idle executions

ProDrifterDK:
## Summary

- add a finite, configurable deadline to every `rlm.host_request()` on both sides of the kernel↔host bridge
- close and detach Python Futures/Comms, abort host handlers on timeout, execution cancellation, Comm close, and kernel disposal, and ignore late results safely
- distinguish top-level requests from detached `asyncio` tasks so cancelling a cell does not cancel legitimate background host requests
- propagate request cancellation through RLM admission, child publication/message delivery, goal/compact/refine/heartbeat, and MCP handler boundaries
- drain the serialized Jupyter execution queue when the matching shell `execute_reply` arrives but IOPub drops its terminal `status: idle`

### Host-request bridge

- The kernel-side bridge previously opened a Comm and awaited a Future with no deadline.
- The host dispatched handlers with no per-request timeout and kept in-flight bookkeeping until the handler settled or the kernel shut down.
- In the observed `goal.complete()` incident, durable goal state remained `active`; that proves the synchronous completion mutation did not finish, but it does not identify whether the Comm open, handler dispatch, or reply path stalled.
- The bridge fix therefore applies independent 120-second deadlines at the narrow Python↔host Comm seam.

### Lost-IOPub-idle execution queue

A separate incident never entered the host-request bridge. An internal auto-snapshot reached the kernel and wrote its state artifacts, but the host never observed its terminal IOPub `idle`. Because `KernelManager` serializes every user and internal cell behind the prior execution promise, the later user cell remained queued before Jupyter assigned it a request ID. The 120-second `host_request()` deadline could not apply.

The fix starts the shell receive pump after readiness probing and accepts only an `execute_reply` whose `parent_header.msg_id` exactly matches the active execution. IOPub `idle` remains authoritative. If it does not arrive, the manager waits a bounded two-second grace for trailing IOPub output, then finishes from the matching shell reply. Reply/timer state is bound to the exact execution and cleared on normal idle, abort, rejection, shutdown, and disposal, so stale replies or timers cannot settle a newer cell.

## Behavior

`host_request(..., timeout_ms=...)` defaults to 120 seconds and accepts `1..3_600_000` ms. `PRIME_AGENT_HOST_REQUEST_TIMEOUT_MS` configures the kernel default. Timeouts raise `TimeoutError` with the request type, elapsed deadline, and an explicit warning that the remote outcome may be unknown; callers should inspect state before retrying.

Existing one-argument host handlers remain source-compatible. New handlers may consume `HostRequestContext.signal` for cooperative cancellation. The kernel and namespace remain live after timeout or cancellation; restart is not part of this change.

For ordinary executions, IOPub `idle` behavior is unchanged. The shell fallback applies only when an exact matching `execute_reply` arrives without `idle`; it preserves output received during the grace and propagates terminal shell abort/error status.

## Regression seam

`kernel-execute-reply-fallback.test.ts` covers exact request-ID matching, normal idle/output preservation, the two-second fallback, stale reply/timer isolation, abort/dispose cleanup, and shell error propagation.

Its real-kernel regression exercises the hidden predecessor directly: a user cell schedules an internal auto-snapshot; the test waits for the snapshot manifest and dill artifact, drops only that snapshot request's exact IOPub `idle`, observes its exact shell reply, and requires two subsequent user cells to complete. It also instruments the Comm entry point and proves `hostRequestCount === 0` and no in-flight host request, demonstrating that this failure is outside the host-request timeout layer.

## Alternatives rejected

- An outer tool timeout alone leaves the top-level await and serial kernel queue active; #865 handles that broader recovery layer separately.
- Restarting the kernel discards user namespace and background work and is unnecessary when shell completion proves the cell has terminated.
- A bare elapsed-time race could release the serial queue while the kernel is still executing and permit overlapping cells.
- The prior experimental recovery commits mixed tool timeout, execute-reply recovery, and restart escalation, making them too broad to cherry-pick.

## Validation

Lost-idle candidate `7b0da32b1802be442521fd71493d0a1847c6baea`:

- `kernel-execute-reply-fallback.test.ts` — 6 unit tests passed, 1 real-kernel test skipped in normal mode
- `kernel-execute-reply-fallback.test.ts --tagsFilter kernel-heavy` — 1 real-kernel auto-snapshot regression passed, 6 unit tests filtered out
- `kernel-host-request-timeout.test.ts` — 7 passed
- `kernel-abort.test.ts` — 5 passed
- live `kernel-goal-skill.test.ts` — 6 passed
- live `kernel-state-roundtrip.test.ts` — 6 passed
- `kernel-startup.test.ts` — 1 passed
- `npm run check` — Biome checked 902 files with no fixes, `tsgo --noEmit`, installer render, and browser smoke passed
- `git diff --check` — passed
- independent exact-SHA review — PASS with no actionable findings; verified request identity, grace/timer isolation, all cleanup paths, output preservation, internal snapshot queue drainage, zero host Comm participation, and compatibility

Earlier host-request validation retained on this branch:

- `prime-agent-runtime/.venv/bin/python -m unittest discover -s prime-agent-runtime/test` — 73 passed
- 200 relevant coding-agent tests passed across recursion, host skills, MCP, kernel abort/bootstrap, and provisioner coverage
- independent adversarial review found two lifecycle gaps in its first pass; both were fixed and the second pass approved with no Critical or Important findings

## Compatibility

This update adds no daemon command, event, response-shape, protocol-version, dependency, or configuration requirement. The lost-idle change is internal to kernel execution lifecycle handling. Existing host-request timeout and abort semantics remain intact.

Addresses the host-request hang and lost-idle queue failure described in #848 and complements, rather than duplicates, #865.

> [!NOTE]
> ### Bound host requests with deadlines and drain lost-idle executions in KernelManager
> - `rlm.host_request` in [rlm/__init__.py](https://github.com/PrimeIntellect-ai/prime-agent/pull/1038/files#diff-8f05bb028ac34c9ff421f45f913f35be2a99010ef924da344d7a427869c7ead3) now enforces a configurable deadline (default 120 s, overridable via `PRIME_AGENT_HOST_REQUEST_TIMEOUT_MS`) and raises `TimeoutError` on local timeout; the Comm is always closed and the callback unregistered on exit.
> - [KernelManager](https://github.com/PrimeIntellect-ai/prime-agent/pull/1038/files#diff-14dd18f6926ed763ba0b55625302fb733f4e4209bb611d3051820048732149ac) now pumps the shell channel and uses an `execute_reply` fallback: if IOPub idle is not received within 2 s of `execute_reply`, execution is force-finished, preventing the serial queue from stalling when Jupyter drops the idle event.
> - Host request handlers across `KernelManager`, `AgentSession`, `McpManager`, and `RlmRuntime` now receive an `AbortSignal` via a `HostRequestContext` and abort cooperatively on timeout, Comm closure, or parent execution abort.
> - Execution-owned host requests (flagged via `_prime_agent_execution_owned`) are linked to the triggering cell's `AbortSignal`, so aborting a cell also cancels its in-flight host requests.
> - Risk: existing unary host request handlers continue to work, but handlers that were previously unbounded may now timeout after 120 s by default.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 9489fda.</sup>
> 
>

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [Unreleased]
 
 - Fixed URLs not opening on click in fullscreen mode on terminals such as Ghostty; clicking a link in the transcript, dock, or overlays now opens it in the browser.
+- Fixed IPython host bridge requests hanging indefinitely when a request, handler, or response is lost, with a configurable 120-second deadline and safe cleanup ([#848](https://github.com/PrimeIntellect-ai/prime-agent/issues/848)).
 
 ## [0.7.2] - 2026-08-11
 
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Fixed URLs not opening on click in fullscreen mode on terminals such as Ghostty; clicking a link in the transcript, dock, or overlays now opens it in the browser.
 - Fixed IPython host bridge requests hanging indefinitely when a request, handler, or response is lost, with a configurable 120-second deadline and safe cleanup ([#848](https://github.com/PrimeIntellect-ai/prime-agent/issues/848)).
+- Fixed completed IPython cells blocking the serial execution queue when Jupyter drops the matching IOPub idle event ([#848](https://github.com/PrimeIntellect-ai/prime-agent/issues/848)).
 
 ## [0.7.2] - 2026-08-11
 
--- a/packages/coding-agent/README.md
+++ b/packages/coding-agent/README.md
@@ -685,6 +685,7 @@ prime-agent --thinking high "Solve this complex problem"
 | `PRIME_AGENT_TRACES_API_KEY` | Prime API key used only for opt-in trace sharing |
 | `PRIME_AGENT_TRACES_BASE_URL` | Override the Prime Agent trace upload API base URL |
 | `PRIME_AGENT_KERNEL_PYTHON` | Use an existing Python environment with `ipykernel` instead of auto-bootstrapping `~/.prime/agent/kernel-venv` |
+| `PRIME_AGENT_HOST_REQUEST_TIMEOUT_MS` | Set the IPython kernel↔host request deadline in milliseconds (default: `120000`; range: `1`–`3600000`) |
 | `VISUAL`, `EDITOR` | External editor for Ctrl+G |
 
 The remaining `PI_*` variables in this table are compatibility names still read by the current runtime. They do not change the application name, command, or default `~/.prime/agent` configuration path.
--- a/packages/coding-agent/package.json
+++ b/packages/coding-agent/package.json
@@ -44,7 +44,7 @@
 		"postinstall": "node postinstall.cjs",
 		"prepublishOnly": "npm run clean && npm run build",
 		"bundle": "node scripts/bundle.mjs",
-		"test:kernel": "vitest --run --no-file-parallelism --tagsFilter kernel-heavy test/acp-kernel-features.test.ts test/acp-cold-cli.test.ts test/kernel-goal-skill.test.ts test/kernel-state-roundtrip.test.ts"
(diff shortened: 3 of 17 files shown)


kevinjosethomas:

Issue #1023, opened 2026-08-08
npm run check fails on Windows when POSIX sh is unavailable


emindagg:
## Problem

`npm run check` fails on a clean native Windows checkout when no POSIX `sh` executable is on `PATH`.

The repository documents Windows as a supported platform and `AGENTS.md` requires contributors to run `npm run check` after code changes, but `scripts/check-installer-render.mjs` unconditionally runs its generated harness through `spawnSync("sh", ...)`.

## Reproduction

On Windows 11 with Node 24.18.0 and npm 11.16.0:

```powershell
git clone https://github.com/PrimeIntellect-ai/prime-agent
cd prime-agent
npm ci
npm run check
```

Biome and type checking pass, then the installer check reports every render case as:

```text
harness exited with unknown
undefinedundefined
```

A direct probe shows the actual failure:

```text
spawnSync sh ENOENT
```

## Root cause

`runCase` checks only `result.status` and concatenates `result.stderr` / `result.stdout`. When spawning fails before a child exists, `status` is null, both output fields are undefined, and the actionable `result.error` is discarded.

The installer itself targets macOS/Linux, so a Windows contributor without Git Bash cannot execute this render harness even though the rest of the required repository check is portable.

## Expected behavior

- Native Windows contributors can run the required `npm run check` without installing an undocumented POSIX shell.
- The installer render check continues to execute fully on supported installer hosts and CI.
- Unexpected child-spawn failures report `result.error` instead of `unknown / undefinedundefined`.

## Proposed fix

Detect an unavailable POSIX shell before running the render cases. On Windows only, skip this installer-specific check with an explicit message when `sh` is absent. On other platforms, or for any other spawn failure, fail with the actual spawn error.

kevinjosethomas:
Hi, thanks for taking the time to report this! Since open sourcing Prime Agent, we’ve received far more Issues than we can reliably triage and maintain. Rather than leave a large backlog that we cannot meaningfully work through, we’re closing the current Issue queue and moving to a discussion-first process.

[...]

Issue #946, opened 2026-08-08
Add native Windows and macOS CI coverage


fettpl:
## Problem and impact

Platform-specific process, path, native module, terminal, and installer behavior is not exercised before merge.

## Reproduction

Current workflow matrix has no Windows/macOS runner, so a platform-only regression cannot affect required CI.

## Evidence

[`.github/workflows/ci.yml:17-116`](https://github.com/PrimeIntellect-ai/prime-agent/blob/a18809e0/.github/workflows/ci.yml#L17-L116) and nightly stress run exclusively on Ubuntu; repository code has Windows-specific terminal/process/venv branches and multiple open platform bugs (#660, #665-#669, #719, #735, #841, #869).

## Acceptance criteria

- Native smoke/test lanes on Windows and macOS for supported Node versions.
- Platform-specific tests no longer skipped merely due runner absence.
- Aggregate/release gates include them or a documented required subset.

## Solution design

Start with focused install, CLI launch, path/session lease, TUI input, and kernel bootstrap smoke lanes; expand based on runtime cost.

## Verification plan

Seed known platform regression fixtures and ensure the corresponding native lane catches them.

## Audit metadata

- Audit ID: `P2-12`
- Priority: `P2`
- Effort: `M`
- Implementation risk: `MED`
- Confidence: `HIGH`
- Suggested repository labels: `enhancement`, `github_actions`
- Package scope: `pkg:tui`, `pkg:coding-agent`
- Audited revision: `fa900d5a` (local merge tree)
- Evidence revision: [`a18809e0`](https://github.com/PrimeIntellect-ai/prime-agent/commit/a18809e0)
- Audit date: 2026-08-08

The issue author could not apply labels. Apply the suggested repository labels above and the package-scope labels once they are provisioned.

The audited tree differs from the linked upstream evidence revision only by an unrelated local design document.

SandroHub013:
A native Windows lane should include daemon lifecycle and Herdr smoke coverage, not only the kernel/installer paths. Named pipes cannot be enumerated like Unix socket files, so `status`/`doctor`/`shutdown` miss a live default daemon when it has no worker descriptor unless discovery explicitly probes the default pipe. Windows force-stop also needs process-tree semantics (`taskkill /T`), rather than only `process.kill(pid)`.

I added local regression coverage for default-pipe inclusion and updated the shutdown path to use Windows tree termination. These would be useful acceptance cases for making the Windows lane required.

[...]

PR #1053, opened 2026-08-09
fix(windows): bring Windows to parity with Linux and macOS

### LINKED ISSUE #660: Windows: kernel bootstrap uses venv bin/python, so the IPython kernel never starts and each retry wipes the venv

### LINKED ISSUE #665: Windows: no supported install path (installer and docs cover macOS/Linux only)

### LINKED ISSUE #666: Windows: fsync on a directory handle fails with EPERM

### LINKED ISSUE #667: Windows: a stale session lease blocks session recovery

### LINKED ISSUE #668: Windows: detached child processes open visible console windows

### LINKED ISSUE #869: Windows: daemon-worker shell commands spawn visible focus-stealing console windows (worker has no console to inherit)
## Summary

On Windows, a daemon-backed session worker runs without a console. Every built-in shell/exec tool invocation (each `gh`, `git`, etc.) therefore causes Windows to allocate a **new visible console window** — which flashes and can steal foreground focus from whatever the user is doing. During an active session this happens many times per minute, making the host machine effectively unusable for other work (in our case it disrupted the user's foreground application to the point of crashing it).

## Environment

- prime-agent 0.7.0 (stable), Windows 11, Node v24.13.0, Git Bash present
- Resident daemon worker (launched detached; also reproduces when the resume client was started hidden — the worker itself is detached from any console either way)

## Evidence

External watcher (polling Win32_Process) during normal operation catches spawns like:

```
PID 26980  PPID <worker node PID>
Cmd: C:\WINDOWS\system32\cmd.exe /c "gh pr view 383 --json mergeable,mergeStateStatus,statusCheckRollup,headRefOid"
```

Each such cmd.exe gets a fresh visible conhost window. Exit codes are all 0 — nothing is functionally wrong; it is purely a windowing/UX defect, invisible to the agent itself.

## Expected

Tool subprocesses spawned by a console-less worker should never create visible windows: spawn with CREATE_NO_WINDOW (for one-off commands), or give the worker a persistent hidden console that children inherit (CREATE_NEW_CONSOLE + SW_HIDE on a long-lived shell host), or use windowless pipes throughout.

## Notes

Related prior reports from this setup: #740 (RLM model discovery under subscription OAuth), #841 (stale session-lease directories on Windows). Same underlying theme: daemon-worker behavior on Windows diverges from the POSIX assumptions. Happy to test a fix build.

### LINKED ISSUE #917: Windows: daemon recovery does not terminate unassigned orphan process trees

### LINKED ISSUE #1047: Windows: %%bash cells execute inside WSL instead of the resolved shell
## Summary

On Windows, `%%bash` cells in the IPython tool execute inside WSL rather than the shell Prime Agent resolved for its Bash tool. The cell then runs in a different operating system: the working directory arrives as `/mnt/c/...`, Windows paths the agent computed are invalid, and environment variables the kernel set do not cross the boundary.

IPython's `ScriptMagics` resolves `bash` from `PATH`. On a Windows machine with WSL installed, the first `bash.exe` on `PATH` is usually `C:\Windows\System32\bash.exe`, the WSL launcher.

## Reproduction

Windows 11, Node 24, Git for Windows installed, WSL installed.

```
prime-agent --print 'run a %%bash cell containing: uname -a; echo PWD=$PWD'
```

Observed, reduced to the kernel level:

```python
import os
os.environ['NO_COLOR'] = '1'
from IPython.core.interactiveshell import InteractiveShell
sh = InteractiveShell.instance()
sh.run_cell('%%bash\nuname -a; echo PWD=$PWD; printf "[%s]" "$NO_COLOR"')
```

```
Linux PCMarcoB 6.18.33.2-microsoft-standard-WSL2 ... x86_64 GNU/Linux
PWD=/mnt/c/Users/<user>/<project>
[]
```

`NO_COLOR` is empty because WSL does not inherit Windows environment variables unless they are listed in `WSLENV`. A direct `subprocess.run([shutil.which('bash')], ...)` from the same kernel prints `[1]`, so the kernel environment is correct; only the magic's shell choice is wrong.

## Evidence

- `packages/coding-agent/src/core/tools/ipython.ts` rewrote a bare `%%bash` cell to `%%script <shell>` only when the user had explicitly configured `shellPath`. With no setting, the cell was left for IPython to resolve.
- `packages/coding-agent/src/utils/shell.ts` `findBashOnPath()` returned the first usable `bash.exe` from `where`, so the WSL launcher could also win for the Bash tool itself when Git Bash is not under `%ProgramFiles%`.
- The `%%script` argument was quoted for POSIX `shlex`. IPython splits magic arguments with `arg_split(..., posix=not sys.platform.startswith("win"))`, and the Windows mode does not strip single quotes, so a configured `shellPath` reached the magic as `'C:\...\bash.exe'` and failed with `Couldn't find program`.
- Autonomous gates ran with `shell: true`, which is `cmd.exe` on Windows, while every other shell surface in the agent is bash.

## Expected behavior

Every shell surface on Windows uses the same resolved bash:

- Bare `%%bash` cells are pinned to the shell the Bash tool resolves.
- `C:\Windows\System32\bash.exe` is ranked last, used only when nothing else is available.
- `%%script` arguments are quoted for the splitter IPython actually uses on the platform.
- Autonomous gate commands run under that same bash, so a gate written once behaves the same everywhere.

### LINKED ISSUE #1048: Windows: owned session worker skips its graceful shutdown and misses a killed frontend
## Summary

On Windows the owned session worker never runs its graceful shutdown, and it does not notice when its frontend is killed.

`installOwnedSessionWorkerOwnerWatch()` ends its termination path with `process.kill(process.pid, "SIGTERM")`. On Unix that delivers a signal and the process runs its own `SIGTERM` listeners. Windows has no signal delivery: Node maps a self-directed `SIGTERM` to `TerminateProcess`, so the process dies immediately and every handler registered for shutdown is skipped.

## Reproduction

```js
process.once("SIGTERM", () => { console.log("handler ran"); process.exit(0); });
setTimeout(() => { console.log("no handler after 500ms"); process.exit(3); }, 500);
process.kill(process.pid, "SIGTERM");
```

On Linux this prints `handler ran`. On Windows it prints nothing and exits 1.

Consequence in the worker: anything the session flushes or releases on `SIGTERM` is skipped whenever the owner watch fires.

## Evidence

- `packages/coding-agent/src/cli/owned-session-worker.ts`, `installOwnedSessionWorkerOwnerWatch()`: the force timer is armed and then `process.kill(process.pid, "SIGTERM")` is issued.
- The same function relies solely on `process.once("disconnect", ...)` to learn that the frontend is gone. On Windows the closed IPC channel is not reliably surfaced when the frontend is killed outright.

## Expected behavior

- Requesting one's own shutdown runs the registered `SIGTERM` listeners on every platform, and exits directly only when nothing is listening.
- The worker detects a killed frontend on Windows as well, for example by polling the parent pid as a backstop for the missing `disconnect`.

### LINKED ISSUE #1049: Kernel dispose() returns before the kernel exits, leaving its working directory locked on Windows
## Summary

`KernelManager.dispose()` kills the kernel and returns without waiting for it to exit. Unix unlinks files that are still open, so nothing notices. Windows keeps the kernel's working directory locked until the process is really gone, so anything the caller deletes after an awaited `dispose()` fails with `EPERM`.

## Reproduction

Windows 11, Node 24.

```ts
const provisioner = new IpythonKernelProvisioner(tempDir, {});
const manager = await provisioner.ensure();
await manager.execute("print(1)");
await provisioner.dispose();
rmSync(tempDir, { recursive: true, force: true });
```

```
EPERM, Permission denied: \\?\C:\Users\<user>\AppData\Local\Temp\<dir>
```

`dispose()` returns in about 2 ms; the directory becomes removable roughly 100 ms later. Retrying the delete succeeds, which confirms the handle is simply still closing.

## Evidence

- `packages/coding-agent/src/core/kernel/index.ts`: `dispose()` awaits in-flight host requests and then calls `cleanupResources()`, which signals the child and removes the temp directory in the same synchronous pass. Nothing waits for the child's `exit`.
- The kernel is spawned with `cwd` set to the session directory, which is what Windows holds open.

Related but separate: `fs.rmSync`'s own `maxRetries` does not cover this. The native implementation surfaces the directory-level `EPERM` without retrying, so passing `maxRetries` has no effect and the retry loop has to be written explicitly.

## Expected behavior

An awaited `dispose()` outlives the kill: the kernel process has exited and its working directory is removable by the time the promise resolves.

### LINKED ISSUE #1050: Windows: --daemon-socket with a filesystem path cannot bind or connect
## Summary

On Windows a daemon socket given as a filesystem path cannot be bound or connected. `prime-agent --daemon-socket C:\path\to\daemon.sock` fails with `EACCES`, because Windows has no Unix domain sockets and `net` will not bind a filesystem path.

Only the default socket works, because `defaultDaemonSocketPath()` returns a named pipe on Windows. Any explicit path, and every worker socket derived from one, is unusable.

## Reproduction

```powershell
prime-agent --daemon-socket C:\Users\<user>\AppData\Local\Temp\d.sock --mode daemon
```

```
Error: listen EACCES: permission denied C:\Users\<user>\AppData\Local\Temp\d.sock
```

The same happens for a client connecting to that path.

## Evidence

- `packages/coding-agent/src/modes/daemon/daemon-socket.ts`: `defaultDaemonSocketPath()` special-cases Windows to `\\.\pipe\prime-agent-daemon`, but nothing translates a caller-supplied path.
- `daemon-mode.ts`, `daemon-supervisor.ts`, `daemon-client.ts`, and `daemon-worker-client.ts` pass the socket path straight to `server.listen()` / `createConnection()`.

A second, related defect in the same area: `resolveDaemonUpdateRestartSocketPath()` in `packages/coding-agent/src/cli/daemon-update-restart.ts` skips `resolve()` entirely on Windows to avoid mangling a `\\.\pipe\...` endpoint. That also leaves a *relative* filesystem socket unresolved, and the update coordinator changes its working directory before using it, so it then looks for the socket in the wrong place.

## Expected behavior

A socket path keeps working as the daemon's identity everywhere — in `status`, logs, worker descriptors, and ownership records — while the transport uses whatever the platform supports. On Windows a filesystem path maps onto a deterministic named pipe, and a relative path is resolved before any working-directory change.

### LINKED ISSUE #1051: ~/.agents/skills is scoped as project when the working directory is under the home directory
## Summary

## Reproduction

```
prime-agent --print '/skills'
```

## Evidence

```ts
let dir = resolvedStartDir;
while (true) {
    const parent = dirname(dir);
    if (parent === dir) break;
    dir = parent;
}
```

## Expected behavior

### LINKED ISSUE #1052: Windows: uv discovery ignores PATHEXT, so a uv.cmd shim is never found
## Summary

Kernel bootstrap only looks for `uv` and `uv.exe` on Windows, so a `uv.cmd` or `uv.bat` shim — what scoop and pipx install — is never found. The bootstrap then reports uv as missing and offers to install it again, even though uv is on `PATH` and runnable from any shell.

## Reproduction

Windows 11 with uv installed through a shim rather than as a bare executable, so that `where uv` prints a `.cmd`:

```
prime-agent --print 'run print(1) in ipython'
```

```
uv is required to set up the Python kernel. Install uv yourself: powershell.exe ... irm https://astral.sh/uv/install.ps1 | iex
```

## Evidence

`packages/coding-agent/src/core/kernel/bootstrap.ts`, `findExecutable()`:

```ts
const candidates = process.platform === "win32" ? [name, `${name}.exe`] : [name];
```

`PATHEXT` is not consulted. A bare `uv` with no extension is also accepted here, but it is only spawnable when the file really is a PE image.

Spawning the shim once found is a second step: Node refuses to execute a `.cmd` or `.bat` without a shell, and a shell spawn on Windows hands the command line to `cmd.exe` verbatim without quoting anything — so a venv path containing a space, such as a profile named `C:\Users\Ada Lovelace`, would split into two arguments.

## Expected behavior

`uv` is discovered under every extension in `PATHEXT`, and running a shim quotes its arguments so a path containing a space survives.

SandroHub013:
## Summary

Brings Windows to parity with Linux and macOS: the full `packages/coding-agent` suite now passes natively on Windows, and the remaining platform-only defects in the daemon, kernel, worker, and shell paths are fixed rather than skipped.

Baseline on a clean Windows 11 checkout was 130 failing tests across 50 files. It is now 0.

| suite | before | after |
| --- | --- | --- |
| coding-agent | 130 failed / 50 files | 0 failed, 3965 passed, 92 skipped |
| tui | 3 failed | 0 failed, 752 passed |
| agent | — | 0 failed, 69 passed |
| ai | 3 failed | 0 failed (excluding cases that need a local Ollama server or provider credentials) |

## Product fixes

- **`%%bash` cells ran inside WSL.** IPython resolves `bash` from `PATH`, and on Windows that is usually the WSL launcher, so cells executed in a different operating system: the working directory arrived as `/mnt/c/...` and the kernel's environment did not cross the boundary. Bare `%%bash` cells are now pinned to the shell the Bash tool resolves, `C:\Windows\System32\bash.exe` is ranked last, and `%%script` arguments are quoted for the splitter IPython actually uses on the platform. Autonomous gates move to the same bash instead of `cmd.exe`. Fixes #1047.
- **Session leases never detected a conflict.** Claiming a lease renames a directory onto the target; Unix reports EEXIST, Windows EPERM. Only the Unix codes were treated as "already held", so the conflict escaped as a raw EPERM, `SessionAlreadyActiveError` was never raised, and two agents could believe they owned the same session. The supervisor launch lock had the same defect. Fixes #667.
- **Filesystem daemon socket paths could not bind.** `--daemon-socket <path>` failed with EACCES. Socket paths stay the daemon's identity everywhere and are mapped onto a deterministic named pipe at the transport boundary only. A relative path is now resolved before the update coordinator changes working directory. Fixes #1050.
- **The owned worker skipped its graceful shutdown.** `process.kill(process.pid, "SIGTERM")` is `TerminateProcess` on Windows: the process dies without running its own SIGTERM listeners. It also did not notice a killed frontend, since the closed IPC channel is not reliably surfaced there. Fixes #1048.
- **`KernelManager.dispose()` returned before the kernel exited**, leaving its working directory locked. Fixes #1049.
- **Orphan reaping skipped Windows entirely**, so a worker's descendants survived shutdown. Both platforms now go through `taskkill /T`. Fixes #917.
- **The command recovery journal fsynced its parent directory** after an atomic rename, a POSIX-only idiom that throws EPERM on Windows. NTFS journals the rename itself. Fixes #666.
- **`uv` discovery ignored `PATHEXT`**, so a `uv.cmd` shim was never found; running one now quotes its arguments, since a shell spawn hands the command line to `cmd.exe` verbatim. Fixes #1052.
- **`~/.agents/skills` was re-scoped from user to project** whenever the working directory sits under the home directory, which is the normal case on Windows. Fixes #1051.
- Kernel bootstrap resolves the venv interpreter at `Scripts\python.exe`. Fixes #660.
- Spawned children are hidden, so no console window flashes. Fixes #668, fixes #869.
- Smaller path fixes: `~/` expansion used string concatenation, producing `C:\Users\x/foo`; `getHomeDir()` preferred `$HOME`, which Git Bash exports as a POSIX path Windows APIs cannot resolve; the TUI render debug log wrote to a hardcoded `/tmp/tui`.

## Install and CI

- `install.ps1`, the Windows counterpart of `install.sh`: same preflight, release-channel resolution, SHA-256 verification, and `npm install -g` with the same bootstrap environment. `scripts/check-installer-ps1.mjs` keeps its release placeholders honest and parses the rendered script on Windows. `build-binaries.yml` publishes it alongside the shell installer. Fixes #665.
- A `windows-latest` test leg in CI covering `tui` and the three `coding-agent` shards. This does not cover macOS, so #946 stays open.
- `docs/windows.md` rewritten: install, the bash requirement, optional components, and the platform differences that remain.

## Test changes

Most test changes are fixtures that encoded POSIX assumptions rather than behaviour: `/tmp` literals that resolve to `C:\tmp`, paths joined with a forward slash, `existsSync(socketPath)` used as proof a daemon bound (a named pipe has no filesystem presence), `npx` spawned without a shell, and ESM specifiers passed as absolute Windows paths instead of `file://` URLs.

Where a case genuinely cannot run on Windows it is skipped with the reason inline, and the platform-specific logic is covered directly instead:

- Suites whose stub `python`, `uv`, or tool binaries are POSIX shell scripts. Windows cannot spawn one, and Node rejects a `.cmd` shim without a shell. `kernel-platform.test.ts` covers the venv layout and installer command; the real-kernel suites cover the rest.
- Cases needing file symlinks, which require `SeCreateSymbolicLinkPrivilege` and have no junction equivalent. Directory links use junctions and still run.
- POSIX mode-bit assertions, and the unwritable-install-path case: Windows ignores the read-only attribute on directories for write access.

Process-heavy budgets were widened for Windows, where process creation costs several times what it does on Unix.

## Known limitations

- `0600`/`0700` modes on config and credential files are no-ops on Windows. Protecting `~/.prime/agent` needs NTFS ACLs; documented in `docs/windows.md` rather than fixed here.
- The IPython fork server stays Linux-only, as it already was for macOS. This affects kernel start latency, not behaviour.
- #1023 (`npm run check` needs a POSIX `sh` for `check-installer-render.mjs`) is untouched.

## Verification

`npm run check` passes on every commit in this branch. The full `packages/coding-agent` suite, `tui`, and `agent` were run natively on Windows 11 with Node 24 and Git for Windows. The built CLI was smoke-tested end to end: an RPC `get_state` round-trip creates a session and answers, and `status` / `doctor` discover a live daemon on `\\.\pipe\prime-agent-daemon`.

> [!NOTE]
> ### Add Windows support to bring the coding agent to parity with Linux and macOS
> - Adds a PowerShell installer ([install.ps1](https://github.com/PrimeIntellect-ai/prime-agent/pull/1053/files#diff-50d51a233d8ff4886061fc5cf1a03642de1f64b3625903e4f876e8cd2a3f5806)) and Windows CI job that gates publishing on passing Windows tests
> - Introduces `daemonSocketEndpoint` in [daemon-socket.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/1053/files#diff-59c4ea71700afcf7d09d14c5d30e2c73fd0c71b28284e8758cd8fbd4b2337dbc) to map filesystem socket paths to Windows named pipes for all daemon listen/connect sites
> - Adds `signalProcessGroupOrProcess` and `requestSelfShutdown` in [child-process.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/1053/files#diff-a9f6b77c57772609afaa172c89bf3bfb5dd1e9d8ab775b567e40388a0f78fe96) to replace direct `process.kill` calls with `taskkill /T`-based process-tree termination on Windows
> - Adds [durable-fs.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/1053/files#diff-44eb30e6cb3f878e9fd2dd09d45f14a593fec962700fb398aad62dcc92a903f9) with `removePathSync`/`removePath` helpers that retry on transient Windows EPERM/EBUSY errors, replacing bare `rmSync` throughout the codebase
> - Fixes kernel bootstrap ([bootstrap.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/1053/files#diff-e87b9a04ab9b4ac02e8dd33af2f2f58320d8b7f6be6853acf4a3e05a34623cae)) to use `Scripts\python.exe` on Windows, install `uv` via PowerShell, and run `.cmd`/`.bat` shims through a shell with proper quoting
> - Adds `windowsHide: true` to all child process spawns to suppress console window flashes on Windows
> - Adds test helpers (`removeTempDirSync`, `symlinkDirSync`, `canCreateFileSymlinks`) in [temp-fs.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/1053/files#diff-bb8b0d1f02a3490b895e1a0b9e624d4d70899c541c4d39d42945f04f82380228) and gates symlink-dependent and POSIX-only tests with conditional skips on Windows
> - Risk: daemon communication switches from Unix sockets to named pipes on Windows, which changes the observable socket identity for tooling that inspects socket files
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 2b3c0a8.</sup>
> 
>


sajjade123:
Tested this branch on Windows 11, Node 22.19.0, Git Bash. Confirmed it fixes the exact failure I was hitting on main (commit a18809e): daemon worker binds its pipe, logs "listening", then times out waiting for worker_auth — every single run, regardless of Node version (also reproduced on v26.1.0), AV exclusions, elevation, or --dist vs tsx mode. Full repro trail available if useful. This branch launches cleanly and responds correctly to a real prompt.

rexcardan:
Validated this branch on a second clean Windows 11 machine (Node v24.12.0, npm 11.6.2, Git for Windows, uv 0.10.4), merged onto current main (a18809e0) since the branch has gone stale against it.

Merge: two conflicts, both insertion-point collisions with post-cut main commits — `packages/coding-agent/CHANGELOG.md` (0.7.1 promotion vs the Windows entries) and `packages/coding-agent/test/config.test.ts` (Homebrew fixture from #846 vs the platform-aware npm-root fixture). Both resolve cleanly keeping both sides.

Results after the merge:

- `npm run check` passes (needs Git's bin dir on PATH for `sh`, i.e. #1023, as documented).
- tui: 737 passed. agent: 69 passed. coding-agent shards: 1388 / 1511 / 1110 passed. Process suite: 8 passed. Kernel suite: 14 passed. 0 failures.
- Daemon binds `\\.\pipe\...` and answers `daemon_hello` (protocol 7); RPC `get_state` round-trips without credentials; interactive TUI launches, accepts input, and restores the terminal on exit.

Five small stabilizations were needed on top of the merge, all in the spirit of changes already in this branch:

- `test/telemetry.test.ts`, `test/session-manager-flush.test.ts`: POSIX mode-bit assertions gated off Windows (content assertions kept on all platforms).
- `test/acp-cold-cli.test.ts`, `test/daemon-supervisor-process.test.ts`: teardown moved to the async `removeTempDir` helper with a 30 s Windows budget.
- `src/modes/agent-connection/daemon-agent-connection.ts`: the owned-session completion request during dispose bounded with the existing `OWNED_SESSION_DISPOSE_RECONNECT_WAIT_MS` race, mirroring the reconnect wait a few lines above; without it, supervisor-replacement teardown can hang on Windows.

Merged branch with those fixes, in case it is useful for rebasing: https://github.com/rexcardan/prime-agent/tree/windows-native (ca394982 merge, 51d626be stabilizations).

kevinjosethomas:

PR #1055, opened 2026-08-09
fix: prevent self-amplifying compaction failure loop (#900)

### LINKED ISSUE #900: Compaction can permanently self-amplify from durable retry debris and unbounded summary input
## Summary

A long-running 0.7.0 root session can enter a self-amplifying, permanently unresponsive compaction loop:

1. provider rejects the normal turn with `context_length_exceeded`;
2. automatic compaction submits an oversized one-shot summary request and fails;
3. overflow recovery retries the same oversized request and fails;
4. each failure persists more retry/output diagnostic entries;
5. the next scheduled or agent-message prompt repeats the cycle.

The inbound message is accepted, so callers/bridges see successful admission, but the session never reaches a reply.

## Observed incident (sanitized)

Provider/model: `openai-codex/gpt-5.6-sol`  
Prime Agent: 0.7.0  
Installed model metadata: 272k context (this is **not** a claim that the backend's true hard limit is exactly 272k; provider usage semantics/metadata appear inconsistent)

- append-only session journal: 46.7 MB / 30,327 entries;
- 16 prior successful compactions;
- 930 persisted `compaction_outcome` entries, 925 caused by context-limit errors;
- 385 persisted `assistant` messages with `stopReason: "error"`;
- filtered pre-compaction estimate: 402,135 tokens;
- unbounded history summarization input: 1,440 messages / 1,098,845 serialized chars;
- first conclusive one-shot failure was about 469k estimated tokens / 1.88M serialized chars;
- exact provider error: `Your input exceeds the context window of this model. Please adjust your input and try again.`

The Slack/agent bridge and message gate were healthy. This was an internal session/compaction failure.

### 1. Retry cleanup is live-state-only

`_handleRetryableError()` removes the failed assistant message from `agent.state.messages`, but it has already been appended to JSONL. `_performCompaction()` rebuilds via `sessionManager.getBranch()`, so the failed partial output returns during replay/compaction.

This preserves useful audit evidence, but durable retry debris needs an explicit model-context exclusion policy.

### 2. Summary input is unbounded

Both `generateSummary()` and `generateTurnPrefixSummary()` serialize all supplied messages into a single provider request. The provider cannot compact input that already exceeds its request context. This applies to both normal history and split-turn prefixes, including one individually oversized message.

### 3. Failure telemetry becomes more model input

`_persistCompactionOutcome()` appends every failure to the journal and live messages. Repeated diagnostics grow the next compaction request. They should remain durable telemetry but not be replayed into model context; live state should at most coalesce to the latest outcome.

### 4. Direct accepted agent messages skip the needed preflight

`acceptAgentMessagePrompt()` sets `skipPrePromptWork: true`; `_turnExecutionPolicy("directPrompt")` consequently sets `preTurnCompaction: "skip"`. An idle session receiving an accepted agent message therefore calls the main provider first, overflows, and only then begins recovery compaction. Admission can remain fast while execution still performs `afterModelSelection` pre-turn compaction.

### 5. Stale usage can suppress threshold compaction

`_getThresholdContextTokens()`, `_thresholdCompactionNeeded()`, and `_checkCompaction()` return early when the latest provider usage predates the newest compaction. A large post-compaction user/custom-message tail can therefore avoid preflight compaction. A raw estimate of current live messages is a safe fallback across this boundary; the old timestamp guard is still useful only for deciding whether an assistant error itself predates the compaction.

## Expected behavior

- Preserve append-only JSONL evidence.
- Exclude failed assistant retries and compaction-outcome diagnostics from model-facing replay, cut-point accounting, and compaction extraction without filtering the branch topology array itself.
- Hard-bound **every** summary request using model context metadata, output budget, safety margin, rolling-summary size, and a conservative chars/token estimate.
- Split on message boundaries; explicitly truncate one individually oversized message; recompute the budget after each sequential rolling summary.
- Compact directly accepted agent messages before the main provider call.
- Use raw current-state estimates when provider usage is absent/stale.
- Add a circuit breaker/backoff so repeated failed compactions do not create hundreds of persisted failure turns.

## Validation of a local fix

Using a no-tools disposable copy of the exact poisoned journal:

- failed assistants and outcomes were absent from rebuilt model context while remaining in JSONL;
- 1,098,845 history chars were processed as three dynamically bounded sequential requests;
- the accepted agent message entered `Auto-compacting...` **before** any new provider error;
- the 17th compaction succeeded: 402,135 -> about 38k displayed tokens;
- the requested reply returned exactly `OK`;
- historical assistant-error count did not increase;
- worst-case recovery completed in about five minutes.

Offline tests also covered a single 900k-character message (bounded to the calculated 555,495-character request budget) and stale usage with small vs. 800k-character post-compaction tails.

## Related work

- #832 requests provider-native compaction hooks.
- #733 adds Codex provider-native compaction but is currently conflicted and does not by itself address durable retry/outcome replay.
- #344 contains useful sequential chunking groundwork, but its fixed 80k target unnecessarily increased worst-case recovery latency; budget should be derived per request.
- #43 introduced the stale usage/compaction boundary guard involved here.

I can provide a synthetic sanitized fixture or break the local fix into source-level PRs if maintainers confirm the preferred durable-entry exclusion semantics.

kartikmengane09-oss:
## Summary

Fixes the self-amplifying compaction failure loop described in #900.

### Problem

Long-running sessions can exceed the model context window. The initial provider call fails, recovery compaction is then attempted with an oversized summarization request, and the compaction failure is persisted and replayed into subsequent model context. This causes repeated failures and progressively larger context until the session becomes unresponsive.

### Changes

- Prevent internal failed/recovery state from contaminating model-facing context while preserving the durable JSONL audit trail.
- Bound compaction/summarization requests against the available model context budget.
- Process oversized history through bounded message-aligned summarization rather than a single oversized request.
- Ensure necessary pre-compaction checks are performed before provider calls, including direct agent-message paths.
- Handle stale provider usage with a current-state token estimate where necessary.
- Prevent uncontrolled repeated compaction failures and preserve valid state when compaction fails.

### Validation

Relevant compaction and session-context regression tests are included/run as part of the change.

Closes #900

> [!NOTE]
> ### Fix self-amplifying compaction failure loop by adding context budget enforcement
> - [`compaction.ts`](https://github.com/PrimeIntellect-ai/prime-agent/pull/1055/files#diff-ad4b3f096b732189266e944cde055f0e15abda3e3e2244b029909b9a89cee11f): `generateSummary` now splits oversized conversations into budget-sized chunks using a chars/4 token estimate, summarizes each chunk independently, then aggregates into a final summary with a second-pass if needed.
> - `compact` validates the produced summary against the model's context window after generation and throws if the result would exceed the safe limit, preventing an oversized compacted state from being committed.
> - [`session-manager.ts`](https://github.com/PrimeIntellect-ai/prime-agent/pull/1055/files#diff-e514ecbe1ca80031cc52600ee3863d993a99616f116355f3fec33c92b532d2cf): `buildSessionContext` now filters out assistant messages with `stopReason: "error"` and `compaction_outcome` custom messages from the model context to avoid replaying failed attempts.
> - Behavioral Change: `compact` now throws instead of silently succeeding when the compacted result is too large for the context window.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 6f1ef3e.</sup>
> 
>

### DIFF
--- a/packages/coding-agent/src/core/compaction/compaction.ts
+++ b/packages/coding-agent/src/core/compaction/compaction.ts
@@ -549,11 +549,213 @@ export function buildSummarizationPrompt(customInstructions?: string, previousSu
 	return `${basePrompt}\n\n${KERNEL_PERSIST_SUMMARY_NOTE}`;
 }
 
+// ============================================================================
+// Input budget and chunking helpers
+// ============================================================================
+
+/**
+ * Conservative overhead budget (chars/4) for the system prompt, prompt wrapper
+ * tags, previous-summary block, and the instruction text appended by
+ * buildSummarizationPrompt(). 2000 tokens = ~8 kB of boilerplate.
+ */
+const SUMMARIZATION_PROMPT_OVERHEAD_TOKENS = 2000;
+
+/**
+ * Compute the safe number of input tokens available for conversation text in a
+ * single summarization request, given the model's context window.
+ *
+ * Returns 0 when the model does not advertise a context window, meaning the
+ * caller should skip the budget check.
+ */
+function safeInputBudget(model: Model<any>, reserveTokens: number): number {
+	const contextWindow = model.contextWindow ?? 0;
+	if (contextWindow <= 0) return 0;
+	// Leave room for: summary output (reserveTokens), prompt overhead, and a
+	// 10% safety margin on what remains.
+	const raw = contextWindow - reserveTokens - SUMMARIZATION_PROMPT_OVERHEAD_TOKENS;
+	return Math.max(0, Math.floor(raw * 0.9));
+}
+
+/**
+ * Estimate the token count of a serialized block of text using the same
+ * chars/4 heuristic used by estimateTokens().
+ */
+function estimateTextTokens(text: string): number {
+	return Math.ceil(text.length / 4);
+}
+
+/**
+ * Split `messages` into groups where each group's serialized text fits within
+ * `budgetTokens`. Splits only at AgentMessage boundaries.
+ *
+ * Returns a single chunk containing all messages when the budget is 0 (unknown
+ * context window) or when everything fits in one chunk.
+ */
+function chunkMessagesByBudget(messages: AgentMessage[], budgetTokens: number): AgentMessage[][] {
+	if (budgetTokens <= 0 || messages.length === 0) return [messages];
+
+	const chunks: AgentMessage[][] = [];
+	let current: AgentMessage[] = [];
+	let currentTokens = 0;
+
+	for (const message of messages) {
+		const llm = convertToLlm([message]);
+		const text = serializeConversation(llm);
+		const tokens = estimateTextTokens(text);
+
+		if (current.length > 0 && currentTokens + tokens > budgetTokens) {
+			// Current chunk would overflow — flush it and start a new one.
+			chunks.push(current);
+			current = [];
+			currentTokens = 0;
+		}
+		current.push(message);
+		currentTokens += tokens;
+	}
+
+	if (current.length > 0) chunks.push(current);
+	return chunks;
+}
+
 /**
  * Generate a summary of the conversation using the LLM.
  * If previousSummary is provided, uses the update prompt to merge.
+ *
+ * When the serialized conversation exceeds the model's safe input budget the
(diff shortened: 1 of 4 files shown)


sethkarten:

PR #1071, opened 2026-08-09
 fix(coding-agent): bind extension registerShortcut handlers in daemon-backed  clients

### LINKED ISSUE #1057: Daemon-backed interactive clients do not bind extension registerShortcut handlers
## Summary

Extension commands load and execute in Prime Agent 0.7.1's default daemon-backed TUI, but shortcuts registered by the same extension are neither listed nor invoked.

This blocks one-key accessibility controls for extensions such as local speech-to-text. A second extension cannot work around it through the public API because normal package extensions run in the daemon worker, while shortcut handlers are bound only from the interactive client's local `ExtensionRunner`.

## Minimal reproduction

```ts
export default function example(pi) {
  pi.registerCommand("shortcut-test", {
    description: "Command path",
    handler: async (_args, ctx) => ctx.ui.notify("command fired"),
  });
  pi.registerShortcut("f9", {
    description: "Shortcut path",
    handler: async (ctx) => ctx.ui.notify("shortcut fired"),
  });
}
```

1. Install the extension and run `/reload` in the normal interactive client.
2. `/shortcut-test` appears in autocomplete and displays the notification.
3. F9 does nothing and the extension shortcut is absent from the shortcut guide.
4. Rebinding it to another valid, conflict-free Prime key has the same result.

## Observed boundary

The daemon command catalog crosses `AgentConnection` and executes handlers in the worker. The daemon-backed `InteractiveMode` is created without local extension binding, while `setupExtensionShortcuts()` requires the local extension runner. There does not appear to be corresponding shortcut metadata/execution support on the daemon connection boundary.

## Expected behavior

A daemon session should expose registered shortcut metadata to attached interactive clients, and the client should route a matched key back to the owning worker handler. Conflict diagnostics and the shortcut guide should behave consistently with in-process extensions.

## Current workaround

Extensions can register short slash commands, and terminal-specific send-text macros can invoke those commands, but neither is a portable replacement for the maintained `registerShortcut` API.

Example accessibility extension and reproduction: https://github.com/3xjn/prime-agent-voice

kritikagarg:
**Problem**

  When prime-agent runs daemon-backed (TUI client connected to a running daemon
  worker), keyboard shortcuts registered by extensions via registerShortcut()
  silently do nothing. The root cause is that bindLocalSessionExtensions is
  hardcoded to false in daemon mode, so setupExtensionShortcuts() is never
  called and defaultEditor.onExtensionShortcut stays undefined. Key presses
  fall straight through to built-in handlers.

  Extensions do load and register their shortcuts — but inside the daemon
  worker process. The TUI client has no access to those in-process closures.

  Fixes #1057.

  **Solution**

  A daemon-protocol round-trip that keeps handler execution in the daemon while
  giving the TUI client enough information to proxy key presses:

  1. After session.bindExtensions() in the daemon, collect the registered
  shortcut key IDs from the extension runner and broadcast them to attached
  clients as a new extension_shortcut_list outbound message.
  2. The TUI client receives extension_shortcut_list and installs a proxy
  onExtensionShortcut handler that matches those keys and forwards presses to
  the daemon via the new extension_shortcut_trigger inbound command.
  3. The daemon handles extension_shortcut_trigger by looking up the handler in
  the extension runner and invoking it with a proper ExtensionContext via
  createContext().

  This approach keeps shortcut handlers executing in the daemon (where the
  extension is loaded), works correctly with multiple attached TUI clients, and
  requires no double-loading of extensions.

  **Changes**

  - daemon-protocol.ts — add extension_shortcut_list outbound message and
  extension_shortcut_trigger inbound command
  - daemon-extension-binding.ts — broadcast shortcut key IDs after
  bindExtensions completes
  - daemon-mode.ts — handle extension_shortcut_trigger by invoking the
  registered handler via extensionRunner.createContext()
  - agent-connection/types.ts — add extension_shortcut_list to
  AgentConnectionEvent; add triggerExtensionShortcut to AgentConnection
  interface
  - daemon-agent-connection.ts — relay extension_shortcut_list as a connection
  event; implement triggerExtensionShortcut
  - in-process-agent-connection.ts — stub triggerExtensionShortcut (direct mode
  handles shortcuts locally)
  - interactive-mode.ts — handle extension_shortcut_list to install the proxy
  handler; clear stale proxy on session rebind

  **Testing**

  - Type-check (tsgo --noEmit): passes
  - Linter (biome check): passes
  - Pre-commit hooks (biome + tsgo + installer render + browser smoke): all
  pass

> [!NOTE]
> ### Fix extension `registerShortcut` handlers to bind in daemon-backed clients
> - Adds `triggerExtensionShortcut` to the `AgentConnection` interface and implements it in `DaemonAgentConnection` by sending an `extension_shortcut_trigger` command to the daemon, which looks up and invokes the matching shortcut handler.
> - After extensions bind, the daemon broadcasts registered shortcut keys via a new `extension_shortcut_list` outbound event; `InteractiveMode` receives this and proxies matching key presses to the daemon.
> - Clears stale shortcut proxies on `defaultEditor.onExtensionShortcut` when rebinding to a new daemon-backed session.
> - Fixes `collectAncestorAgentsSkillDirs` to stop traversal at the user home directory, preventing `~/.agents/skills` from being included in project-scoped skill dirs.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 212b922.</sup>
> 
>

### DIFF
--- a/packages/coding-agent/src/core/package-manager.ts
+++ b/packages/coding-agent/src/core/package-manager.ts
@@ -434,9 +434,15 @@ function collectAncestorAgentsSkillDirs(startDir: string): string[] {
 	const skillDirs: string[] = [];
 	const resolvedStartDir = resolve(startDir);
 	const gitRepoRoot = findGitRepoRoot(resolvedStartDir);
+	const homeDir = resolve(getHomeDir());
 
 	let dir = resolvedStartDir;
 	while (true) {
+		// Stop before reaching the home directory: ~/.agents/skills is user-scope,
+		// not project-scope, and is added separately by the caller.
+		if (dir === homeDir) {
+			break;
+		}
 		skillDirs.push(join(dir, ".agents", "skills"));
 		if (gitRepoRoot && dir === gitRepoRoot) {
 			break;
@@ -2179,9 +2185,7 @@ export class DefaultPackageManager implements PackageManager {
 			themes: join(projectBaseDir, "themes"),
 		};
 		const userAgentsSkillsDir = join(getHomeDir(), ".agents", "skills");
-		const projectAgentsSkillDirs = collectAncestorAgentsSkillDirs(this.cwd).filter(
-			(dir) => resolve(dir) !== resolve(userAgentsSkillsDir),
-		);
+		const projectAgentsSkillDirs = collectAncestorAgentsSkillDirs(this.cwd);
 
 		const addResources = (
 			resourceType: ResourceType,
--- a/packages/coding-agent/src/modes/agent-connection/daemon-agent-connection.ts
+++ b/packages/coding-agent/src/modes/agent-connection/daemon-agent-connection.ts
@@ -731,6 +731,14 @@ export class DaemonAgentConnection implements AgentConnection {
 		});
 	}
 
+	async triggerExtensionShortcut(key: string): Promise<void> {
+		await this.requestOk({
+			type: "extension_shortcut_trigger",
+			activeSessionId: this.activeSessionId,
+			key,
+		});
+	}
+
 	async prompt(message: string, options?: AgentConnectionPromptOptions): Promise<void> {
 		await this.promptWithAdmissionCancellation("prompt", message, options);
 	}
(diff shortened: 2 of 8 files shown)


kevinjosethomas:

PR #1081, opened 2026-08-09
fix(coding-agent): keep shutdown admission through delays

Kenmege:
## Summary

- coalesce concurrent shutdown-admission renewals so the timer and explicit checks cannot contend with each other
- allow the same live, token-matching owner to renew after expiry when no successor claimed the lease
- make the ENG-4603 macOS fixture use a symlink so Homebrew Node keeps its `@rpath/libnode` resolution

## Root cause

`shutdown --force` can spend longer than the 5-second lease inside synchronous daemon discovery before the event loop gets its first refresh. In the reproduced failure, the lease was created at `16:10:19.904`, expired at `16:10:24.904`, and first renewed at `16:10:29.401`; the token, PID, process start ID, and live process identity were still unchanged. The old code permanently fenced out its own shutdown. Concurrent explicit checks could also bypass the timer's in-flight promise and create avoidable registry-lock contention.

Renewal still runs under the supervisor registry guard. If a successor claims an expired lease first, it removes/replaces the record, and the predecessor fails the token fence. The existing successor-reclaim regression continues to cover that case.

## Validation

- `npx tsx ../../node_modules/vitest/dist/cli.js --run test/suite/regressions/4603-worker-recovery.test.ts -t "renews an expired shutdown admission"`
- `npx tsx ../../node_modules/vitest/dist/cli.js --run test/suite/regressions/4603-worker-recovery.test.ts -t "coalesces concurrent shutdown admission renewals"`
- `npx tsx ../../node_modules/vitest/dist/cli.js --run test/suite/regressions/4603-worker-recovery.test.ts -t "shutdown --force removes hidden supervisors"`
- `npm run check`
- `npm run build` in `packages/coding-agent`

The full ENG-4603 file also ran all six test bodies; under sustained local load its final shutdown case intermittently timed out in the test harness's post-test `waitForExit` cleanup after the isolated end-to-end case passed. No residual Prime Agent process remained.

> [!NOTE]
> ### Fix daemon shutdown admission loss during delayed or concurrent lease refresh
> - Expired but otherwise valid shutdown admissions are now renewed instead of treated as lost; the `expiresAt` check is removed from the loss condition in `renewOnce()`.
> - Concurrent calls to `assertOrRenew()` in [`daemon-supervisor-ownership.ts`](https://github.com/PrimeIntellect-ai/prime-agent/pull/1081/files#diff-6f9ef6ce6a9d2bc51c61fe94d59310efffd26d8f7b3966f5355606b2847d06a5) now coalesce into a single in-flight renewal promise instead of racing.
> - `release()` now awaits any in-flight renewal with errors ignored, so cleanup is not blocked or aborted by a renewing promise.
> - New regression tests cover expired admission renewal and concurrent renewal coalescing.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 3392ae9.</sup>
> 
>

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [Unreleased]
 
+- Fixed long or concurrent daemon shutdown checks aborting cleanup when their lease refresh was delayed or self-contended.
 - Added privacy-safe pseudonymous product analytics for onboarding, command use, execution modes, run outcomes, TTFT, latency, usage, tools, retries, and compactions, with disclosure and opt-out controls ([ENG-4682](https://linear.app/primeintellect/issue/ENG-4682/add-privacy-safe-posthog-analytics-to-prime-agent)).
 - Changed sent agent messages in the IPython cell UI to show only the message text with a `╰─` gutter when expanded, matching received messages, and hid the raw `agent_message.send` receipt dictionary.
 - Fixed Homebrew installs attempting to self-update their versioned Cellar keg instead of directing users to `brew upgrade prime-agent` ([#844](https://github.com/PrimeIntellect-ai/prime-agent/issues/844))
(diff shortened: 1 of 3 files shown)


kevinjosethomas:

PR #1084, opened 2026-08-09
fix(runtime): preserve MCP tool inputSchema from SDK snake_case fields

### LINKED ISSUE #1073: McpIntegration.list_tools() always returns an empty inputSchema (camelCase getattr vs snake_case SDK field)

## Cause

```python
```

```python
from mcp.types import Tool
list(Tool.model_fields)
'inputSchema'
```

## Reproduction

```python
t["inputSchema"]
{}
```

```json
{"type": "object",
```

## Impact

## Why the tests pass

```python
Tool = type("Tool", (), {})

t.inputSchema = schema
```

## Suggested fix

```python
```

```python
```

sdshah09:
## Summary
- Fixed `McpIntegration.list_tools()` always returning empty `inputSchema` by reading the MCP SDK's snake_case `input_schema` attribute (camelCase is only the wire alias).
- Applied the same snake_case-first lookup for `is_error` / `structured_content` in `_parse_result`.
- Updated unit fakes and added a regression test so SDK-shaped attributes are exercised.

Fixes #1073

## Test plan
- [x] `cd prime-agent-runtime && PYTHONPATH=src uv run python -m unittest discover -s test -p 'test_mcp_base.py' -v`
- [x] Confirmed `_sdk_field` returns the schema from a real `mcp.types.Tool`

> [!NOTE]
> ### Fix MCP tool `inputSchema` population by reading SDK snake_case `input_schema` field
> The MCP Python SDK exposes fields like `input_schema`, `is_error`, and `structured_content` as snake_case attributes, but the code was reading camelCase names (`inputSchema`, `isError`, `structuredContent`), causing tool schemas and structured results to be empty or ignored.
>
> - Adds `_sdk_field()` helper in [mcp_base.py](https://github.com/PrimeIntellect-ai/prime-agent/pull/1084/files#diff-3e0a705745bc6d6096fecf5672e6173e25ee363911b2fc0ff065fa59f0117524) that prefers the snake_case attribute and falls back to the camelCase alias, so both naming conventions are handled.
> - Updates `list_tools()` to use `_sdk_field(t, "input_schema", "inputSchema")` so tool entries include the correct schema.
> - Updates `_parse_result()` to read `is_error` and `structured_content` via the same helper.
> - New tests verify snake_case fields are read correctly and camelCase aliases still work as fallbacks.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 33320d9.</sup>
> 
>

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -5,6 +5,7 @@
 - Added privacy-safe pseudonymous product analytics for onboarding, command use, execution modes, run outcomes, TTFT, latency, usage, tools, retries, and compactions, with disclosure and opt-out controls ([ENG-4682](https://linear.app/primeintellect/issue/ENG-4682/add-privacy-safe-posthog-analytics-to-prime-agent)).
 - Changed sent agent messages in the IPython cell UI to show only the message text with a `╰─` gutter when expanded, matching received messages, and hid the raw `agent_message.send` receipt dictionary.
 - Fixed Homebrew installs attempting to self-update their versioned Cellar keg instead of directing users to `brew upgrade prime-agent` ([#844](https://github.com/PrimeIntellect-ai/prime-agent/issues/844))
+- Fixed MCP `list_tools()` returning empty `inputSchema` values by reading the SDK's snake_case `input_schema` field ([#1073](https://github.com/PrimeIntellect-ai/prime-agent/issues/1073))
 
 ## [0.7.1] - 2026-08-07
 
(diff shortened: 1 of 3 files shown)


sethkarten:
Thank you for the report and proposed work. This root cause is now covered by maintainer-owned stacked PR #1164, authored independently from `upstream/main`.

We did not inspect or reuse this PR's diff, branch, commits, implementation code, or tests; its public description/comments were used only as a bug report. To keep one review surface, this PR is superseded by #1164 and is being closed.

PR #1085, opened 2026-08-09
fix(coding-agent): emit the Bedrock provider into the CLI bundle

pmclaugh:
# Bedrock provider is missing from the CLI bundle

## Summary

The published `bin` (`dist/bundle/cli.js`) cannot use any Bedrock model. The first
request fails with:

```
module.streamSimple is not a function
```

`dist/` built unbundled works fine, and the compiled Bun binary works, so this only
affects npm/node installs -- which is every `npm i -g prime-agent` user.

## Root cause

`packages/ai/src/providers/register-builtins.ts` loads nine providers lazily. Eight use
a literal specifier that esbuild resolves and emits as a chunk:

```ts
import("./anthropic.js")
import("./openai-completions.js")
```

Bedrock deliberately goes through a variable specifier so browser bundles do not pull
in the AWS SDK:

```ts
const importNodeOnlyProvider = (specifier: string): Promise<unknown> => import(specifier);
...
importNodeOnlyProvider("./amazon-bedrock.js")
```

esbuild cannot resolve a variable specifier, so it emits no chunk and no warning. At
runtime the import still resolves relative to the emitted chunk directory, to a file
that was never written.

Confirmed by building at `a18809e0`: chunks appear for all eight literal-specifier
providers and `dist/bundle/amazon-bedrock.js` is absent.

The indirection itself is correct and load-bearing -- removing it (making the specifier
literal) makes `npm run check` fail, because the AWS SDK then enters the browser smoke
build and `@smithy/node-http-handler` cannot resolve `node:http`. So the provider must
stay invisible to static analysis, and the bundler has to be told about it explicitly.

## Fix

Declare the provider as a second bundle entry named after the specifier the runtime
uses, so the file lands exactly where the import looks. `splitting: true` keeps it a
lazily loaded chunk, so the AWS SDK is still only read when a Bedrock model is used.

The entry imports only pi-ai's public `./bedrock-provider` export and re-exports the two
names `loadBedrockProviderModule()` destructures. `packages/ai` is untouched.

Because esbuild cannot verify this emission, `scripts/bundle.mjs` now asserts the
artifact exists and fails the build otherwise -- the original failure mode was a silent
omission that only surfaced at runtime for one provider.

## Verification

- `dist/bundle/amazon-bedrock.js` emitted (679 KB) and exports `streamBedrock` and
  `streamSimpleBedrock` as functions.
- Still lazy: the AWS SDK appears only in that chunk, which is not in the static import
  closure of `cli.js` (2 files, 4 KB eager).
- `cli.js` unchanged in name, size and `0755` mode.
- `npm run check` passes, including the browser smoke build.
- Removing the entry point reproduces the bug: the build throws
  `bundle is missing .../amazon-bedrock.js`, and the new test fails.

## Tests

`packages/coding-agent/test/bundle-bedrock-entry.test.ts` asserts the coupling between
the runtime specifier in `register-builtins.ts` and the entry names in `bundle.mjs`,
and that the entry bundles to a provider-shaped module. Nothing in the type system
relates those two files, which is why this broke silently.

> [!NOTE]
> ### Emit the Bedrock provider as a separate chunk in the CLI bundle
> - Adds [bedrock-bundle-entry.js](https://github.com/PrimeIntellect-ai/prime-agent/pull/1085/files#diff-510c77121e475fa6dab6c0a262ee12bc64963dcd4474c32dbd5b50efcc33b2c5) as an explicit entry point that re-exports `streamBedrock` and `streamSimpleBedrock`, allowing the bundler to emit an `amazon-bedrock.js` chunk resolvable by the runtime dynamic import.
> - Updates [bundle.mjs](https://github.com/PrimeIntellect-ai/prime-agent/pull/1085/files#diff-8606e8936a1a9a467bc717dab8da4d6bafd17f129a8e044e3054e4567b18803d) to declare two named entries (`cli` and `amazon-bedrock`) and adds a post-build guard that throws if `dist/bundle/amazon-bedrock.js` was not emitted.
> - Adds tests in [bundle-bedrock-entry.test.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/1085/files#diff-d4242b5cc302b2a83dd0751f84a8c23cf0c11f017e2f8c9cc9f1560a113f9f0c) that verify every runtime provider specifier has a matching bundle entry and that the emitted chunk exports the expected symbols.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 68e1ed1.</sup>
> 
>

### DIFF
--- a/packages/coding-agent/scripts/bedrock-bundle-entry.js
+++ b/packages/coding-agent/scripts/bedrock-bundle-entry.js
@@ -1,0 +1,14 @@
+// Bundle entry for the Bedrock provider.
+//
+// pi-ai loads Bedrock through a variable specifier (importNodeOnlyProvider) so
+// that browser bundles don't pull in the AWS SDK. esbuild therefore cannot see
+// the import and emits no chunk for it, while the runtime import still resolves
+// relative to the emitted chunk directory. Building this file as an entry named
+// "amazon-bedrock" puts the module exactly where that runtime import looks.
+//
+// Only pi-ai's public ./bedrock-provider export is used, and the named exports
+// match what loadBedrockProviderModule() destructures.
+import { bedrockProviderModule } from "@earendil-works/pi-ai/bedrock-provider";
+
+export const streamBedrock = bedrockProviderModule.streamBedrock;
+export const streamSimpleBedrock = bedrockProviderModule.streamSimpleBedrock;
--- a/packages/coding-agent/scripts/bundle.mjs
+++ b/packages/coding-agent/scripts/bundle.mjs
@@ -11,7 +11,7 @@
  * compiled Bun binary), keyed off the __PI_BUNDLED__ define below, so extension
  * imports of pi packages share the bundle's module instances.
  */
-import { chmodSync, readFileSync, rmSync } from "node:fs";
+import { chmodSync, existsSync, readFileSync, rmSync } from "node:fs";
 import { execFileSync } from "node:child_process";
 import { dirname, join } from "node:path";
 import { fileURLToPath } from "node:url";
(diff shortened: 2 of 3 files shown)


sethkarten:
Thank you for the report and proposed work. This root cause is now covered by maintainer-owned stacked PR #1158, authored independently from `upstream/main`.

We did not inspect or reuse this PR's diff, branch, commits, implementation code, or tests; its public description/comments were used only as a bug report. To keep one review surface, this PR is superseded by #1158 and is being closed.

PR #1095, opened 2026-08-09
fix(coding-agent): wait for the kernel to exit in dispose

### LINKED ISSUE #1049: Kernel dispose() returns before the kernel exits, leaving its working directory locked on Windows

r0h1tb:
`cleanupResources` killed the kernel and removed its temp directory in the same
synchronous pass, so `dispose()` resolved while the child was still exiting. The
kernel is spawned with `cwd` set to that directory; Windows keeps it locked
until the process is really gone, so a caller deleting it after an awaited
`dispose()` got `EPERM`.

fixes #1049

## The defect is not Windows-specific

Nothing ever waited for `"exit"`, so the promise resolved early on every
platform. Unix only hides it by unlinking files that are still open.

That is worth stating because it changes how it is tested. Rather than assert a
Windows filesystem error, the regression asserts the actual contract on the
process itself: once `dispose()` resolves, the kernel pid must be gone. That
reproduces deterministically on macOS and Linux.

Against unpatched `main`:

```
FAIL  test/suite/regressions/1049-kernel-dispose-waits-for-exit.test.ts
  → dispose() resolved before the kernel exited
  - false
  + true
Tests  1 failed | 1 passed (2)
```

With the fix, both pass.

## Change

`dispose()` defers temp-dir removal, waits for the process to actually exit,
then removes the directory.

- Direct children are awaited on their `"exit"` event.
- A forked kernel is not a direct child and emits no `"exit"`, so its pid is
  polled the same way `forkedKernelDied` already checks liveness — including
  treating `EPERM` as alive rather than gone.
- Both paths are bounded by `KERNEL_EXIT_TIMEOUT_MS` (2s) so a wedged kernel
  cannot hang disposal.
- Removal retries explicitly. As the issue notes, `rmSync`'s own `maxRetries`
  does not cover this: the native implementation surfaces the directory-level
  `EPERM` without retrying.

`disposeSync()` is deliberately unchanged. It runs from `process.on("exit")`
where nothing async can be awaited, so it keeps the previous best-effort
synchronous removal. `cleanupResources` gained an option rather than becoming
async, so that path is untouched.

## Verification

macOS arm64, Node v25.6.1.

```
test/suite/regressions/1049-kernel-dispose-waits-for-exit.test.ts   2 passed
test/kernel-startup.test.ts
test/kernel-state-snapshot.test.ts
test/kernel-fork-server.test.ts                                    18 passed
```

`npm run check` exits 0 (biome, tsgo --noEmit, installer render, browser smoke).
The pre-commit hook ran it again independently and passed.

The fake kernel in the test is a `/bin/sh` script that records its pid and
delays exit on SIGTERM, matching the style of `test/kernel-startup.test.ts`. No
provider, no API keys, no real IPython.

## Note on placement

AGENTS.md says issue regressions go under
`packages/coding-agent/test/suite/regressions/<issue-number>-<short-slug>.test.ts`,
so that is where it is. It does not use `test/suite/harness.ts` or the faux
provider, since it exercises `KernelManager` process lifecycle directly and
never starts an agent session. Happy to move it next to the other
`test/kernel-*.test.ts` files if you would rather keep the harness invariant
strict for that directory.

> [!NOTE]
> ### Fix `KernelManager.dispose` to wait for kernel process exit before removing temp directory
> - `dispose()` now captures the child process and PID before cleanup, sends SIGTERM via `cleanupResources`, then awaits the new `ensureKernelExited` helper before removing the temp directory.
> - `ensureKernelExited` waits up to `KERNEL_EXIT_TIMEOUT_MS` for a graceful exit, then escalates to SIGKILL and waits a further `KERNEL_SIGKILL_TIMEOUT_MS` if the process is still alive.
> - Temp directory removal is deferred until after exit and retried on transient failures via `removeTempDirWithRetries`; diagnostics are appended if the process or removal fails.
> - Adds a regression test suite covering: exit waiting, SIGKILL escalation, and caller/temp-dir removability after `dispose()` resolves.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 6e27c00.</sup>
> 
>

### DIFF
--- a/packages/coding-agent/src/core/kernel/index.ts
+++ b/packages/coding-agent/src/core/kernel/index.ts
@@ -39,6 +39,15 @@ const SNAPSHOT_MAX_OUTPUT_CHARS = 1_000_000;
 // on-disk copy is the fallback if this is exceeded.
 const SNAPSHOT_DISPOSE_TIMEOUT_MS = 5000;
 const KERNEL_ABORT_GRACE_MS = 1000;
+// An awaited dispose() must outlive the kill: the kernel is spawned with `cwd`
+// set to its temp dir, and Windows keeps that directory locked until the
+// process is really gone, so a caller deleting it after dispose() gets EPERM.
+const KERNEL_EXIT_TIMEOUT_MS = 2000;
+const KERNEL_EXIT_POLL_INTERVAL_MS = 10;
+// rmSync's own maxRetries does not cover this: the native implementation
+// surfaces the directory-level EPERM without retrying.
+const TEMP_DIR_REMOVE_ATTEMPTS = 10;
+const TEMP_DIR_REMOVE_DELAY_MS = 20;
 const KERNEL_BUSY_REUSE_WAIT_MS = 5000;
 const KERNEL_BUSY_INTERRUPT_INTERVAL_MS = 500;
 const MAX_LATE_SENT_AGENT_MESSAGE_HANDLERS = 256;
@@ -1285,7 +1294,55 @@ export class KernelManager {
 		await this.control.send(encode(msg, this.connection.key));
 	}
 
-	private cleanupResources(killSignal: NodeJS.Signals = "SIGTERM"): void {
+	/**
+	 * Resolves once the killed kernel has actually exited, or the timeout
+	 * elapses. A forked kernel is not a direct child and emits no "exit", so it
+	 * is polled the same way `forkedKernelDied` checks liveness.
+	 */
+	private static async waitForKernelExit(child: ChildProcess | undefined, pid: number | undefined): Promise<void> {
+		const deadline = Date.now() + KERNEL_EXIT_TIMEOUT_MS;
+
+		if (child && child.exitCode === null && child.signalCode === null) {
+			await new Promise<void>((resolve) => {
+				const done = () => {
+					globalThis.clearTimeout(timer);
+					child.removeListener("exit", done);
+					resolve();
+				};
+				const timer = globalThis.setTimeout(done, KERNEL_EXIT_TIMEOUT_MS);
+				if (typeof timer === "object" && "unref" in timer) timer.unref();
+				child.once("exit", done);
+			});
+			return;
+		}
+
+		if (pid === undefined) return;
+		while (Date.now() < deadline) {
+			try {
+				process.kill(pid, 0);
+			} catch (error) {
+				// ESRCH means gone. EPERM means the pid exists but is not ours to
+				// signal, which still counts as alive.
+				if (!(error instanceof Error && (error as NodeJS.ErrnoException).code === "EPERM")) return;
+			}
+			await new Promise((resolve) => globalThis.setTimeout(resolve, KERNEL_EXIT_POLL_INTERVAL_MS));
+		}
+	}
+
+	/** Removes a kernel temp dir, retrying while the OS still holds a handle. */
+	private static async removeTempDirWithRetries(dir: string): Promise<void> {
+		for (let attempt = 0; attempt < TEMP_DIR_REMOVE_ATTEMPTS; attempt++) {
+			try {
+				rmSync(dir, { recursive: true, force: true });
+				return;
+			} catch {
+				await new Promise((resolve) => globalThis.setTimeout(resolve, TEMP_DIR_REMOVE_DELAY_MS));
+			}
+		}
+		// Leave the temp dir for OS tmp cleanup.
+	}
+
+	private cleanupResources(killSignal: NodeJS.Signals = "SIGTERM", options: { removeTempDir?: boolean } = {}): void {
 		this.clearSnapshotTimer();
 		this.lateSentAgentMessageHandlers.clear();
 		if (this.forkedLivenessTimer) {
@@ -1314,14 +1371,16 @@ export class KernelManager {
 		this.kernel = undefined;
 		this.kernelPid = undefined;
 		this.connection = undefined;
-		if (this.tempDir) {
+		// dispose() defers this so it can wait for the kernel to exit first;
+		// disposeSync() has no such option and removes it best-effort here.
+		if (this.tempDir && options.removeTempDir !== false) {
 			try {
 				rmSync(this.tempDir, { recursive: true, force: true });
 			} catch {
 				// Leave the temp dir for OS tmp cleanup.
 			}
 		}
-		this.tempDir = undefined;
+		if (options.removeTempDir !== false) this.tempDir = undefined;
 		this.startPromise = undefined;
 	}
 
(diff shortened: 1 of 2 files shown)


r0h1tb:
Both fixed in 3a195bd.

You are right that the contract was best-effort in exactly the case the timeout
exists for. `ensureKernelExited` now sends SIGKILL when the SIGTERM deadline
passes and waits a second, shorter interval (`KERNEL_SIGKILL_TIMEOUT_MS`, 1s)
before giving up. The fast path is untouched — a well-behaved kernel exits on
SIGTERM long before the deadline and never reaches the escalation.

The regression covers it directly: a fake kernel that does `trap '' TERM` and
ignores the signal outright, standing in for one blocked in a native call. With
the escalation removed it fails as

```
× escalates to SIGKILL when the kernel ignores SIGTERM
AssertionError: dispose() resolved with a SIGTERM-ignoring kernel still alive
  expected true to be false
Tests  1 failed | 2 passed
```

and the other two still pass, so it pins the escalation specifically rather than
the original wait.

On the smaller point: `removeTempDirWithRetries` now returns its last error
instead of swallowing everything, and `dispose()` records a kernel diagnostic on
final failure. A retry loop is still right for a transient lock, but a genuine
EACCES or a bad path is now visible instead of silently leaving a directory
behind. Behaviour is otherwise unchanged — it still gives up rather than
throwing, since disposal should not fail on cleanup.

Also took the pointer to `session-lease.ts`: the liveness check is extracted as
`isPidAlive` so the EPERM-means-alive convention is explicit and greppable
against `isProcessAlive`. I left it local rather than exporting from
session-lease, to avoid coupling the modules for four lines — say the word if
you would rather have one shared helper.

`npm run check` exits 0; the kernel test files pass (21 tests across 4 files).

r0h1tb:
Correction to my previous comment, before it misleads review.

I accepted the framing that `removeTempDirWithRetries` was "racing a directory
the OS still has locked" without checking which directory it operates on. It is
not the locked one.

- `this.tempDir` is `mkdtempSync(join(tmpdir(), "prime-agent-kernel-"))`, the
  connection-file directory holding `connection.json` (`makeConnection`).
- The kernel is spawned with `cwd: this.options.cwd` — the caller's directory —
  on both the direct-spawn and forkserver paths.

Windows holds the **cwd** open, and `removeTempDirWithRetries` never touches it.
So the retry loop is defensive housekeeping for the connection dir, and the
claim in my commit message that its exhaustion "is the original #1049 symptom"
is wrong.

This does not weaken the escalation — it relocates why it matters. The issue's
repro is

```ts
await provisioner.dispose();
rmSync(tempDir, { recursive: true, force: true });   // EPERM
```

where `tempDir` is what the caller passed as `cwd`. Nothing in KernelManager
removes that directory; the caller does, right after the promise resolves. So
the only thing that makes it removable is the kernel actually being gone — which
is precisely what SIGTERM-without-escalation failed to guarantee.

Added 8fdd604 asserting exactly that: a kernel that ignores SIGTERM, then the
caller removing its own cwd immediately after `await dispose()`. It passes with
the escalation and is the closest thing to the reported failure that reproduces
off Windows.

Also confirmed the two kill paths are mutually exclusive — the forkserver path
sets only `kernelPid` and the direct path only `kernel` — so `ensureKernelExited`
always has exactly one of them, and the SIGKILL branch cannot double-signal.

4 tests in the regression file, 22 across the four kernel files, `npm run check`
exits 0.

r0h1tb:
Good catch — fixed in the latest commit.

`isForkServerEnabled()` returns false on non-Linux, so this never showed up
locally. Set `PRIME_AGENT_KERNEL_FORKSERVER=0` for these tests, matching
`ipython-provisioner.test.ts`, which pins direct-spawn for the same reason.

25 tests pass across the regression file plus `kernel-startup`,
`kernel-fork-server` and `ipython-provisioner`.

BILLKISHORE:
Framing this as a contract on the process rather than a Windows filesystem error is the right call, and it makes the regression deterministic on macOS and Linux. Treating `EPERM` from `process.kill(pid, 0)` as alive also matches the convention `isProcessAlive` uses in `session-lease.ts`, which is the safe direction.

The gap is that the contract is best-effort in exactly the case the timeout exists for. `dispose()` sends `SIGTERM`, waits up to `KERNEL_EXIT_TIMEOUT_MS`, then proceeds regardless, and there is no escalation. A kernel that is slow to handle `SIGTERM`, blocked in a long-running native call, or forked such that signal delivery is less certain, is still alive when `dispose()` resolves. `removeTempDirWithRetries` then gets 10 attempts over 200ms against a directory the OS still has locked, so it exhausts them and leaves the directory behind, which is the original #1049 symptom.

Escalating on timeout would close it: if the process is still alive when the deadline passes, send `SIGKILL` and wait a second short interval before giving up. That keeps the fast path unchanged, since a well-behaved kernel exits on `SIGTERM` long before the deadline, and makes "once `dispose()` resolves, the pid is gone" true rather than usually true.

Smaller point: `removeTempDirWithRetries` swallows every error and gives up silently. A retry loop for a transient lock makes sense, but a genuine `EACCES` or a bad path currently leaves a directory behind with no signal that anything went wrong. Logging on the final failure would make that visible without changing behaviour.

kevinjosethomas:

PR #1102, opened 2026-08-09
fix(ai): repair interrupted tool result ordering

junhoyeo:
## Summary

- Repair interrupted tool turns when a persisted user/custom message separates a tool call from its delayed real result.
- Hoist matching real results before the interposed message, synthesize only genuinely missing results, and discard invalid orphan or duplicate results.
- Heal already-persisted affected sessions in memory without rewriting append-only session evidence.

## Root cause

An update restart can persist this temporal order while aborting an in-flight tool:

```text
assistant(tool call T)
custom update-restart marker
real tool result T: Request was aborted
```

The marker becomes a user message in model context. The existing transform synthesized a replacement result before that user boundary but later emitted the real result too. Anthropic then rejected the late result because its immediately preceding message had no matching `tool_use`, permanently blocking every subsequent request that replayed the same history.

The provider transform is the shared repair boundary for this invariant. Matching is bounded at the next assistant message, occurs after tool-call ID normalization, preserves real results over synthetic ones, and is a no-op for well-formed turns.

## Validation

- `cd packages/ai && npx tsx ../../node_modules/vitest/dist/cli.js --run test/transform-messages-copilot-openai-to-anthropic.test.ts` — 10 passed
- `npm run check` — passed on the upstream `main` base; Biome checked 900 files with no fixes, type checking and installer/browser smoke checks passed
- Replayed the original affected local session through `buildSessionContext` → `convertToLlm` → patched `transformMessages`: 1,904 persisted entries, 377 transformed messages, the real interrupted result preserved, and zero tool-pairing violations
- GPT-5.6 Terra adversarial review — PASS after reducing the patch to the provider-boundary repair and avoiding update-daemon lifecycle changes

> [!NOTE]
> ### Fix tool result ordering when an assistant turn is interrupted by a user message
> - Reworks the second pass in [`transformMessages`](https://github.com/PrimeIntellect-ai/prime-agent/pull/1102/files#diff-48d527355851437ed4f7f9a76cffbb8aace06900f010e02177e16b335579c674) to correctly handle tool results that arrive after an interposed user message during an interrupted assistant turn.
> - Hoists valid tool results across an intervening user message to match pending tool calls, marks them consumed to avoid double-processing, and synthesizes error results for any unresolved tool calls at user boundaries and end-of-conversation.
> - Drops orphaned tool results (no matching pending call) and duplicate tool results.
> - Behavioral Change: messages with roles other than `assistant`, `toolResult`, and `user` are no longer forwarded in the second pass.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 0e0452c.</sup>
> 
>

### DIFF
--- a/packages/ai/CHANGELOG.md
+++ b/packages/ai/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## [Unreleased]
 
+- Fixed interrupted tool turns replaying late tool results after an interposed user message ([#1102](https://github.com/PrimeIntellect-ai/prime-agent/pull/1102) by [@junhoyeo](https://github.com/junhoyeo)).
+
 ## [0.7.1] - 2026-08-07
 
 ## [0.7.0] - 2026-08-05
--- a/packages/ai/src/providers/transform-messages.ts
+++ b/packages/ai/src/providers/transform-messages.ts
@@ -152,31 +152,40 @@ export function transformMessages<TApi extends Api>(
 		return msg;
 	});
 
-	// Second pass: insert synthetic empty tool results for orphaned tool calls
-	// This preserves thinking signatures and satisfies API requirements
+	// Second pass: pair every emitted tool result with a currently pending tool call.
+	// A persisted user message can race an abort result; recover that result from the
+	// remainder of the same turn before synthesizing anything at the user boundary.
 	const result: Message[] = [];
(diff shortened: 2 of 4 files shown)


sethkarten:

PR #1104, opened 2026-08-10
feat(coding-agent): rank and budget the continual harness overview

dattgoswami:
Fixes Issue #819
Maintainers: this wants the `pkg:coding-agent` label. I cannot apply it - the label does not exist
in this repository yet, and an outside contributor has no triage permission.

## What this changes

`formatHarnessStateForPrompt` renders the continual harness into the system prompt. It sorted
entries on `[path, title, id]` and kept the first six per kind, so which lessons the model saw was
decided by how their paths happened to be spelled, not by recency, version, or relevance. A lesson
written by `/refine` reached the next turn only if its path sorted into the first six for its kind;
everything else collapsed into `- +N more <kind> entries`, which names nothing the model can look
up. The options object already accepted `maxEntriesPerKind`/`maxContentLength`, but neither
production call site passed them and no setting reached them, so the limits behaved as hardcoded.

Three changes:

1. **Ranking.** Entries are ordered `updated_at` descending, then `version` descending, then a
   code-point comparison of `[path, title, id, scope]`. The entry a refinement just wrote is now
   always rendered first, with its content. This is expectation A(a) in the issue.
2. **Addressable menu.** Entries past `maxEntriesPerKind` get a one-line stub carrying their
   `scope:id` — the exact address `rlm.harness.get(kind, "global:<id>")` accepts, verified against
   `_strip_scope_prefix` (`prime-agent-runtime/src/rlm/harness.py:59`). An entry the prompt names is
   an entry the model can actually read, so the address is never truncated: only the title and path
   absorb the line clip, and an entry that will not fit is reported in the count rather than
   rendered in a broken form.
3. **A budget, not a count.** The menu spends at most `maxListedChars` characters — 8,000 by
   default, about 2,000 tokens — across all kinds together, so the prompt size stops tracking the
   store size. A `continualHarness` settings block exposes that budget along with the detail count,
   content clip, and detail rail, resolved in `SettingsManager` and passed through **both**
   `buildSystemPrompt` render paths. This is expectation A(b).

Selection is a pure exported function, `selectOverviewEntries`, so a future relevance- or
utility-ranked policy has one place to land. It takes an unused `query` argument reserved for that.

## Measurements

Issue #819's own 49-entry repro fixture, rendered with no options, against `main` at `a18809e00`:

| render | chars | entries named | newest entry visible |
| --- | --- | --- | --- |
| `main` | 3,951 | 6 / 49 | no |
| this branch | 6,480 | 49 / 49 | yes, with content |
| every body in full (reference) | 15,290 | 49 / 49 | yes |

`main` reproduces the issue's numbers exactly, including its rendered ids
(`mem-000, mem-026, mem-001, mem-027, mem-002, mem-028`) and its `- +43 more memory entries` line.
Full addressability costs 2,529 characters here, roughly 630 tokens, and the whole overview is 42%
of what dumping every body would cost.

What a maintainer actually cares about is the **whole system prompt**, not the overview in
isolation, so the growth tables below are `buildSystemPrompt` output. Ordinary metadata means
~220-character bodies and ordinary ids and paths; all four kinds are filled, which is the ceiling
rather than a shape anyone has.

| entries per kind | `main` chars (~tokens) | branch chars (~tokens) | `main` named | branch named |
| --- | --- | --- | --- | --- |
| 6 | 14,685 (~3,671) | 14,685 (~3,671) | 24 | 24 |
| 50 | 14,826 (~3,707) | 22,662 (~5,666) | 24 | 152 |
| 200 | 14,858 (~3,715) | 22,637 (~5,659) | 24 | 147 |
| 500 | 14,858 (~3,715) | 22,637 (~5,659) | 24 | 147 |
| 5,000 | 14,882 (~3,721) | 22,718 (~5,680) | 24 | 144 |

The branch's prompt stops growing once the menu budget is spent. Overview-only, memory alone, which
is the kind that actually grows:

| memory entries | `main` chars | branch chars | `main` named | branch named |
| --- | --- | --- | --- | --- |
| 6 | 3,900 | 3,900 | 6 | 6 |
| 49 | 3,935 | 6,387 | 6 | 49 |
| 200 | 3,943 | 11,902 | 6 | 140 |
| 500 | 3,943 | 11,895 | 6 | 139 |
| 5,000 | 3,949 | 11,921 | 6 | 135 |

### The bound

Stub lines interpolate `id`, `title`, and `path`, none of which is validated anywhere in the
codebase, and there are as many of them as the store has entries. Capping the *number* of stubs
therefore caps neither their cost nor the prompt. The menu instead has a single character budget:

```
stub menu <= maxListedChars = 8,000 characters, about 2,000 tokens, for the whole overview
```

It is split evenly between the kinds that actually have entries to list, so one large kind cannot
starve the others and a memory-only store spends the whole budget on memory. Stubs render in rank
order until the share is spent; an entry whose line does not fit is skipped rather than ending the
menu, so one entry with a very long id costs itself and not everything ranked below it. Everything
unrendered is reported in the `+N more <kind> entries` count.

Measured, as the difference against the same render with the menu disabled:

| store | menu cost |
| --- | --- |
| ordinary metadata, 200 entries per kind | 7,779 chars (~1,945 tokens) |
| ordinary metadata, 5,000 entries per kind | 7,836 chars (~1,959 tokens) |
| 1,000-character ids, titles, paths, 200 per kind | 4,093 chars (~1,023 tokens) |
| 1,000-character ids, titles, paths, 5,000 per kind | 4,097 chars (~1,024 tokens) |

**Why a budget rather than a bigger count cap.** `models.generated.ts:1571` ships a model with a
16,384-token context window, and this repository ships several smaller ones. A per-kind count
ceiling of 200 stubs produced a ~16,596-token system prompt for an ordinary 200-entry-per-kind
store — the prompt did not fit before a single user token. The same store is now ~5,659 tokens.
"Every entry is addressable" and "the prompt fits a 16k window" cannot both hold for a 20,000-entry
store; what #819 actually asks for is that a newly written lesson reaches the next turn, and that
comes from the ranking, not from naming all 20,000. Long addresses buy fewer named entries rather
than a longer menu, which is the correct trade and is what keeps every rendered address complete.

Within the budget each stub line targets 160 characters — about twice a realistic one, since
`- [global:mem-042] workspace-resolver-fix (memory/w/workspace-resolver-fix.md, v7)` is 82 — and
the `[scope:id]` address is exempt from that clip at any id length.

**What is not bounded here, stated plainly.** With 1,000-character ids, titles, and paths the whole
prompt is ~22,630 tokens, but `main` is ~21,563 tokens for the same store. That cost is the detail
tier's unclipped headline, a pre-existing `main` behaviour bounded by `maxEntriesPerKind`. Clipping
it would change what `main` produces for small stores, so it is left alone and is not claimed as
bounded. This PR adds ~1,024 tokens on top of it.

## Backward compatibility

- A store with `maxEntriesPerKind` entries or fewer per kind renders **the same lines** as `main`,
  each byte-for-byte, and renders them **byte-identically** when the entries also tie on
  `updated_at` and `version`. Verified against a `git worktree` at `upstream/main` at 0, 1, 3, and 6
  entries per kind across all four kinds, and again at 6 entries per kind with 1,000-character ids,
  titles, and paths (86,253 characters, byte-identical). When timestamps differ the line set is
  identical and the order is recency-first, which is the point of the change.
- Callers passing `maxEntriesPerKind` get exactly `min(N, stored)` content-bearing entries. Nothing
  else reduces that count.
- `detailBudget` has no binding default. It is an opt-in character rail for operators who raise
  `maxEntriesPerKind` a long way; when set, the top-ranked entry always keeps its content, so no
  configuration can hide the lesson a refinement just wrote.
- Invalid limits normalize in one place, in the renderer, so a settings value and a direct call
  behave identically. Anything that is not a number, is not finite, or is below a key's minimum
  falls back to that key's default; fractional values are floored; nothing is clamped, because
  `maxListedEntries: -1` is a typo and clamping it to `0` would disable the menu with no
  diagnostic. Every stored entry is always accounted for as content, a stub, or the overflow count.

## The `+N more <kind> entries` line

AGENTS.md says to ask before removing intentional functionality, so, stated plainly: that line is
**intentionally demoted, not deleted**, and it is precisely what #819 identifies as the defect —
"the model has no in-prompt way to reach the hidden entries". A count is not addressable; a stub is.

It still renders for everything the menu budget cannot afford, where a count is genuinely the only
option left. Setting either `continualHarness.maxListedChars: 0` or
`continualHarness.maxListedEntries: 0` restores the previous render exactly, byte-identical to
`main` at 8, 49, and 500 entries, so the old behavior stays reachable for anyone who prefers it.
Happy to change the default if maintainers would rather keep the count.

## Determinism

`selectOverviewEntries` reads no clock, no filesystem, no randomness, no ambient locale, and no
ambient timezone, so two renders of the same `harness_state.json` are byte-identical on any machine
— which prompt caching depends on. Three deliberate changes get it there:

- `updated_at` is compared as a parsed epoch rather than a raw string. The TypeScript writer emits
  `...Z` and the Python writer emits `...+00:00` into the same file, so lexicographic comparison
  would order the two writers inconsistently. `loadHarnessState` does not validate timestamps, so
  missing or unparseable values rank last rather than throwing.
- An ISO date-time carrying **neither an offset nor a trailing `Z`** is read as UTC. `Date.parse`
  reads that form in the host timezone, and `loadHarnessState` accepts it, so without normalizing
  the same file ranked differently on two machines:

  ```
  updated_at of 2026-06-01T03:00:00Z and 2026-06-01T00:00:00
    TZ=UTC                 before = fixed, local     after = fixed, local
    TZ=America/Edmonton    before = local, fixed     after = fixed, local
  ```

  The normal writers both emit offsets, so this is a robustness gap rather than something users hit
  today — but the determinism claim above is load-bearing and the input is unvalidated.
- The `[path, title, id]` tiebreak is now a code-point comparison with `scope` appended, replacing
  `localeCompare`. **This changes tiebreak order relative to `main`**, and it is a deliberate trade:
  `localeCompare` with no explicit locale is ICU-dependent, so `main` orders the same store
  differently on different machines.

  ```
  paths memory/ä.md and memory/z.md, equal updated_at and version
    LC_ALL unset     main = ä, z     branch = z, ä
    LC_ALL=sv_SE     main = z, ä     branch = z, ä

  paths memory/Zebra.md and memory/apple.md
    main = apple, Zebra              branch = Zebra, apple
  ```

  The mixed-case ASCII case matters as much as the non-ASCII one and is easy to miss:
  `localeCompare` sorts `apple` before `Zebra`, code points do not.

  Appending `scope` also makes the order total. `mergeHarnessStates` deliberately keeps a local entry
  alongside a global one with the same id, and those pairs previously compared equal, so their order
  followed input order.

`Date.parse` truncates sub-millisecond precision, so same-millisecond writes fall through to
`version` and then to the tiebreak.

## Review round

The branch carries three commits. The first implemented the change above; two adversarial review
rounds followed, and each later commit fixes everything the round before it found. They are
described here rather than buried, because in both rounds a defect invalidated a claim the previous
description made, and a reviewer is entitled to know that the numbers above are the corrected ones.

**Round 1**, against the first commit:

| # | Severity | Defect | Resolution |
| --- | --- | --- | --- |
| 1 | High | `maxListedEntries` capped stub *count*, not stub *length*; ids, titles, and paths are unvalidated, so 1,000-character metadata rendered 1.6M characters | Each assembled stub line clipped to 160 characters, giving the closed-form ceiling above |
| 2 | High | A default 4,000-character `detailBudget` silently overrode the caller's requested depth: `maxEntriesPerKind: 50` rendered 16 entries, and six entries with long titles rendered one body instead of six | Default detail budget removed; `maxEntriesPerKind` alone decides depth |
| 3 | Medium | `localeCompare` made ordering machine-dependent, and `scope` was missing from the tiebreak, so exact local/global pairs compared equal and followed input order | Code-point comparison with `scope` appended |
| 4 | Medium | Direct callers were unvalidated: `maxListedEntries: NaN` poisoned both the slice and the overflow count, dropping four of ten entries with nothing in the prompt indicating they existed | One clamp, in the renderer, so settings and direct callers cannot disagree |
| 5 | Medium | The regression test called the formatter directly, contrary to `test/suite/README.md`, and could not see the settings handoff at all — deleting it left every test green | Rewritten on `test/suite/harness.ts`, asserting through `AgentSession`'s rebuilt system prompt |
| 6 | Low | No CHANGELOG entry | Added, issue-attributed, as the final commit |
| 7 | Low | Two files outside the originally declared scope | Documented below; no code change |
| 8 | High | The description asserted two things findings 1 and 2 disprove | Rewritten |

**On findings 1 and 2, which are one design error.** The two budgets sat on the wrong tiers. The
detail tier was already bounded by a count cap and got a character budget on top; the stub tier,
the one that actually grows with the store, got only a count cap. The principle applied to both:
**counts are authoritative, characters bound only what the caller did not ask for.**

That is why the detail budget was dropped outright rather than derived from the other limits. A
detail line is a headline plus content, and the headline holds the unbounded `id`, `title`, and
`path`. Any budget computed from `maxContentLength` still binds on six entries with 1,000-character
titles — exactly the failure in finding 2. On a tier already capped by count, a character budget can
only ever reduce the count below what was requested, so "never contradicts the caller" and "actually
binds" are mutually exclusive. The per-line stub clip achieves the bound without that contradiction,
which is also why a separate per-kind stub character budget was rejected: it would reproduce finding
2's shape one tier down.

**Two claims from the first description did not survive, and are not restated here.** The old
"byte-identical for small stores" claim was verified only against fixtures with uniform timestamps;
ranking by `updated_at` changes order relative to `main` whenever timestamps differ, which was true
of the first commit too. The claim above is narrower and both halves of it are measured. The old
`maxEntriesPerKind` claim was simply false until finding 2 was fixed.

**Round 2**, against the round-1 fixes:

| # | Severity | Defect | Resolution |
| --- | --- | --- | --- |
| 1 | High | Clipping the assembled stub line cut through the id. Ids are unvalidated, so a 200-character id rendered a 160-character line with no closing bracket; in the 5,000-entry adversarial fixture, 0 of 800 stubs kept a usable address | The `[scope:id]` address is built first and never clipped; only the title and path absorb it, and an entry that cannot fit goes to the count instead of rendering broken |
| 2 | High | The per-kind count ceiling allowed ~32,200 tokens across four kinds, and an ordinary 200-entry-per-kind store already produced a ~16,596-token system prompt — larger than the 16,384-token window `models.generated.ts:1571` ships | `maxListedChars`: one 8,000-character budget for the whole menu, split evenly between the kinds that need it. That store is now ~5,659 tokens and does not grow |
| 3 | Medium | `Date.parse` reads an offsetless ISO date-time in the host timezone, so a state file `loadHarnessState` accepts ranked differently under `TZ=UTC` and `TZ=America/Edmonton` | Offsetless timestamps are read as UTC |
| 4 | Low | `docs/settings.md` promised out-of-range values fall back to defaults; the renderer clamped them to the minimum, so `maxListedEntries: -1` disabled the stub menu silently | The renderer now does what the documentation said. `0` remains the deliberate opt-out |
| 5 | Low | Two files outside the originally declared scope, restated from round 1 | Declined, with reasons, below |

**Findings 1 and 2 were one design error**, and one mechanism fixes both. A count cap bounds neither
the cost of a line nor the size of the prompt; a character budget bounds both, and it makes a long
address cost what it actually costs. Once the menu is budgeted, there is no reason to truncate an
address to fit a line — an unaddressable stub is worse than no stub, because it advertises knowledge
the model cannot retrieve, which is the defect #819 is about.

**One claim from the round-1 description did not survive.** It said the clip "keeps the leading
`[scope:id]`, which is the part the model needs". That was true only for short ids. It is now true
unconditionally, and tested at 200-character and 1,000-character ids by parsing every rendered stub
rather than by matching a substring.

## Files outside the originally declared scope

- `src/core/agent-session.ts` — one added property in an existing object literal (`harnessOverview`).
  `buildSystemPrompt` is a pure function with no settings access, and its only production caller is
  that literal, so without this line the `continualHarness` key is inert. A regression test now
  fails if the line is removed.
- `test/settings-manager.test.ts` — covers `getContinualHarnessSettings`, in the file that already
  tests the sibling `autoRefine` resolver.

`docs/settings.md` and `CHANGELOG.md` are also updated, both as repository policy requires.

`settings-manager.ts` needed no further change for the new `maxListedChars` key:
`ContinualHarnessSettings` is a type alias of `HarnessOverviewLimits`, so the settings surface and
the renderer cannot drift apart.

## Tests

`test/suite/regressions/819-harness-overview-cap.test.ts` uses the issue's own fixture through
`test/suite/harness.ts` and asserts against `AgentSession`'s rebuilt system prompt, so it covers the
settings handoff as well as the renderer. It **fails on a clean `main` worktree**:

```
x renders the most recently written entry with its content
    expected '# Continual Harness State...' to contain
      '- [global:mem-new] workspace-resolver-fix (memory/w/workspace-resolver-fix.md, v7): ...'
x names every stored entry so the model can address it by id
    expected [ 'mem-000', 'mem-026', ...(4) ] to have a length of 49 but got 6
x applies continualHarness settings to the session's system prompt
    expected [ 'mem-000', 'mem-026', ...(4) ] to deeply equal [ 'mem-new' ]

Test Files  1 failed (1)
     Tests  3 failed | 2 passed (5)
```

Deleting the `harnessOverview` line from `agent-session.ts` fails the third of those, which is the
point of routing the test through the session rather than the formatter. A fifth case asserts that a
1,600-entry store still builds a system prompt under 8,000 tokens, so the fix for #819 cannot be
traded for a prompt that does not fit.

In `test/refinement.test.ts`: `selectOverviewEntries` unit tests (recency ranking, version and
code-point tiebreaks, the local/global collision, determinism under input reordering, budget
exhaustion, reserved slots, empty store, unparseable timestamps, input not mutated) and renderer
tests for the limits — exact detail depth across five values of `maxEntriesPerKind`; the menu budget
held at four kinds x 5,000 entries with 1,000-character metadata and at ordinary 200 and 500 per
kind; every rendered stub parsed with `/^- \[(local|global):([^\]]+)\]/` and its captured id
checked against the store, at 200-character and 1,000-character ids; offsetless timestamps ranked as
UTC; the documented fallback asserted per key; and a 30-case table asserting every entry stays
accounted for under `NaN`, positive and negative `Infinity`, `-1`, `0`, and fractional limits.

`npm run check` is green with no errors, warnings, or infos. 151 tests pass across
`test/refinement.test.ts`, `test/system-prompt.test.ts`, `test/settings-manager.test.ts`, and the
regression file — identically under `TZ=America/Edmonton` and under `LC_ALL=sv_SE.UTF-8`, since both
guard the determinism claim above. A further 332 pass across six neighbouring refine and prompt
suites.

## Scope

This closes defect **A** only. Defect **B** — `rlm.harness` defaulting to the session-local store,
which is empty in an `rlm()`-spawned child — is untouched and lives on the Python side.

The three-way cap disagreement the issue notes (prompt 6, refiner `overviewForPrompt` 40, Python
`rlm.harness.overview()` 20) is also untouched. `overviewForPrompt` shares no code with this render
path and belongs in its own change.

> [!NOTE]
> ### Rank and budget the continual harness overview in system prompts
> - Entries in the harness overview are now ranked by recency and version using a deterministic comparator, then split into content-bearing detailed entries and a stub menu constrained by per-kind count caps and a shared character budget (`maxListedChars`).
> - The stub menu is dynamically sized against the active model's context window: [`system-prompt.ts`](https://github.com/PrimeIntellect-ai/prime-agent/pull/1104/files#diff-db1412afec77d3c6d261173c97d9073fabd7f40adb493137581377a806e8b529) measures the base prompt length and computes an affordable budget via `affordableOverviewListedChars`, shrinking the menu if the window is tight.
> - Entry IDs containing brackets or control characters are now rejected on create/update (both in TypeScript and Python) and excluded from rendering, but counted in the `+N more` overflow and still deletable for legacy cleanup.
> - [`agent-session.ts`](https://github.com/PrimeIntellect-ai/prime-agent/pull/1104/files#diff-a72e8a0b5f514ca1def756ee863a6e8a556df115e7167a50ca4f5e66b14448ae) rebuilds the system prompt whenever the model's context window changes, keeping the harness budget consistent across model switches.
> - `SettingsManager` exposes `getContinualHarnessSettings()` so per-session `continualHarness` limits flow into prompt generation; see [`docs/settings.md`](https://github.com/PrimeIntellect-ai/prime-agent/pull/1104/files#diff-fab1566cd80cf0a84c29db9db081b306b113186245fa2356e42a14f9c3b6b552) for the new configuration keys.
> - Behavioral Change: entries with non-string title/content, empty IDs, or unaddressable IDs are silently dropped or normalized at load time; create proposals with an empty ID now fall back to a title-derived slug instead of storing an empty-ID entry.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 568aab7.</sup>
> 
>

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -5,7 +5,7 @@
 - Added privacy-safe pseudonymous product analytics for onboarding, command use, execution modes, run outcomes, TTFT, latency, usage, tools, retries, and compactions, with disclosure and opt-out controls ([ENG-4682](https://linear.app/primeintellect/issue/ENG-4682/add-privacy-safe-posthog-analytics-to-prime-agent)).
 - Changed sent agent messages in the IPython cell UI to show only the message text with a `╰─` gutter when expanded, matching received messages, and hid the raw `agent_message.send` receipt dictionary.
 - Fixed Homebrew installs attempting to self-update their versioned Cellar keg instead of directing users to `brew upgrade prime-agent` ([#844](https://github.com/PrimeIntellect-ai/prime-agent/issues/844))
-- Fixed the continual harness overview hiding recent entries behind a path-ordered cap, ranking entries by recency and listing every remaining entry by its addressable id, with `continualHarness` settings for the limits ([#819](https://github.com/PrimeIntellect-ai/prime-agent/issues/819))
+- Fixed the continual harness overview hiding recent entries behind a path-ordered cap, ranking entries by recency and naming the rest by their addressable ids within a bounded menu, with `continualHarness` settings for the limits ([#1104](https://github.com/PrimeIntellect-ai/prime-agent/pull/1104) by [@dattgoswami](https://github.com/dattgoswami))
 
 ## [0.7.1] - 2026-08-07
 
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -5,7 +5,8 @@
 - Added privacy-safe pseudonymous product analytics for onboarding, command use, execution modes, run outcomes, TTFT, latency, usage, tools, retries, and compactions, with disclosure and opt-out controls ([ENG-4682](https://linear.app/primeintellect/issue/ENG-4682/add-privacy-safe-posthog-analytics-to-prime-agent)).
 - Changed sent agent messages in the IPython cell UI to show only the message text with a `╰─` gutter when expanded, matching received messages, and hid the raw `agent_message.send` receipt dictionary.
 - Fixed Homebrew installs attempting to self-update their versioned Cellar keg instead of directing users to `brew upgrade prime-agent` ([#844](https://github.com/PrimeIntellect-ai/prime-agent/issues/844))
-- Fixed the continual harness overview hiding recent entries behind a path-ordered cap, ranking entries by recency and naming the rest by their addressable ids within a bounded menu, with `continualHarness` settings for the limits ([#1104](https://github.com/PrimeIntellect-ai/prime-agent/pull/1104) by [@dattgoswami](https://github.com/dattgoswami))
+- Fixed the continual harness overview hiding recent entries behind a path-ordered cap, ranking entries by recency and naming the rest by their addressable ids within a menu budgeted against the session model's context window, with `continualHarness` settings for the limits ([#1104](https://github.com/PrimeIntellect-ai/prime-agent/pull/1104) by [@dattgoswami](https://github.com/dattgoswami))
+- Fixed harness entry ids containing `[`, `]`, or control characters rendering as overview addresses that resolve to nothing: the TypeScript refiner and the Python harness now reject such ids at create/update, and legacy entries carrying one are reported only through the `+N more` count while remaining deletable ([#1104](https://github.com/PrimeIntellect-ai/prime-agent/pull/1104) by [@dattgoswami](https://github.com/dattgoswami))
 
 ## [0.7.1] - 2026-08-07
 
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -5,6 +5,7 @@
 - Added privacy-safe pseudonymous product analytics for onboarding, command use, execution modes, run outcomes, TTFT, latency, usage, tools, retries, and compactions, with disclosure and opt-out controls ([ENG-4682](https://linear.app/primeintellect/issue/ENG-4682/add-privacy-safe-posthog-analytics-to-prime-agent)).
 - Changed sent agent messages in the IPython cell UI to show only the message text with a `╰─` gutter when expanded, matching received messages, and hid the raw `agent_message.send` receipt dictionary.
 - Fixed Homebrew installs attempting to self-update their versioned Cellar keg instead of directing users to `brew upgrade prime-agent` ([#844](https://github.com/PrimeIntellect-ai/prime-agent/issues/844))
+- Fixed the continual harness overview hiding recent entries behind a path-ordered cap, ranking entries by recency and listing every remaining entry by its addressable id, with `continualHarness` settings for the limits ([#819](https://github.com/PrimeIntellect-ai/prime-agent/issues/819))
 
 ## [0.7.1] - 2026-08-07
 
--- a/packages/coding-agent/docs/settings.md
+++ b/packages/coding-agent/docs/settings.md
@@ -126,6 +126,37 @@ prime-agent --offline
 }
 ```
 
+### Continual Harness
+
+Controls how much of the continual harness (`/refine` prompt notes, memories, skills, and subagent
+specs) is rendered into the system prompt. Entries are ranked by `updated_at`, then `version`, then
+path, so the most recently refined entry always keeps its content. Every entry that does not fit the
+detail budget is still listed as a one-line stub carrying its `scope:id`, which is the address
+`rlm.harness.get(kind, "global:<id>")` accepts.
+
+| Setting | Type | Default | Description |
+|---------|------|---------|-------------|
+| `continualHarness.maxEntriesPerKind` | number | `6` | Entries per kind rendered with their content |
+| `continualHarness.maxContentLength` | number | `180` | Characters kept per rendered content, `ref=`, and `args=` value |
+| `continualHarness.detailBudget` | number | `4000` | Characters spent on content-bearing entries per kind |
+| `continualHarness.maxListedEntries` | number | `200` | One-line stubs rendered per kind after the content-bearing entries |
+
+`detailBudget` does not bind at the default `maxEntriesPerKind`; it is the ceiling that keeps a
+raised `maxEntriesPerKind` from dominating the prompt. The top-ranked entry always keeps its content
+regardless of the budget. Entries beyond `maxListedEntries` collapse into a `+N more <kind> entries`
+count; setting it to `0` turns the stub list off entirely and restores the count-only overview.
+
+```json
+{
+  "continualHarness": {
+    "maxEntriesPerKind": 6,
+    "maxContentLength": 180,
+    "detailBudget": 4000,
+    "maxListedEntries": 200
+  }
+}
+```
+
 ### Branch Summary
 
 | Setting | Type | Default | Description |
--- a/packages/coding-agent/docs/settings.md
+++ b/packages/coding-agent/docs/settings.md
@@ -130,28 +130,35 @@ prime-agent --offline
 
 Controls how much of the continual harness (`/refine` prompt notes, memories, skills, and subagent
 specs) is rendered into the system prompt. Entries are ranked by `updated_at`, then `version`, then
-path, so the most recently refined entry always keeps its content. Every entry that does not fit the
-detail budget is still listed as a one-line stub carrying its `scope:id`, which is the address
+path, so the most recently refined entry always keeps its content. Every entry past
+`maxEntriesPerKind` is still listed as a one-line stub carrying its `scope:id`, which is the address
 `rlm.harness.get(kind, "global:<id>")` accepts.
 
 | Setting | Type | Default | Description |
 |---------|------|---------|-------------|
 | `continualHarness.maxEntriesPerKind` | number | `6` | Entries per kind rendered with their content |
 | `continualHarness.maxContentLength` | number | `180` | Characters kept per rendered content, `ref=`, and `args=` value |
-| `continualHarness.detailBudget` | number | `4000` | Characters spent on content-bearing entries per kind |
+| `continualHarness.detailBudget` | number | unset | Optional ceiling on the characters spent on content-bearing entries per kind |
 | `continualHarness.maxListedEntries` | number | `200` | One-line stubs rendered per kind after the content-bearing entries |
 
-`detailBudget` does not bind at the default `maxEntriesPerKind`; it is the ceiling that keeps a
-raised `maxEntriesPerKind` from dominating the prompt. The top-ranked entry always keeps its content
-regardless of the budget. Entries beyond `maxListedEntries` collapse into a `+N more <kind> entries`
+`maxEntriesPerKind` decides how many entries keep their content; a store with that many entries or
+fewer renders every one of them in full. `detailBudget` is unset by default and is an opt-in rail
+for operators who raise `maxEntriesPerKind` a long way: when it is set, entries that do not fit drop
+to a stub instead, except the top-ranked one, which always keeps its content.
+
+Stub lines are clipped to 160 characters each, so the stub menu costs at most
+`maxListedEntries * 161` characters per kind - 32,200 at the default, whatever the stored titles,
+paths, and ids look like. Entries beyond `maxListedEntries` collapse into a `+N more <kind> entries`
 count; setting it to `0` turns the stub list off entirely and restores the count-only overview.
+
+Values that are not finite positive numbers fall back to the defaults above rather than reaching the
+prompt.
 
 ```json
 {
   "continualHarness": {
     "maxEntriesPerKind": 6,
     "maxContentLength": 180,
-    "detailBudget": 4000,
     "maxListedEntries": 200
   }
 }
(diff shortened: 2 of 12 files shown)


kevinjosethomas:

PR #1128, opened 2026-08-10
fix(coding-agent): preserve saved cwd on daemon resume

### LINKED ISSUE #1124: Resumed sessions run in the daemon's launch directory: the supervisor re-injects `defaultSessionConfig.cwd` into a create whose `cwd` was deliberately stripped
## Summary

After `prime-agent shutdown`, reactivating any saved session from the TUI runs it in **the directory the daemon was launched from** instead of the session's own recorded directory. The session picker still shows the correct path, because the picker reads the session file header while the runtime is bound to a different directory. Tools, `%%bash`, and the IPython kernel all operate in the wrong project.

This is silent: nothing warns, and the UI actively suggests everything is fine. For anyone working across several checkouts or worktrees, a resumed session writes into the wrong repository.

## Reproduction

1. Start `prime-agent` in project **A**; create a session.
2. Start a session in project **B** (its header records `cwd: B`).
3. `prime-agent shutdown`.
4. Start `prime-agent` again **from A**.
5. Open the session from **B** in the session view. The row shows `B`.
6. Run `pwd` (or any file tool).

**Expected:** `B` — the session's recorded directory.
**Actual:** `A` — the directory the daemon was launched from.

Observed in normal use: after a shutdown/restart cycle, **many resumed sessions at once** were working in the wrong directory. That batch behaviour is consistent with the mechanism below — after a shutdown no session is resident, so every reactivation takes the same path and inherits the same single wrong directory.

Step 3 matters: while a worker for that session is still resident, `createOrReuseWorker` reuses it and the correct cwd survives. After a shutdown there is no worker, so every reactivation takes the `launchWorker` path below.

## Root cause

The create protocol encodes "use the session's own stored directory" as **the absence of `config.cwd`**. The supervisor destroys that signal by merging its defaults back in before forwarding to the worker.

**1. The client deliberately strips `cwd`** — `modes/agents-view/agents-view-mode.ts:194-206`:

```ts
// Stripping cwd opens the session in its own stored directory; overrideCwd is
// sent when that directory no longer exists so the daemon doesn't reject it.
export function createAgentsViewResumeConfig(config, overrideCwd) {
	const resumeConfig = { ...config };
	if (overrideCwd) resumeConfig.cwd = overrideCwd;
	else delete resumeConfig.cwd;      // "use the session's own cwd"
	return resumeConfig;
}
```

**2. The supervisor re-injects it** — `modes/daemon/daemon-supervisor.ts:2100-2103`:

```ts
const createCommand: DaemonCreateCommand = {
	...withoutSupervisorCreateFields(command),
	config: mergeAgentSessionRuntimeConfig(this.defaultSessionConfig, command.config),
};
```

`this.defaultSessionConfig.cwd` is the directory the daemon process was started in (`cli/daemon-launch.ts:375` spawns it with `cwd: spawnCwd ?? process.cwd()`). The worker is then also spawned there — `daemon-supervisor.ts:2118`:

```ts
cwd: createCommand.config?.cwd ?? process.cwd(),
```

**3. The worker reads the re-injected default as an explicit override** — `modes/daemon/daemon-mode.ts:1321` and `:1440`:

```ts
const cwdOverride = command.config?.cwd ? resolve(command.config.cwd) : undefined;
...
sessionManager = await SessionManager.openAsync(sessionPath, config.sessionDir, cwdOverride);
```

`cwdOverride` is now truthy, so `openAsync` uses it **instead of the session header's `cwd`**. Everything downstream binds to `sessionManager.getCwd()`, so services, tools, and the kernel all land in the daemon's launch directory.

The picker is unaffected because it renders `summary.cwd` read from the session file — which is why the path looks right while execution is wrong.

## Why tests did not catch it

Both sides of the interface are covered; the seam between them is not, and the fixtures make the defect unobservable.

- `test/agents-view-missing-cwd.test.ts` verifies the stripping contract by calling `SessionManager.openAsync` **directly**, with no supervisor in the path.
- `test/daemon-mode.test.ts` uses `defaultSessionConfig: { cwd: "/tmp" }` together with `config: { cwd: "/tmp" }`. When the default equals the session's directory, "use the default" and "use the session's cwd" give the same answer, so the override semantics cannot fail.
- `test/daemon-supervisor-process.test.ts` passes `config: { cwd: projectDir, ... }` explicitly on **every** create — the omitted-`cwd` case the TUI actually sends is never constructed.

The missing case is a single test: default cwd = `A`, session header cwd = `B`, client omits `cwd`, assert the session runs in `B`.

## Suggested fix

Preserve the absence of `cwd` when forwarding a `create` that targets an existing session, so the worker's override check keeps its meaning:

```ts
const mergedConfig = mergeAgentSessionRuntimeConfig(this.defaultSessionConfig, command.config);
if (command.config?.cwd === undefined && command.sessionPath) {
	delete mergedConfig.cwd;
}
```

The process-spawn `cwd` at `daemon-supervisor.ts:2118` should likewise not double as an override signal.

More robustly: representing "use the session's own directory" as a *missing field* in an object that is later merged with defaults is fragile by construction — merging exists precisely to fill in absent fields, so any merge along the path silently revives the wrong value. An explicit `cwdOverride` field on the create command (or a `cwd: "session"` sentinel) would make this unrepresentable. Note that #276/#303 fixed the same architectural pattern for client env vars, where the daemon served every session the identity of whichever client spawned it first.

## Related

- #287 — touches `createAgentsViewResumeConfig`, but only the missing-directory fallback.
- #1039 — fixes cwd loss during archived-session *revival* (client-side `reviveConfig`); different path, does not touch the supervisor merge.
- #1025 — "tools and kernel run in the wrong directory" for newly spawned `rlm()` children; same symptom, different cause.
- #681 — asks for `/cd` because "a session starts in the wrong directory"; likely a symptom of this.

## Environment

- prime-agent 0.7.1 (verified against `main` @ `a18809e`)
- macOS (arm64)

aoright:
## Summary

- Preserve an omitted `cwd` when the supervisor forwards a saved-session create to its worker, so `SessionManager.openAsync` uses the cwd recorded in the session header.
- Keep the merged cwd separately for the worker process launch, preserving existing spawn behavior without turning it into a session override.
- Add a process-level A/B regression that starts the daemon in project A, resumes a persisted session from project B, and asserts the worker reports B.

This is a backward-compatible behavior fix; the daemon wire shape and schema are unchanged.

## Testing

- `npx tsx ../../node_modules/vitest/dist/cli.js --run test/suite/regressions/1124-resume-recorded-cwd.test.ts`
- `npm run check`

Fixes #1124

> [!NOTE]
> ### Fix resumed daemon sessions to use saved working directory instead of daemon launch directory
> When resuming a session from a `sessionPath`, the spawned worker process was starting in the daemon's launch directory instead of the session's recorded `cwd`. The fix introduces `prepareWorkerLaunch` in [`daemon-supervisor.ts`](https://github.com/PrimeIntellect-ai/prime-agent/pull/1128/files#diff-18f2d5ec391785bebc6ddf3c1bd3b795e4ed6e02d0aa3852f5bf247d61d012f8), which strips `cwd` from the config for resumed sessions and passes it directly to the spawn options instead. A regression test in [`1124-resume-recorded-cwd.test.ts`](https://github.com/PrimeIntellect-ai/prime-agent/pull/1128/files#diff-5ba33e711e86b9739b321b1d66439f7d386372e6bcf01ddc1ab07b00a8924022) verifies the correct directory is used.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 88eb3be.</sup>
> 
>

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -5,6 +5,7 @@
 - Added privacy-safe pseudonymous product analytics for onboarding, command use, execution modes, run outcomes, TTFT, latency, usage, tools, retries, and compactions, with disclosure and opt-out controls ([ENG-4682](https://linear.app/primeintellect/issue/ENG-4682/add-privacy-safe-posthog-analytics-to-prime-agent)).
 - Changed sent agent messages in the IPython cell UI to show only the message text with a `╰─` gutter when expanded, matching received messages, and hid the raw `agent_message.send` receipt dictionary.
 - Fixed Homebrew installs attempting to self-update their versioned Cellar keg instead of directing users to `brew upgrade prime-agent` ([#844](https://github.com/PrimeIntellect-ai/prime-agent/issues/844))
+- Fixed resumed sessions using the daemon launch directory instead of their recorded working directory ([#1124](https://github.com/PrimeIntellect-ai/prime-agent/issues/1124)).
 
 ## [0.7.1] - 2026-08-07
 
--- a/packages/coding-agent/src/modes/daemon/daemon-supervisor.ts
+++ b/packages/coding-agent/src/modes/daemon/daemon-supervisor.ts
@@ -389,6 +389,24 @@ function withoutSupervisorCreateFields(command: DaemonCreateCommand): DaemonCrea
 	return workerCommand;
 }
 
+function prepareWorkerLaunch(
+	defaultSessionConfig: AgentSessionRuntimeConfig,
+	command: DaemonCreateCommand,
+): { createCommand: DaemonCreateCommand; processCwd: string | undefined } {
+	const config = mergeAgentSessionRuntimeConfig(defaultSessionConfig, command.config);
+	const processCwd = config.cwd;
+	if (command.sessionPath && command.config?.cwd === undefined) {
+		delete config.cwd;
+	}
+	return {
+		createCommand: {
+			...withoutSupervisorCreateFields(command),
+			config,
+		},
+		processCwd,
+	};
+}
+
 function responseWithId(response: DaemonResponse, id: string | undefined): DaemonResponse {
 	return { ...response, id };
 }
@@ -2098,10 +2116,7 @@ export class DaemonSupervisor {
 		const recoveryStopRevision = existing?.stopRevision;
 		const launchEnv =
 			ownerClientId || existing?.descriptor.ownerClientId ? (command.launchEnv ?? existing?.launchEnv) : undefined;
-		const createCommand: DaemonCreateCommand = {
-			...withoutSupervisorCreateFields(command),
-			config: mergeAgentSessionRuntimeConfig(this.defaultSessionConfig, command.config),
-		};
+		const { createCommand, processCwd } = prepareWorkerLaunch(this.defaultSessionConfig, command);
 		const workerId = existing?.descriptor.workerId ?? createActiveSessionId();
 		const rootActiveSessionId = existing?.descriptor.rootActiveSessionId ?? createActiveSessionId();
 		const socketPath = existing?.descriptor.socketPath ?? workerSocketPath(this.socketPath, workerId);
@@ -2115,7 +2130,7 @@ export class DaemonSupervisor {
 		const launch = createCliSubprocessLaunchSpec(["--mode", "daemon", "--daemon-socket", socketPath]);
 		await this.assertRecoveryAllowed();
 		const child: ChildProcess = spawn(launch.command, launch.args, {
-			cwd: createCommand.config?.cwd ?? process.cwd(),
+			cwd: processCwd ?? process.cwd(),
 			detached: true,
 			env: createCliSubprocessEnv({
 				...process.env,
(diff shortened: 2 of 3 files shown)


MingTeer (approved):
Verified the fix end-to-end on our fork (17 commits behind upstream/main). Traced the full path and it closes the loop:

- Bug: the supervisor merges `defaultSessionConfig.cwd` into the forwarded `command.config`, so the worker's `cwdOverride = command.config?.cwd` is always set, and `SessionManager.openAsync(sessionPath, dir, cwdOverride)` lets it override the header cwd. Resumed sessions silently run in the daemon launch directory.
- Fix: deleting `config.cwd` when `sessionPath` is set and the client sent no explicit cwd makes `cwdOverride` undefined, so the header cwd wins; runtime cwd then comes from `sessionManager.getCwd()`. The worker-side `if (!config.cwd) throw` is unaffected since the worker re-merges with its own defaults. `mergeAgentSessionRuntimeConfig` always returns a fresh object, so the `delete` can't pollute shared config. Scope is correctly limited to resume-without-explicit-cwd; new sessions and `--cwd` behave as before.
- A/B check: the regression test fails against the pre-fix `daemon-supervisor.ts` and passes with the fix.

One non-blocking note on the test: it spreads `...process.env` into the spawned supervisor. When run from inside a prime-agent session (e.g. via the IPython tool or a tmux pane), `PRIME_AGENT_INTERNAL_DAEMON_WORKER` and friends leak in, the spawned process identifies as a worker instead of a supervisor, and the handshake times out. Scrubbing `PRIME_AGENT_INTERNAL_*` from the test env would make it more portable.

We cherry-picked this onto our fork because the silent wrong-directory behavior writes into the wrong repository. Hope it lands upstream.

kevinjosethomas:

PR #1139, opened 2026-08-10
feat(coding-agent): add a bundled firecrawl skill (web, developer and research indexes)

rakshith48:
## What

Adds a bundled `firecrawl` Python skill: web search that can **read** the pages it finds, plus Firecrawl's developer and research indexes. It sits alongside `websearch` rather than replacing it.

```python
await firecrawl("query")                      # web search
await firecrawl("query", fetch_content=True)  # search + page markdown
await firecrawl.scrape(url)                   # one page as markdown

await firecrawl.developer("TypeError: ...")   # issues, PRs, READMEs, docs

await firecrawl.research("query")             # semantic search over abstracts
await firecrawl.inspect_paper(id)             # canonical metadata
await firecrawl.related_papers(id, intent)    # similar / citers / references
await firecrawl.read_paper(id, question)      # in-body passages
```

## Why

The harness has exactly one tool (`ipython`), and the only bundled web capability is Serper search, which returns snippets and URLs. Following up on a result means shelling out to `curl` and parsing HTML.

The **developer index** is the piece that matters most for a coding agent: it answers an error string from the issue that reported it or the PR that fixed it, and returns the matched passages, so the usual search → scrape → hope round trip collapses into one call. The **research index** fits the evaluation and literature work Prime Agent is aimed at.

## Design notes

- **REST, not the SDK.** `firecrawl-py`'s import name is `firecrawl`, which the skill package itself would shadow in the kernel venv. The skill calls the v2 REST API over `httpx`, already in the kernel venv, so it adds no new dependency.
- **No console script.** A skill's script name must match its import name exactly, and a `firecrawl` binary (the Firecrawl CLI) is already common on developer machines — declaring one would shadow it, or be shadowed by it, in `%%bash` cells depending on PATH order. The kernel API is the supported entry point.
- **Output budgets.** Every entry point truncates: 8 KB search and paper listings, 12 KB developer results and passages, 20 KB scrape. Paper listings show a 500-char abstract preview and point at `inspect_paper`/`read_paper`, because ten full abstracts overflow the cap and mid-elision silently drops the later results.
- **Credentials mirror Serper.** Paste the key via `/login` → **MCP Connections** → "Firecrawl (web search + scrape)"; it is injected into the kernel only when a `firecrawl` skill is actually loaded. `websearch-credential.ts` gains a `BUNDLED_SKILL_CREDENTIALS` table that the env injection and both `/login` selectors iterate — the previous per-credential chain of early returns would have skipped the second credential outright. Happy to rename that file if you'd prefer it match its new contents; I left the name alone to keep the diff small.
- `FIRECRAWL_API_URL` points the skill at a self-hosted instance; `bundledSkills.firecrawl: false` disables it.
- **No `/map`.** Deliberately not wrapped — URL discovery is better served by `developer` for code questions and by search for everything else.

## Verification

Exercised against the live API through a real IPython kernel booted by the `ipython` tool, with the skill installed into a throwaway kernel venv under an isolated HOME: search, hydrated search, scrape, all five index calls, `help()` signature, the missing-key message, and the absence of a console script. All pass. The script that drives this needs a paid API key so it can't run in CI; happy to include it if you want it in the repo.

Also covered: 5 new unit tests (bundled-skill loading and disabling, per-skill credential injection including the "don't leak the other key" case), `biome`, `tsgo --noEmit`, and the pre-commit hook. Rebased on current `main`.

One thing I hit that may be worth a separate fix: 6 tests fail on unmodified `main` on my machine — 4 extension tests and 2 kernel tests — because they read `~/.agents/skills` and other real user directories rather than an isolated HOME, so anything the developer has installed locally leaks into assertions. Unrelated to this change, but it's why the two new resource-loader tests pin `HOME` to a temp dir.

> [!NOTE]
> ### Add bundled Firecrawl skill for web search, scraping, and research index queries
> - Adds a new built-in `firecrawl` skill ([api.py](https://github.com/PrimeIntellect-ai/prime-agent/pull/1139/files#diff-844d85a065c30181f53ae99b569e942f9528ef59945bdd31344c827d4264505f)) exposing async functions: `run` (web search), `scrape`, `developer` (GitHub/docs index), `research`, `related_papers`, `inspect_paper`, and `read_paper`.
> - API key resolution checks the `FIRECRAWL_API_KEY` env var first, then falls back to `auth.json`; the key is injected into the kernel environment only when the firecrawl skill is loaded.
> - The login/logout UI in interactive mode now lists Firecrawl alongside Serper under MCP Connections, using the generalized `BUNDLED_SKILL_CREDENTIALS` registry in [websearch-credential.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/1139/files#diff-b48778e89f037156f6c5da1809e2fe8a087bb65db1d4125fd8caef2d3878b3f9).
> - The skill is enabled by default and can be disabled via `bundledSkills.firecrawl: false` in settings.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 6112907.</sup>
> 
>


kevinjosethomas:

PR #1149, opened 2026-08-10
fix(coding-agent): recover supervisor ownership when its registry entry is pruned

### LINKED ISSUE #1148: Daemon supervisor permanently fails with supervisor_generation_stale after the OS prunes its registry entry from the temp directory
On macOS, a daemon supervisor that has been running for about three days starts rejecting every command with `supervisor_generation_stale` and cannot recover for the rest of its process lifetime. Restarting the daemon is the only fix, which drops all resident sessions.

## Cause

The supervisor registry defaults to a subdirectory of the daemon socket directory, which lives under `os.tmpdir()`. On macOS that is the per-user `/var/folders/.../T` directory, which the OS prunes of entries that have not been touched for roughly three days (`daemon-supervisor-ownership.ts:250`):

```ts
return environment[DAEMON_SUPERVISOR_REGISTRY_DIR_ENV] ?? resolve(defaultDaemonSocketDir(), "supervisor-owners");
```

`owner.json` is only written when ownership is acquired (`:332`) and on a phase change (`:298`). Nothing refreshes it afterwards, so a supervisor that stays up long enough has its own registry entry deleted underneath it while it is perfectly healthy.

`assertCurrent()` then treats a missing record identically to a record owned by someone else (`:130-137`):

```ts
const current = readOwnerRecord(this.ownerDirectory);
if (!current || !sameOwnerRecord(current, this.record)) {
	throw new DaemonSupervisorOwnershipLostError(this.record.generation);
}
```

This runs for every journaled command, so every subsequent `create` fails. The state is unrecoverable in-process: `sameOwnerRecord` compares a `token` that only ever existed in the deleted file and in the supervisor's memory, so no external repair can reconstruct a matching record.

0.4.0 fixed the startup side of this ("Fixed daemon startup after an interrupted supervisor leaves an empty ownership directory") through `readOwnerRecordForScope`, but that self-heal only runs when a new supervisor tries to acquire ownership. It never runs for the live process.

## Evidence

Supervisor `2bcd9be6-3ee5-4612-8436-3c6ce7ee3799`, up since Aug 6 10:29, registry at `/var/folders/.../T/prime-agent-501/supervisor-owners/`:

- `2bcd9be6-....owner/` still exists but is empty. Both `owner.json` and `scope.json` are gone. Directory mtime is Aug 10 03:45.
- The same 03:45 sweep deleted `connection.json` from two `prime-agent-kernel-*` temp directories created Aug 6 and Aug 7. The directory created Aug 8, under three days old, was untouched.
- The supervisor logged no error. Its last entry is a routine idle-child passivation, and the process is still listening on `daemon.sock`.
- Every `prime-agent` launch fails in `createDaemonClientConnection` with `Daemon supervisor generation 2bcd9be6-... no longer owns its registry entry`.

## Expected

A missing record should not be treated as a takeover. The live supervisor should restore its own entry under the registry guard, after confirming no live supervisor has claimed its socket or descriptor directory, and should refresh the record periodically so it does not age out of the temp directory in the first place.

Related: #879 is the same failure surface from a different trigger, where `processStartId` derived from `ps -o lstart=` changes with the system timezone.

## Environment

prime-agent 0.7.0, also present on `main` at ebfe770e. macOS 26.5.2 (Darwin 25.5.0), Node 25.2.1.

gbusto:
Fixes #1148.

The supervisor registry lives under `os.tmpdir()`, and `owner.json` was only written on acquire and on a phase change. macOS prunes `/var/folders` entries untouched for about three days, so a healthy long-lived supervisor loses its own registry entry. `assertCurrent` treated that missing record exactly like a takeover, so every journaled command failed with `supervisor_generation_stale` for the rest of the process lifetime. Nothing could repair it from outside either, because `sameOwnerRecord` compares a token that existed only in the deleted file and in the supervisor's memory.

## Changes

- `assertCurrent` now distinguishes a vanished record from a stolen one. A record belonging to another generation is still fatal. A missing record is restored under `withDaemonSupervisorRegistryGuard`, re-reading inside the lock, and only after scanning for a live supervisor that has claimed the same socket or descriptor directory. If one exists, ownership really is gone and the error is still raised, so two supervisors can never both consider themselves the owner.
- Added an hourly `unref`'d refresh that rewrites the record, keeping its mtime well inside the pruning window so the entry does not age out to begin with. Cleared on `release()`.

Both halves matter: without the recovery path an already-broken daemon stays broken, and without the refresh a healthy daemon still stalls once every three days before healing itself.

`0.4.0` fixed the startup side of this through `readOwnerRecordForScope`, but that self-heal only runs when a new supervisor acquires ownership, never for the live process.

## Testing

`packages/coding-agent/test/suite/regressions/1148-supervisor-registry-entry-pruned.test.ts` covers four cases: recovery when `owner.json` is deleted, recovery when the whole owner directory is removed, and the two paths that must still fail, namely a foreign record in our slot and a live supervisor holding our socket.

Reverting only the source change fails exactly the two recovery cases and leaves the two ownership-lost cases passing, so the test pins the bug without loosening the fatal path.

`npm run check` is clean. The new test passes, as do `4600-supervisor-singleton`, `daemon-supervisor-admission`, and `daemon-supervisor-monitor` (70 tests). I did not run `daemon-supervisor-process`, which is excluded from `test:ci`.

## Note

`assertDaemonSupervisorOwnerCurrent` has the same missing-record-is-fatal logic on the worker validation path. It recovers indirectly once the supervisor rewrites the record, so I left it alone to keep this diff focused. Happy to include it if you would prefer.

> [!NOTE]
> ### Fix supervisor ownership recovery when OS prunes its registry entry from temp directory
> - `assertCurrent()` in [`DaemonSupervisorOwnership`](https://github.com/PrimeIntellect-ai/prime-agent/pull/1149/files#diff-6f9ef6ce6a9d2bc51c61fe94d59310efffd26d8f7b3966f5355606b2847d06a5) now treats a missing owner record as recoverable: if the current token matches, it rewrites the scope and owner records instead of throwing `supervisor_generation_stale`.
> - A periodic refresh (every ~1 hour) rewrites the owner record proactively to reduce the chance of OS temp pruning causing issues for long-running daemons.
> - `release()` cancels and awaits any in-flight refresh before performing the cleanup rename-and-delete sequence.
> - Adds a [regression test suite](https://github.com/PrimeIntellect-ai/prime-agent/pull/1149/files#diff-f766d4a6ec5b2be9bbd21bbde7c44774729b7a25a1fe40a4c12cf3b8af009079) covering recovery from deleted files, deleted directories, token conflicts, and competing live supervisors.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized ecde15f.</sup>
> 
>

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -8,6 +8,7 @@
 - Fixed Homebrew installs attempting to self-update their versioned Cellar keg instead of directing users to `brew upgrade prime-agent` ([#844](https://github.com/PrimeIntellect-ai/prime-agent/issues/844))
 - Fixed the agents view collapsing expanded subagent lists when returning from an opened agent ([ENG-5105](https://linear.app/primeintellect/issue/ENG-5105/keep-the-agents-view-state-persistent)).
 - Kept the subagent summary row visible and selectable while its list is expanded in the agents view, so pressing enter on it collapses the list again ([ENG-5105](https://linear.app/primeintellect/issue/ENG-5105/keep-the-agents-view-state-persistent)).
+- Fixed a long-running daemon rejecting every command with `supervisor_generation_stale` after the OS pruned its registry entry from the temp directory ([#1148](https://github.com/PrimeIntellect-ai/prime-agent/issues/1148)).
 
 ## [0.7.1] - 2026-08-07
 
--- a/packages/coding-agent/src/modes/daemon/daemon-supervisor-ownership.ts
+++ b/packages/coding-agent/src/modes/daemon/daemon-supervisor-ownership.ts
@@ -22,6 +22,10 @@ const REGISTRY_LOCK_UPDATE_MS = 1000;
 const REGISTRY_LOCK_RETRIES = 500;
 const REGISTRY_LOCK_RETRY_MS = 10;
 const STARTUP_FENCE_POLL_MS = 250;
+// The registry lives under the OS temp directory, which macOS prunes of entries that
+// have not been touched for ~3 days. Rewrite our own entry well inside that window so
+// a long-lived supervisor never ages out of its own registry.
+const OWNER_RECORD_REFRESH_MS = 60 * 60 * 1000;
 const SHUTDOWN_ADMISSION_FILE_NAME = "shutdown-admission.json";
 const SHUTDOWN_ADMISSION_LEASE_MS = 5000;
 const SHUTDOWN_ADMISSION_REFRESH_MS = 1000;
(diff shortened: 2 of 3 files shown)


sethkarten:
Thank you for the report and proposed work. This root cause is now covered by maintainer-owned stacked PR #1161, authored independently from `upstream/main`.

We did not inspect or reuse this PR's diff, branch, commits, implementation code, or tests; its public description/comments were used only as a bug report. To keep one review surface, this PR is superseded by #1161 and is being closed.

PR #1160, opened 2026-08-10
[stack 3/8] fix(coding-agent): make persisted state crash-safe

sethkarten:
# Stack 3/8 — fix(coding-agent): make persisted state crash-safe

> **Active review snapshot — do not merge yet.** The complete stack is open for architecture/design review, while final cumulative audit, CI, Cursor Bug Bot, and Macroscope findings are being remediated. Branches will be force-updated after validation.

**Base:** `stack/external-02-security`  
**Review order:** merge only after the preceding stack layer is accepted. This PR is not intended to merge independently out of order.

## Stack navigation

1. [#1158](https://github.com/PrimeIntellect-ai/prime-agent/pull/1158) — ci: harden verification and release compatibility
2. [#1159](https://github.com/PrimeIntellect-ai/prime-agent/pull/1159) — fix(security): harden session and autonomous execution boundaries
3. [#1160](https://github.com/PrimeIntellect-ai/prime-agent/pull/1160) — fix(coding-agent): make persisted state crash-safe
4. [#1161](https://github.com/PrimeIntellect-ai/prime-agent/pull/1161) — fix(daemon): fence worker and supervisor lifecycle state
5. [#1162](https://github.com/PrimeIntellect-ai/prime-agent/pull/1162) — fix(coding-agent): repair queued and archived session lifecycle
6. [#1163](https://github.com/PrimeIntellect-ai/prime-agent/pull/1163) — fix(coding-agent): complete Windows kernel and daemon startup
7. [#1164](https://github.com/PrimeIntellect-ai/prime-agent/pull/1164) — fix(providers): harden MCP OAuth and Codex transports
8. [#1165](https://github.com/PrimeIntellect-ai/prime-agent/pull/1165) — fix(runtime): bound transcript and autonomous recovery

## Summary

- Repair torn JSONL tails without losing concurrent appends, with bounded manual recovery for oversized tails.
- Serialize session mutations across processes and recover safely from dead lock owners.
- Make settings, auth, and migration writes private, atomic, durable on POSIX, and restart-idempotent.

## Validation

- `npm run check`; persistence/session/config focused suites; final crash-lock duration suite pending remediation recheck.
- Residual/non-blocking: Windows crash-injection not run; POSIX directory fsync skipped on Windows

## Provenance

- Authored independently from `upstream/main` using issue reports and PR descriptions/comments only.
- No external contributor branch, diff, commit, implementation code, or test code was fetched, inspected, copied, or reused.
[...]

PR #1159, opened 2026-08-10
[stack 2/8] fix(security): harden session and autonomous execution boundaries

sethkarten:
# Stack 2/8 — fix(security): harden session and autonomous execution boundaries

> **Active review snapshot — do not merge yet.** The complete stack is open for architecture/design review, while final cumulative audit, CI, Cursor Bug Bot, and Macroscope findings are being remediated. Branches will be force-updated after validation.

**Base:** `stack/external-01-verification`  
**Review order:** merge only after the preceding stack layer is accepted. This PR is not intended to merge independently out of order.

## Stack navigation

1. [#1158](https://github.com/PrimeIntellect-ai/prime-agent/pull/1158) — ci: harden verification and release compatibility
2. [#1159](https://github.com/PrimeIntellect-ai/prime-agent/pull/1159) — fix(security): harden session and autonomous execution boundaries
3. [#1160](https://github.com/PrimeIntellect-ai/prime-agent/pull/1160) — fix(coding-agent): make persisted state crash-safe
4. [#1161](https://github.com/PrimeIntellect-ai/prime-agent/pull/1161) — fix(daemon): fence worker and supervisor lifecycle state
5. [#1162](https://github.com/PrimeIntellect-ai/prime-agent/pull/1162) — fix(coding-agent): repair queued and archived session lifecycle
6. [#1163](https://github.com/PrimeIntellect-ai/prime-agent/pull/1163) — fix(coding-agent): complete Windows kernel and daemon startup
7. [#1164](https://github.com/PrimeIntellect-ai/prime-agent/pull/1164) — fix(providers): harden MCP OAuth and Codex transports
8. [#1165](https://github.com/PrimeIntellect-ai/prime-agent/pull/1165) — fix(runtime): bound transcript and autonomous recovery

## Summary

- Contain session, artifact, snapshot, auth, editor, export, harness, and debug-file paths with private atomic storage.
- Cancel admitted RLM work when its kernel host is disposed and bound clipboard fallback behavior.
- Keep `--no-session` descendants in memory and document the actual autonomous-execution isolation boundary.
- Pin workflow actions by digest and disable persisted checkout credentials.

## Validation

- `npm run check`; focused Vitest 184/184; Python harness 36/36; workflow digest and documentation-link validation.
- Residual/non-blocking: dill still executable for valid snapshots; cooperative cancellation; broader sensitive sinks; no full built-in sandbox

## Provenance

- Authored independently from `upstream/main` using issue reports and PR descriptions/comments only.
[...]

PR #1158, opened 2026-08-10
[stack 1/8] ci: harden verification and release compatibility

sethkarten:
# Stack 1/8 — ci: harden verification and release compatibility

> **Active review snapshot — do not merge yet.** The complete stack is open for architecture/design review, while final cumulative audit, CI, Cursor Bug Bot, and Macroscope findings are being remediated. Branches will be force-updated after validation.

**Base:** `main`  
**Review order:** merge only after the preceding stack layer is accepted. This PR is not intended to merge independently out of order.

## Stack navigation

1. [#1158](https://github.com/PrimeIntellect-ai/prime-agent/pull/1158) — ci: harden verification and release compatibility
2. [#1159](https://github.com/PrimeIntellect-ai/prime-agent/pull/1159) — fix(security): harden session and autonomous execution boundaries
3. [#1160](https://github.com/PrimeIntellect-ai/prime-agent/pull/1160) — fix(coding-agent): make persisted state crash-safe
4. [#1161](https://github.com/PrimeIntellect-ai/prime-agent/pull/1161) — fix(daemon): fence worker and supervisor lifecycle state
5. [#1162](https://github.com/PrimeIntellect-ai/prime-agent/pull/1162) — fix(coding-agent): repair queued and archived session lifecycle
6. [#1163](https://github.com/PrimeIntellect-ai/prime-agent/pull/1163) — fix(coding-agent): complete Windows kernel and daemon startup
7. [#1164](https://github.com/PrimeIntellect-ai/prime-agent/pull/1164) — fix(providers): harden MCP OAuth and Codex transports
8. [#1165](https://github.com/PrimeIntellect-ai/prime-agent/pull/1165) — fix(runtime): bound transcript and autonomous recovery

## Summary

- Gate PRs and releases on the locked Python runtime suite across Python 3.10–3.14.
- Verify the complete emitted node-only provider bundle closure, including Bedrock.
- Support npm 12 remote-dependency policy with accurate interactive consent and a real local-HTTP regression.

## Validation

- `npm run check`; targeted installer/release suites; real npm 12 policy smoke; Python 3.10–3.14 locked runtime suites; release pack smoke.
- Residual/non-blocking: npm12 transitive scripts warnings; broader #773 TUI/update routing not included

## Provenance

- Authored independently from `upstream/main` using issue reports and PR descriptions/comments only.
- No external contributor branch, diff, commit, implementation code, or test code was fetched, inspected, copied, or reused.
- The implementation and regression tests in this stack are maintainer-owned.

[...]

PR #1006, opened 2026-08-08
fix(coding-agent): acquire the bootstrap lock with an atomic rename

Adolanium:
Fixes #1005.

## What was broken

`acquireBootstrapLock` reclaimed a stale lock with `rm(lockDir)` and then retried `mkdir`. Two processes can both judge the same lock stale, and one process's `rm` can delete a lock directory the other just created, with the loser's pid write landing inside the winner's directory. Both then return believing they hold the lock, so two venv rebuilds run concurrently and can corrupt the environment. Fresh installs and fleet boots are where several processes start together.

## The fix

Adopts the pattern the daemon launch lock and session leases already use: create a uniquely named candidate directory, write the pid file inside it, and atomically rename it onto the lock path (rename fails if the target exists). Stale reclamation renames the lock aside before deleting it, so a live lock created in the interim is never touched. One wrinkle: Windows fails a directory rename onto an existing target with `EPERM` rather than `EEXIST`, so that is treated as a lock-held signal too.

Signature and staleness semantics are unchanged.

## Testing

- New regression file `test/suite/regressions/1005-bootstrap-lock.test.ts`: concurrent acquires serialize, a dead-pid lock is reclaimed with no leftover candidate or stale dirs, and a live lock injected between the stale rename-aside and the delete survives untouched.
- `npx tsx ../../node_modules/vitest/dist/cli.js --run test/suite/regressions/1005-bootstrap-lock.test.ts test/kernel-bootstrap.test.ts`: the 3 new tests pass; the 17 kernel-bootstrap failures reproduce identically on unmodified main (the suite's fakes are POSIX-only shell scripts that cannot spawn on Windows).
- `npm run check` clean.

> [!NOTE]
> ### Fix `acquireBootstrapLock` to use atomic rename preventing concurrent venv rebuilds
> - Rewrites lock acquisition in [`bootstrap.ts`](https://github.com/PrimeIntellect-ai/prime-agent/pull/1006/files#diff-e87b9a04ab9b4ac02e8dd33af2f2f58320d8b7f6be6853acf4a3e05a34623cae) to use an atomic directory rename: creates a UUID+PID candidate directory, writes a PID file, then renames it to the lock path atomically.
> - Fixes a race where stale lock reclamation could delete a newly created lock from another process; now renames the stale lock aside before removing it, ignoring `ENOENT` if it disappears.
[...]

PR #1161, opened 2026-08-10
[stack 4/8] fix(daemon): fence worker and supervisor lifecycle state

### LINKED ISSUE #1045: One failed worker descriptor poisons global heartbeats_list; stopWorker EPERM/pid-reuse false-alive marks exited workers "failed"
## Summary

A single worker descriptor stuck at `lifecycle: "failed"` makes the **global** `heartbeats_list` command fail entirely with:

```
Error: Cannot list heartbeats while session worker is failed
```

Observed on v0.7.1 (macOS, Apple Silicon); both contributing defects are still present on `main` in `packages/coding-agent/src/modes/daemon/daemon-supervisor.ts`.

## Root cause chain (observed live)

1. An idle session worker received its eviction shutdown and logged `shutting down (exit 0)` — the process really exited (supervisor log timestamps 02:37:12Z).
2. `stopWorker` then polled `isWorkerProcessAlive()` → `isProcessAlive(pid)`, which treats `EPERM` from `kill(pid, 0)` as **alive**:

   ```ts
   function isProcessAlive(pid: number): boolean {
     try { process.kill(pid, 0); return true; }
     catch (error) { return (error as NodeJS.ErrnoException).code === "EPERM"; }
   }
   ```

   The worker's pid (320, a low pid) was held by/reused for a root-owned process after exit, so the probe returned `EPERM` → "alive". The descriptor **has** a recorded `processStartId`, and the recovery path checks it (`processIdentityMatches`), but the `stopWorker` liveness loop never consults it.
3. Result: SIGTERM grace expired → SIGKILL → still "alive" → `Error: Session worker <id> did not stop` (`Idle eviction sweep failed`). On the next supervisor start, the retried intentional stop hit the same false positive → `Could not complete intentional stop` → `lifecycle: "failed"` persisted with `lastError: "Session worker <id> did not stop after SIGKILL"`.
4. The user-visible amplification: the global `heartbeats_list` fan-out filters workers with `isVisibleWorker`, which only excludes client-owned workers — not failed ones. A failed worker with no client and no cached `heartbeatSnapshot` produces a failure snapshot, and the handler returns the **first failure for the whole command**:

   ```ts
   const failed = snapshots.find((snapshot) => snapshot.response)?.response;
   if (failed) { return failed; }
   ```

   So one zombie descriptor poisons every global heartbeat listing indefinitely, while the session-scoped path (`activeSessionId` set) keeps working — which makes the failure look intermittent/mysterious from the TUI.

## Reproduction sketch

1. Create a worker, record its pid, let it exit while `stopWorker` believes a pid-colliding process (or anything answering `EPERM`) is the worker.
2. Supervisor marks it `lifecycle: "failed"` after the SIGKILL deadline.
3. Send a global `heartbeats_list` (no `activeSessionId`) over the daemon socket → the whole command fails with `Cannot list heartbeats while session worker is failed`, regardless of how many healthy workers exist.

## Workaround (verified)

`retry_worker` for the failed worker's `rootSessionId` over the daemon socket clears the state without a daemon restart (once the phantom pid is genuinely gone): lifecycle → `recovering` → respawn → `ready`, and global `heartbeats_list` succeeds again.

## Suggested fixes

- `stopWorker`'s `isWorkerProcessAlive` should verify `processStartId` (as the recovery path already does) before trusting `kill(pid, 0)`/`EPERM`; a pid whose start-id no longer matches is dead for stop purposes. This is the same false-alive pattern as #868, in a second call site.
- `heartbeats_list` should not fail the entire aggregation on one failed worker: skip `lifecycle === "failed"` workers (they cannot host firing heartbeats) or annotate them in the response, instead of returning the first failure snapshot as the whole command's result.

### LINKED ISSUE #1131: Bug: Daemon startup permanently fails with Lock file is already being held after a crashed startup attempt leaves an orphaned lock directory
### Summary

If a `prime-agent` daemon-supervisor bootstrap process is killed/crashes partway through startup, it leaves behind an orphaned mkdir-based lock directory that is never cleaned up. Every subsequent `prime-agent` invocation then fails immediately with:

```
Error: Prime Agent daemon exited during startup (code 1).
Error: Lock file is already being held
    at ensureDaemonRunning (.../prime-agent/dist/bundle/chunk-VNU2AJHD.js:60733:3)
```

This happens even though a separate, fully healthy daemon process may already be running and actively serving sessions — the CLI does not distinguish "a stale lock from a dead process" from "a live daemon is already up," so it hard-fails instead of either cleaning up the stale lock or attaching to the existing daemon.

### Environment

- `prime-agent` version: `0.7.1`
- OS: macOS
- Node: `v25.8.0`
- Install: global npm (`/Users/.../\.npm-global/lib/node_modules/prime-agent`)

### Steps to reproduce (as observed)

1. A `prime-agent` startup attempt began at `07:36:xx` local time (short-lived process, ~0.5s CPU time) and appears to have crashed or been interrupted before it finished its bootstrap sequence.
2. It left behind an orphaned lock directory at:
   ```
   /var/folders/<hash>/T/prime-agent-501/.supervisor-launch-<workerid>.lock
   ```
   This directory is created via a `proper-lockfile`-style `mkdir()`-based lock (see below), which — unlike an flock-based lock — is **not** automatically released when the owning process dies.
3. Every subsequent `prime-agent` invocation (tried repeatedly over ~15+ minutes, in different terminal tabs, on different git branches) failed identically:
   ```
   supervisor: Daemon supervisor startup failed: Error: Lock file is already being held
       at ...chunk-VNU2AJHD.js:1637:43
       at callback (...chunk-VNU2AJHD.js:608:24)
       at FSReqCallback.oncomplete (node:fs:192:5)
   ```
   Retries happened automatically roughly every 10 seconds and never succeeded.
4. Meanwhile, a genuinely healthy daemon (a different, long-running `prime-agent` process, uptime since the prior Saturday, ~87 min CPU time) was still alive and actively serving multiple worker sockets in the same temp directory (`daemon.sock` and six `worker-*.sock` files). The new CLI invocations were not attaching to it — they were independently trying (and failing) to bootstrap a *new* supervisor.

### Root cause

In `chunk-VNU2AJHD.js` (bundled `proper-lockfile`-style locking):

```js
function getLockFile(file, options2) {
  return options2.lockfilePath || `${file}.lock`;
}
function acquireLock(file, options2, callback) {
  const lockfilePath = getLockFile(file, options2);
  options2.fs.mkdir(lockfilePath, (err) => {
    ...
    options2.fs.stat(lockfilePath, (err2, stat4) => {
      ...
      if (!isLockStale(stat4, options2)) {
        return callback(Object.assign(new Error("Lock file is already being held"), { code: "ELOCKED", file }));
      }
      ...
    });
  });
}
```

- The lock is a plain directory created with `mkdir`. If the process holding it dies without explicitly `rmdir`-ing it, the directory persists indefinitely on disk.
- Staleness is judged purely from the lock directory's `mtime` vs. a `stale` threshold. In this case the orphaned lock's `mtime` was old enough in wall-clock terms that it plausibly should have been considered stale, but `ensureDaemonRunning` (`chunk-VNU2AJHD.js:60733`) treated the `ELOCKED` failure as fatal rather than retrying with a stale-check override or falling back to "attach to existing daemon if one is reachable."
- There is no self-healing path: once this specific lock directory exists, `prime-agent` cannot start again until the directory is manually removed, regardless of whether the process that created it is still alive.

### Workaround (used to unblock ourselves)

Manually remove the orphaned lock directory (not the daemon socket itself):

```bash
rm -rf "/var/folders/<hash>/T/prime-agent-501/.supervisor-launch-<workerid>.lock"
```

After removal, `prime-agent --version` and a full `prime-agent` launch both succeeded immediately with no error. Note: `daemon.sock.lock` in the same directory is a *different*, legitimately-held lock (continuously re-acquired by the live daemon process) and should not be removed — deleting it and observing it immediately reappear was how we confirmed it was healthy rather than stale, as distinct from `.supervisor-launch-*.lock`, which did not reappear after deletion.

### Suggested fixes

1. **Clean up on crash**: wrap supervisor bootstrap in a `try/finally` (or register `process.on('exit'/'SIGINT'/'SIGTERM')` handlers) that always releases the `.supervisor-launch-*.lock` directory, even on unhandled exceptions.
2. **Distinguish stale-but-unheld locks from a live daemon**: before failing hard on `ELOCKED`, check whether a daemon is actually reachable on the corresponding socket (e.g., `daemon.sock`). If it is, skip supervisor bootstrap and attach to the existing daemon instead of erroring.
3. **Lower/verify the staleness threshold**: confirm the `stale` option passed to the lock actually reflects a reasonable bootstrap timeout (bootstrap should take seconds, not indefinitely persist as "not stale").
4. **Better error message**: when `ensureDaemonRunning` fails with `ELOCKED`, surface the lock file path and a hint (e.g., "If no other `prime-agent` process is starting up, delete this lock file and retry") instead of just the raw `mkdir`/`fs.stat` stack trace.

### Additional log excerpt

```
[2026-08-10T11:19:18.291Z] supervisor: Daemon supervisor startup failed: Error: Lock file is already being held
    at file:///.../prime-agent/dist/bundle/chunk-VNU2AJHD.js:1637:43
    at callback (file:///.../prime-agent/dist/bundle/chunk-VNU2AJHD.js:608:24)
    at FSReqCallback.oncomplete (node:fs:192:5)
```
(repeats every ~10s across multiple terminal tabs and git branches, ~07:19–07:36 local time, until the lock directory was manually removed)

### LINKED ISSUE #1148: Daemon supervisor permanently fails with supervisor_generation_stale after the OS prunes its registry entry from the temp directory

sethkarten:
# Stack 4/8 — fix(daemon): fence worker and supervisor lifecycle state

> **Active review snapshot — do not merge yet.** The complete stack is open for architecture/design review, while final cumulative audit, CI, Cursor Bug Bot, and Macroscope findings are being remediated. Branches will be force-updated after validation.

**Base:** `stack/external-03-persistence`  
**Review order:** merge only after the preceding stack layer is accepted. This PR is not intended to merge independently out of order.

## Stack navigation

1. [#1158](https://github.com/PrimeIntellect-ai/prime-agent/pull/1158) — ci: harden verification and release compatibility
2. [#1159](https://github.com/PrimeIntellect-ai/prime-agent/pull/1159) — fix(security): harden session and autonomous execution boundaries
3. [#1160](https://github.com/PrimeIntellect-ai/prime-agent/pull/1160) — fix(coding-agent): make persisted state crash-safe
4. [#1161](https://github.com/PrimeIntellect-ai/prime-agent/pull/1161) — fix(daemon): fence worker and supervisor lifecycle state
5. [#1162](https://github.com/PrimeIntellect-ai/prime-agent/pull/1162) — fix(coding-agent): repair queued and archived session lifecycle
6. [#1163](https://github.com/PrimeIntellect-ai/prime-agent/pull/1163) — fix(coding-agent): complete Windows kernel and daemon startup
7. [#1164](https://github.com/PrimeIntellect-ai/prime-agent/pull/1164) — fix(providers): harden MCP OAuth and Codex transports
8. [#1165](https://github.com/PrimeIntellect-ai/prime-agent/pull/1165) — fix(runtime): bound transcript and autonomous recovery

## Summary

- Prevent PID reuse or unverifiable process identity from signaling unrelated workers.
- Generation-fence and renew supervisor launch locks across reclaim, release, and spawn races.
- Refresh registry ownership safely and preserve healthy cached heartbeat snapshots.

## Validation

- `npm run check`; final focused daemon/ownership/heartbeat/session-lease suite 34/34.

## Provenance

- Authored independently from `upstream/main` using issue reports and PR descriptions/comments only.
- No external contributor branch, diff, commit, implementation code, or test code was fetched, inspected, copied, or reused.
- The implementation and regression tests in this stack are maintainer-owned.

### Fixed on merge

- Fixes #1045 — One failed worker descriptor poisons global heartbeats_list; stopWorker EPERM/pid-reuse false-alive marks exited workers "failed"
- Fixes #1131 — Bug: Daemon startup permanently fails with Lock file is already being held after a crashed startup attempt leaves an orphaned lock directory
- Fixes #1148 — Daemon supervisor permanently fails with supervisor_generation_stale after the OS prunes its registry entry from the temp directory

### Independently superseded pull requests

- #1046 (open) — fix(coding-agent): verify worker process identity in stopWorker; skip failed workers in global heartbeats_list — **will be closed with a pointer to this independently authored PR**
- #1006 (open) — fix(coding-agent): acquire the bootstrap lock with an atomic rename — **will be closed with a pointer to this independently authored PR**
- #1149 (open) — fix(coding-agent): recover supervisor ownership when its registry entry is pruned — **will be closed with a pointer to this independently authored PR**

## Reviewer notes

- Please review this layer against its immediate stack base, not against `main`, to avoid cumulative duplicate diffs.
- No merge is requested; the complete stack is being left for human review.


packages/coding-agent/src/modes/daemon/daemon-mode.ts
```diff
@@ -757,52 +982,117 @@ export class AgentDaemon {
...
 					}
-					const staleDirectory = `${lockDirectory}.stale-${process.pid}-${token}`;
-					try {
-						renameSync(lockDirectory, staleDirectory);
-						rmSync(staleDirectory, { recursive: true, force: true });
-					} catch (reclaimError) {
-						if ((reclaimError as NodeJS.ErrnoException).code !== "ENOENT") {
-							throw reclaimError;
+					return current;
+				} catch (error) {
```
varindersingh83, line 1065: This catch path gives a registry-guard error the same `false` result as proven generation loss. A transient contention or I/O error can therefore abort launch, after which `finally` may remove the still-valid generation.

Could this return distinct `current`, `lost`, and `unknown/error` outcomes, with bounded retry for the last case, and only release after a guarded read proves ownership loss?

cc @sethkarten @snimu

packages/coding-agent/src/modes/daemon/daemon-supervisor.ts
```diff
@@ -518,6 +518,53 @@ function isProcessAlive(pid: number): boolean {
...
+			return lastStatus;
+		}
+		const now = Date.now();
+		if (forceIdentityCheck || lastStatus === undefined || now >= nextIdentityCheckAt) {
+			nextIdentityCheckAt = now + identityPollIntervalMs;
+			const observedProcessStartId = getProcessStartId(descriptor.pid);
+			lastStatus =
+				observedProcessStartId === undefined
+					? "unknown"
+					: observedProcessStartId === descriptor.processStartId
```
varindersingh83, line 553: The portable `ps -o lstart` identity has whole-second precision, but equality here is treated as sufficient authorization for later SIGTERM/SIGKILL paths. If a PID is reused within that second, an unrelated process can compare as `match`.

Could `ps:` identities be treated as coarse and fail closed for destructive signalling, with a regression covering same-second PID reuse?

sethkarten:
Closing this abandoned implementation snapshot. Remaining work is tracked in #1381; this PR remains available as design and validation history.

PR #1164, opened 2026-08-10
[stack 7/8] fix(providers): harden MCP OAuth and Codex transports

sethkarten:
# Stack 7/8 — fix(providers): harden MCP OAuth and Codex transports

> **Active review snapshot — do not merge yet.** The complete stack is open for architecture/design review, while final cumulative audit, CI, Cursor Bug Bot, and Macroscope findings are being remediated. Branches will be force-updated after validation.

**Base:** `stack/external-06-windows`  
**Review order:** merge only after the preceding stack layer is accepted. This PR is not intended to merge independently out of order.

## Stack navigation

1. [#1158](https://github.com/PrimeIntellect-ai/prime-agent/pull/1158) — ci: harden verification and release compatibility
2. [#1159](https://github.com/PrimeIntellect-ai/prime-agent/pull/1159) — fix(security): harden session and autonomous execution boundaries
3. [#1160](https://github.com/PrimeIntellect-ai/prime-agent/pull/1160) — fix(coding-agent): make persisted state crash-safe
4. [#1161](https://github.com/PrimeIntellect-ai/prime-agent/pull/1161) — fix(daemon): fence worker and supervisor lifecycle state
5. [#1162](https://github.com/PrimeIntellect-ai/prime-agent/pull/1162) — fix(coding-agent): repair queued and archived session lifecycle
6. [#1163](https://github.com/PrimeIntellect-ai/prime-agent/pull/1163) — fix(coding-agent): complete Windows kernel and daemon startup
7. [#1164](https://github.com/PrimeIntellect-ai/prime-agent/pull/1164) — fix(providers): harden MCP OAuth and Codex transports
8. [#1165](https://github.com/PrimeIntellect-ai/prime-agent/pull/1165) — fix(runtime): bound transcript and autonomous recovery

## Summary

- Preserve MCP tool schemas and use the correct Codex discovery client version.
- Harden Codex SSE/WebSocket framing, cancellation, cleanup, and fallback state.
- Implement SSRF-safe RFC 9728/8414/8707 OAuth discovery, native MCP challenges, refresh binding, and RFC 6052 NAT64 policy.

## Validation

- `npm run check`; targeted AI provider suites 99/99; MCP manager 14/14.
- Residual/non-blocking: WWW-Authenticate-only OAuth challenge discovery remains follow-up; provider stack must include independent #1153 empty-catalog commit

## Provenance

- Authored independently from `upstream/main` using issue reports and PR descriptions/comments only.
[...]

PR #1163, opened 2026-08-10
[stack 6/8] fix(coding-agent): complete Windows kernel and daemon startup

sethkarten:
# Stack 6/8 — fix(coding-agent): complete Windows kernel and daemon startup

> **Active review snapshot — do not merge yet.** The complete stack is open for architecture/design review, while final cumulative audit, CI, Cursor Bug Bot, and Macroscope findings are being remediated. Branches will be force-updated after validation.

**Base:** `stack/external-05-lifecycle`  
**Review order:** merge only after the preceding stack layer is accepted. This PR is not intended to merge independently out of order.

## Stack navigation

1. [#1158](https://github.com/PrimeIntellect-ai/prime-agent/pull/1158) — ci: harden verification and release compatibility
2. [#1159](https://github.com/PrimeIntellect-ai/prime-agent/pull/1159) — fix(security): harden session and autonomous execution boundaries
3. [#1160](https://github.com/PrimeIntellect-ai/prime-agent/pull/1160) — fix(coding-agent): make persisted state crash-safe
4. [#1161](https://github.com/PrimeIntellect-ai/prime-agent/pull/1161) — fix(daemon): fence worker and supervisor lifecycle state
5. [#1162](https://github.com/PrimeIntellect-ai/prime-agent/pull/1162) — fix(coding-agent): repair queued and archived session lifecycle
6. [#1163](https://github.com/PrimeIntellect-ai/prime-agent/pull/1163) — fix(coding-agent): complete Windows kernel and daemon startup
7. [#1164](https://github.com/PrimeIntellect-ai/prime-agent/pull/1164) — fix(providers): harden MCP OAuth and Codex transports
8. [#1165](https://github.com/PrimeIntellect-ai/prime-agent/pull/1165) — fix(runtime): bound transcript and autonomous recovery

## Summary

- Use the Windows managed-venv interpreter layout and official PowerShell uv bootstrap.
- Avoid catalog/worker startup livelocks and perform fresh process-identity validation.
- Require a native Windows named-pipe smoke covering supervisor, worker authentication, RPC, and catalog readiness.

## Validation

- `npm run check`; focused daemon suites 52/52; required native `windows-latest` supervisor/worker/catalog smoke must pass on the PR.

## Provenance

- Authored independently from `upstream/main` using issue reports and PR descriptions/comments only.
- No external contributor branch, diff, commit, implementation code, or test code was fetched, inspected, copied, or reused.
- The implementation and regression tests in this stack are maintainer-owned.

### Fixed on merge

[...]

PR #1162, opened 2026-08-10
[stack 5/8] fix(coding-agent): repair queued and archived session lifecycle

sethkarten:
# Stack 5/8 — fix(coding-agent): repair queued and archived session lifecycle

> **Active review snapshot — do not merge yet.** The complete stack is open for architecture/design review, while final cumulative audit, CI, Cursor Bug Bot, and Macroscope findings are being remediated. Branches will be force-updated after validation.

**Base:** `stack/external-04-daemon-foundations`  
**Review order:** merge only after the preceding stack layer is accepted. This PR is not intended to merge independently out of order.

## Stack navigation

1. [#1158](https://github.com/PrimeIntellect-ai/prime-agent/pull/1158) — ci: harden verification and release compatibility
2. [#1159](https://github.com/PrimeIntellect-ai/prime-agent/pull/1159) — fix(security): harden session and autonomous execution boundaries
3. [#1160](https://github.com/PrimeIntellect-ai/prime-agent/pull/1160) — fix(coding-agent): make persisted state crash-safe
4. [#1161](https://github.com/PrimeIntellect-ai/prime-agent/pull/1161) — fix(daemon): fence worker and supervisor lifecycle state
5. [#1162](https://github.com/PrimeIntellect-ai/prime-agent/pull/1162) — fix(coding-agent): repair queued and archived session lifecycle
6. [#1163](https://github.com/PrimeIntellect-ai/prime-agent/pull/1163) — fix(coding-agent): complete Windows kernel and daemon startup
7. [#1164](https://github.com/PrimeIntellect-ai/prime-agent/pull/1164) — fix(providers): harden MCP OAuth and Codex transports
8. [#1165](https://github.com/PrimeIntellect-ai/prime-agent/pull/1165) — fix(runtime): bound transcript and autonomous recovery

## Summary

- Wake idle programmatic prompts without puncturing update-restart suspension.
- Batch child usage attribution per turn while preserving in-memory accounting.
- Terminate POSIX bash process groups with cooperative grace.
- Revive archived sessions once and retry late concurrent prompts against the revived ID.

## Validation

- `npm run check`; final clean-environment lifecycle suites 189/189.
- Residual/non-blocking: #1072 only archived-session recovery; closing/mutation-drain leak remains; Windows process tree containment remains

## Provenance

- Authored independently from `upstream/main` using issue reports and PR descriptions/comments only.
- No external contributor branch, diff, commit, implementation code, or test code was fetched, inspected, copied, or reused.
[...]

PR #1113, opened 2026-08-10
fix(coding-agent): pause a stalled goal instead of looping continuations

romankhadka:
## Summary

- Fixes #986: an active goal with no token budget injected identical `goal_context` continuations forever when the model was blocked on user input (a required approval, credential, or unanswered question).
- `_getGoalContinuationMessages` now checks the run for stalled continuation windows before re-arming. When the trailing three goal-context windows all ended without a tool call or a user message, the goal pauses (`status: "paused"` with a new optional `pausedBy: "user" | "host"` field and an explanatory `lastReason`) instead of injecting another identical context.
- A goal paused this way (`pausedBy: "host"`) resumes automatically on the next genuine user message, whether it arrives via `prompt()`, `steer()`, or `followUp()`. Host-generated prompts and custom-message deliveries (heartbeats, agent messages) do not resume it. An explicit `/goal pause` sets `pausedBy: "user"` and still requires `/goal resume`.
- The continuation prompt now tells the model to state what it is waiting for and end the turn without tool calls when blocked, so the pause and the model's behavior agree.

## Design notes

- The threshold is three consecutive stalled windows. A continuation after a toolless turn is a legitimate nudge (the existing goal suite depends on two consecutive nudges before completion), so a one-strike rule would break intended behavior. Three preserves the nudge and bounds the loop.
- Detection is a pure function over the run's messages (`goalContinuationIsStalled` in `goals.ts`), so no extra mutable state is needed: any tool call, user message, or new run resets the window naturally.
- The detector treats any tool call as progress: a model that polls a tool (for example `goal.get`) on every continuation is not paused by this change. That variant does observable work each turn and is bounded by a token budget when one is set; the updated continuation prompt instructs blocked models to end the turn without tool calls, which makes sanctioned waiting detectable. Stream-level degenerate repetition is a separate concern (#1029).
- `pausedBy` is optional and normalized to exist only while `status` is `"paused"`; older builds reading a persisted goal state see an ordinary paused goal. A discriminator was chosen over a boolean so future pause causes extend the same field.

## Validation

- `cd packages/coding-agent && npx tsx ../../node_modules/vitest/dist/cli.js --run test/suite/regressions/986-goal-continuation-loop.test.ts test/suite/agent-session-goal.test.ts test/suite/agent-session-queue.test.ts`: 147 passed
[...]

PR #1165, opened 2026-08-10
[stack 8/8] fix(runtime): bound transcript and autonomous recovery

### LINKED ISSUE #900: Compaction can permanently self-amplify from durable retry debris and unbounded summary input

### LINKED ISSUE #986: Goal Mode Keeps Looping After Completion
# Bug Report: Prime Agent Goal Mode Keeps Looping After Completion

## Summary

Prime Agent’s **Goal Mode** enters an infinite continuation loop when the goal has no remaining actionable work and is waiting for explicit user input or external resources.

Instead of pausing and waiting for the user, Goal Mode repeatedly injects a new `goal_context` continuation and generates essentially the same response again.

## Observed Behavior

The user asked, **“Which offer?”**, and the agent explained the offer. After that, the automatic `goal_context` was triggered without the user replying or confirming whether the offer should be executed.

The agent correctly determined that it should not proceed without confirmation:

> The user has not confirmed the offer. I should keep the response short and stable while waiting for confirmation regarding the CRM/Subscription story or the resources required by BILL-014 and BILL-015. I will not execute anything without confirmation because the offer was explicitly conditional on the user’s approval.

The current state was:

- `BILL-001` through `BILL-013`: completed
- `BILL-014` and `BILL-015`: in backlog
- `BILL-014` requires an Asaas Sandbox key to run the evidence harness
- `BILL-015` requires explicit human approval for the go-live rollout
- The optional CRM/Subscription Storybook story for `/home/assinatura` was awaiting a yes/no response from the user

The agent then returned an appropriate waiting-state response:

> State unchanged: BILL-001 through BILL-013 are done. BILL-014 and BILL-015 remain in the backlog, waiting for:
>
> 1. An Asaas Sandbox key to run the evidence harness.
> 2. Go-live approval for the rollout sequence.
>
> The goal has not been marked as complete because both stories require real evidence and human approval.
>
> The CRM/Subscription Storybook offer for the `/home/assinatura` page remains open and is awaiting a yes/no response. Without confirmation, I will not execute unrelated or marginal work.

However, Goal Mode immediately triggered another continuation:

```text
Goal continuation · implement all BILL stories with TaskFleet
```

The agent produced the same waiting-state response again. This cycle continued repeatedly, even though:

- No new user message had been received.
- No new resources had become available.
- No executable work remained.
- The agent had already identified that human confirmation was required.
- Repeating the response could not make progress toward the goal.

## Steps to Reproduce

1. Start Prime Agent in Goal Mode.
2. Give it a multi-task goal.
3. Allow the agent to complete all tasks that can be completed autonomously.
4. Leave one or more tasks blocked by:
   - Missing credentials or external resources.
   - Required human approval.
   - An unanswered yes/no question.
5. Let the agent report that it is waiting for the user.
6. Do not send a new user message.
7. Observe that Goal Mode automatically triggers another goal continuation.
8. Observe that the agent repeats the same blocked-state response indefinitely.

## Expected Behavior

When the agent determines that:

- No actionable tasks remain,
- Progress depends entirely on human input or unavailable external resources, and
- The waiting condition has already been communicated,

Goal Mode should transition into a stable paused state.

It should not trigger another continuation until one of the following occurs:

- The user sends a new message.
- A required resource becomes available.
- An external event changes the task state.
- A timeout or retry policy explicitly allows another attempt.
- The user manually resumes the goal.

A blocked goal should be distinguishable from both a completed goal and an actively executable goal.

## Actual Behavior

Goal Mode repeatedly schedules itself despite having no new information or executable work.

Each continuation produces nearly identical reasoning and output:

```text
The user did not respond to the story offer. Only the automatic
goal_context was received. Keep the response stable and short.

State unchanged: BILL-001 through BILL-013 are done. BILL-014 and
BILL-015 remain in the backlog, awaiting an Asaas Sandbox key and
go-live approval.

The CRM/Subscription Storybook offer remains open, awaiting a yes/no
response. Without confirmation, do not execute marginal work.
```

The same continuation and response are then generated again.

## Impact

This behavior:

- Consumes tokens and compute without making progress.
- Produces repetitive output in the conversation.
- Makes Goal Mode appear unable to recognize a blocked state.
- Can continue indefinitely until manually interrupted.
- Creates unnecessary cost for long-running autonomous sessions.
- Risks hitting context limits while repeatedly appending equivalent messages.
- Makes it difficult to distinguish actual progress from scheduler-generated activity.

### LINKED ISSUE #1029: No safeguard against degenerate model repetition: "The the the..." thinking stream until manual abort

sethkarten:
# Stack 8/8 — fix(runtime): bound transcript and autonomous recovery

> **Active review snapshot — do not merge yet.** The complete stack is open for architecture/design review, while final cumulative audit, CI, Cursor Bug Bot, and Macroscope findings are being remediated. Branches will be force-updated after validation.

**Base:** `stack/external-07-providers`  
**Review order:** merge only after the preceding stack layer is accepted. This PR is not intended to merge independently out of order.

## Stack navigation

1. [#1158](https://github.com/PrimeIntellect-ai/prime-agent/pull/1158) — ci: harden verification and release compatibility
2. [#1159](https://github.com/PrimeIntellect-ai/prime-agent/pull/1159) — fix(security): harden session and autonomous execution boundaries
3. [#1160](https://github.com/PrimeIntellect-ai/prime-agent/pull/1160) — fix(coding-agent): make persisted state crash-safe
4. [#1161](https://github.com/PrimeIntellect-ai/prime-agent/pull/1161) — fix(daemon): fence worker and supervisor lifecycle state
5. [#1162](https://github.com/PrimeIntellect-ai/prime-agent/pull/1162) — fix(coding-agent): repair queued and archived session lifecycle
6. [#1163](https://github.com/PrimeIntellect-ai/prime-agent/pull/1163) — fix(coding-agent): complete Windows kernel and daemon startup
7. [#1164](https://github.com/PrimeIntellect-ai/prime-agent/pull/1164) — fix(providers): harden MCP OAuth and Codex transports
8. [#1165](https://github.com/PrimeIntellect-ai/prime-agent/pull/1165) — fix(runtime): bound transcript and autonomous recovery

## Summary

- Repair interrupted tool-result ordering without orphaning or duplicating results.
- Bound degenerate reasoning generation with diagnostics and an opt-out.
- Persist goal pause/resume semantics for externally blocked work.
- Compact complete input with conservative budgets, bounded progress, and fail-closed summaries.

## Validation

- `npm run check`; focused compaction/autonomy suites 76 passed with 2 credential-gated skips; final adversarial review cleared.
- Residual/non-blocking: guard limited to OpenAI reasoning; oversized compaction message truncates; no extra cooldown

## Provenance

- Authored independently from `upstream/main` using issue reports and PR descriptions/comments only.
- No external contributor branch, diff, commit, implementation code, or test code was fetched, inspected, copied, or reused.
- The implementation and regression tests in this stack are maintainer-owned.

### Fixed on merge

- Fixes #1029 — No safeguard against degenerate model repetition: "The the the..." thinking stream until manual abort
- Fixes #986 — Goal Mode Keeps Looping After Completion
- Fixes #900 — Compaction can permanently self-amplify from durable retry debris and unbounded summary input

### Independently superseded pull requests

- #1102 (open) — fix(ai): repair interrupted tool result ordering — **will be closed with a pointer to this independently authored PR**
- #1099 (open) — fix(ai): abort degenerate generation instead of streaming it to the cap — **will be closed with a pointer to this independently authored PR**
- #1113 (open) — fix(coding-agent): pause a stalled goal instead of looping continuations — **will be closed with a pointer to this independently authored PR**
- #1055 (open) — fix: prevent self-amplifying compaction failure loop (#900) — **will be closed with a pointer to this independently authored PR**

### Partial/distinct overlap — remains open

- #984 — transformMessages orphans tool results when the parent assistant turn was aborted or errored
- #991 — fix(ai): drop tool results whose parent tool call was aborted

## Reviewer notes

- Please review this layer against its immediate stack base, not against `main`, to avoid cumulative duplicate diffs.
- No merge is requested; the complete stack is being left for human review.

### DIFF
--- a/packages/ai/src/providers/openai-completions.ts
+++ b/packages/ai/src/providers/openai-completions.ts
@@ -34,7 +34,9 @@ import type {
 import { AssistantMessageEventStream } from "../utils/event-stream.js";
 import { headersToRecord } from "../utils/headers.js";
 import { parseStreamingJson } from "../utils/json-parse.js";
+import { REPETITION_GUARD_DISABLED_ENV, RepetitionGuard } from "../utils/repetition-guard.js";
 import { sanitizeSurrogates } from "../utils/sanitize-unicode.js";
+import { formatStreamFailureMessage, recordStreamFailure, StreamFailureError } from "../utils/stream-failure.js";
 import { isCloudflareProvider, resolveCloudflareBaseUrl } from "./cloudflare.js";
 import { buildCopilotDynamicHeaders, hasCopilotVisionInput } from "./github-copilot-headers.js";
 import { buildBaseOptions } from "./simple-options.js";
@@ -173,6 +175,10 @@ export const streamOpenAICompletions: StreamFunction<"openai-completions", OpenA
 			let thinkingBlock: ThinkingContent | null = null;
 			const toolCallBlocksByIndex = new Map<number, StreamingToolCallBlock>();
 			const toolCallBlocksById = new Map<string, StreamingToolCallBlock>();
+			const repetitionGuard =
+				typeof process !== "undefined" && process.env[REPETITION_GUARD_DISABLED_ENV] === "1"
+					? undefined
+					: new RepetitionGuard();
 			const blocks = output.content as StreamingBlock[];
 			const getContentIndex = (block: StreamingBlock) => blocks.indexOf(block);
 			const finishBlock = (block: StreamingBlock) => {
@@ -330,6 +336,14 @@ export const streamOpenAICompletions: StreamFunction<"openai-completions", OpenA
 					if (foundReasoningField) {
 						const delta = deltaFields[foundReasoningField];
 						if (typeof delta === "string" && delta.length > 0) {
+							const repetition = repetitionGuard?.push(delta);
+							if (repetition) {
+								throw new StreamFailureError("Model reasoning became repetitively degenerate", {
+									kind: "degenerate_output",
+									providerErrorType: `repetition:${repetition.reason}`,
+									raw: JSON.stringify(repetition),
+								});
+							}
 							const block = ensureThinkingBlock(foundReasoningField);
 							block.thinking += delta;
 							stream.push({
@@ -407,10 +421,11 @@ export const streamOpenAICompletions: StreamFunction<"openai-completions", OpenA
 				delete (block as { streamIndex?: number }).streamIndex;
 			}
 			output.stopReason = options?.signal?.aborted ? "aborted" : "error";
-			output.errorMessage = error instanceof Error ? error.message : JSON.stringify(error);
+			output.errorMessage = formatStreamFailureMessage(error);
 			// Some providers via OpenRouter give additional information in this field.
 			const rawMetadata = (error as any)?.error?.metadata?.raw;
 			if (rawMetadata) output.errorMessage += `\n${rawMetadata}`;
+			recordStreamFailure(model, output, error);
 			stream.push({ type: "error", reason: output.stopReason, error: output });
 			stream.end();
 		}
--- a/packages/ai/CHANGELOG.md
+++ b/packages/ai/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [Unreleased]
 
+- Fixed runaway repetitive reasoning streams consuming their full output budget ([#1099](https://github.com/PrimeIntellect-ai/prime-agent/pull/1099)).
 - Fixed interrupted tool results being replayed after interposed user messages ([#1102](https://github.com/PrimeIntellect-ai/prime-agent/pull/1102)).
 
 ## [0.7.1] - 2026-08-07
--- a/packages/ai/CHANGELOG.md
+++ b/packages/ai/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## [Unreleased]
 
+- Fixed interrupted tool results being replayed after interposed user messages ([#1102](https://github.com/PrimeIntellect-ai/prime-agent/pull/1102)).
+
 ## [0.7.1] - 2026-08-07
 
 ## [0.7.0] - 2026-08-05
--- a/packages/ai/src/providers/transform-messages.ts
+++ b/packages/ai/src/providers/transform-messages.ts
@@ -152,69 +152,66 @@ export function transformMessages<TApi extends Api>(
 		return msg;
 	});
 
-	// Second pass: insert synthetic empty tool results for orphaned tool calls
-	// This preserves thinking signatures and satisfies API requirements
+	// Second pass: keep every tool result adjacent to its assistant tool call.
+	// Interrupted turns can persist a user/custom message before the real result arrives.
+	// Look ahead only until the next assistant turn, hoist the first matching real
+	// result for each call, and synthesize results only for calls still unresolved.
 	const result: Message[] = [];
-	let pendingToolCalls: ToolCall[] = [];
-	let existingToolResultIds = new Set<string>();
-	const insertSyntheticToolResults = () => {
-		if (pendingToolCalls.length > 0) {
-			for (const tc of pendingToolCalls) {
-				if (!existingToolResultIds.has(tc.id)) {
-					result.push({
-						role: "toolResult",
-						toolCallId: tc.id,
-						toolName: tc.name,
-						content: [{ type: "text", text: "No result provided" }],
-						isError: true,
-						timestamp: Date.now(),
-					} as ToolResultMessage);
-				}
-			}
-			pendingToolCalls = [];
-			existingToolResultIds = new Set();
-		}
-	};
-
 	for (let i = 0; i < transformed.length; i++) {
 		const msg = transformed[i];
 
-		if (msg.role === "assistant") {
-			// If we have pending orphaned tool calls from a previous assistant, insert synthetic results now
-			insertSyntheticToolResults();
+		if (msg.role === "toolResult") {
+			// Results are emitted with their matching assistant below. Anything left is
+			// an orphan or duplicate and is invalid model context.
+			continue;
+		}
 
-			// Skip errored/aborted assistant messages entirely.
-			// These are incomplete turns that shouldn't be replayed:
-			// - May have partial content (reasoning without message, incomplete tool calls)
-			// - Replaying them can cause API errors (e.g., OpenAI "reasoning without following item")
-			// - The model should retry from the last valid state
-			const assistantMsg = msg as AssistantMessage;
-			if (assistantMsg.stopReason === "error" || assistantMsg.stopReason === "aborted") {
-				continue;
+		if (msg.role !== "assistant") {
+			result.push(msg);
+			continue;
+		}
+
+		// Skip errored/aborted assistant messages entirely.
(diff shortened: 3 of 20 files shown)


packages/ai/src/providers/openai-completions.ts
```diff
@@ -330,6 +336,14 @@ export const streamOpenAICompletions: StreamFunction<"openai-completions", OpenA
 					if (foundReasoningField) {
 						const delta = deltaFields[foundReasoningField];
 						if (typeof delta === "string" && delta.length > 0) {
+							const repetition = repetitionGuard?.push(delta);
```
gogoyqj, line 339: Nice catch on the degenerate repetition issue. The repetitionGuard is a solid safety net — detecting and throwing StreamFailureError on degenerate output is the right defensive measure.

 One suggestion: consider also adding anti-repetition guidance to the system prompt as a complementary layer. The guard catches repetition after it starts (wasting tokens in the process), while prompt-level 
 instructions can prevent it from starting.

 Claude Code's system prompt includes explicit anti-repetition directives like:
 Do NOT read this resource again unless you think it may have changed.
 Do not re-select memories you already returned for an earlier query.
 When you already know which part of the file you need, only read that part.

 Something similar in buildRlmPrompt() (e.g. in rlm.ts) would reduce how often the guard triggers:
 Do not re-run the same tool call with identical arguments unless the underlying data may have changed.
 If you find yourself repeating the same action with the same result, stop and reconsider your approach.
 Do not re-read files or re-execute commands whose results you already have in context.

 This way the prompt prevents most repetition, and the repetitionGuard catches the rest. Neither alone is sufficient — prompts can't guarantee compliance, and code-level detection without prompt guidance means 
 the guard fires frequently and degrades UX.

sethkarten:
Closing this abandoned implementation snapshot. Remaining work is tracked in #1384; this PR remains available as design and validation history.

PR #1187, opened 2026-08-11
feat(coding-agent): add async bash() to IPython kernel

samsja:
## Summary

Adds an async `bash()` function to the IPython kernel bootstrap code so `await bash("...")` runs shell commands via `asyncio.create_subprocess_exec` without blocking the kernel event loop.

## Motivation

`%%bash` cells block the IPython kernel until the command finishes — the kernel can't respond to interrupts or other messages while a shell command runs. This is the same anti-pattern as `time.sleep()` polling. `await bash("...")` uses `asyncio.create_subprocess_exec` so the kernel event loop stays responsive.

## Changes

- `ipython.ts`: Added `_PrimeAgentBashResult` class and `async bash()` function to `RLM_BOOTSTRAP_BASE_CODE`. Supports `timeout` (raises `TimeoutError`) and `cwd` parameters. Returns stdout, stderr, and returncode.
- `rlm.ts`: Updated the IPython control prompt to prefer `await bash("...")` over `%%bash` cells.
- `system-prompt.test.ts`: Updated exact-match block and added `toContain` assertion for the new prompt text.

## Usage

```python
result = await bash("echo hello")           # BashResult(stdout="hello\n", ...)
result = await bash("sleep 5", timeout=2)   # raises TimeoutError
print(result.stdout, result.stderr, result.returncode)
```

## Testing

- `npx vitest run test/system-prompt.test.ts` — 25/25 pass
- `npx vitest run test/ipython-bootstrap.test.ts` — 6/6 pass (including real kernel tests)
- Manually verified in a real IPython kernel: stdout/stderr capture, exit codes, and timeout all work correctly
- `npm run check` — clean

---

> [!NOTE]
> **Medium Risk**
> Changes how shell commands run in the kernel (new process groups and aggressive descendant termination), which can differ from `%%bash` behavior for background jobs; scope is confined to the agent IPython runtime on POSIX.
> 
> **Overview**
> Adds an async **`bash()`** helper to the IPython kernel bootstrap so agents can run **`await bash("...")`** instead of blocking **`%%bash`** cells, keeping the kernel event loop responsive to interrupts and cancellation.
> 
> The implementation spawns commands in a new POSIX session, streams stdout/stderr into **tail-bounded buffers** (default 32 KiB per stream, configurable via `max_output_bytes`), returns a structured result with truncation metadata, and **tears down the process group** on timeout, cancellation, or after the shell exits (including background descendants). Result objects render captured output in IPython when used as the final cell expression.
> 
> **RLM / system prompts** now tell the model to prefer `await bash("...")` over `%%bash`, while still documenting `%%bash` rules. **Changelog** and **bootstrap / real-kernel tests** cover output bounds, UTF-8 handling, descendant cleanup, and SIGTERM/SIGKILL escalation.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 02543c75e57fc1cef3fe3595267a1014b5bc30d7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Add async `bash()` coroutine to the IPython kernel for non-blocking shell execution
> - Injects an async `bash()` function into the IPython kernel bootstrap so shell commands run without blocking the event loop, replacing the recommended use of `%%bash` magic cells.
> - Captures stdout/stderr into bounded buffers with a configurable byte cap, reporting truncation flags and total byte counts via a structured `_PrimeAgentBashResult` return value.
> - Handles timeouts and cancellation with robust process-group termination: POSIX uses `killpg` in a new session; Windows uses `taskkill /T /F`.
> - Updates the RLM system prompt in [rlm.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/1187/files#diff-64c77a41480409e73a6d87d381a618f83c44304c3677d755b3ef66d7a5ddce29) to prefer `await bash("...")` over `%%bash`, with guidance on responsiveness, cleanup, and output capture.
> - Risk: `bash()` spawns a new process group/session on POSIX, which changes the process tree structure relative to `%%bash` or `!cmd` shell escapes.
>
> 
> #### Changes since #1187 opened
>
> - Removed Windows support from `bash()` function in IPython kernel [91eae25]
> - Updated IPython bootstrap test expectations for Windows bash() unavailability [91eae25]
> - Modified `_PrimeAgentBashResult.__repr__` method to return the captured command output (combined stdout/stderr and exit code) instead of a metadata summary [785b9b7]
> - Added test case validating that `await bash()` renders command output when used as the final expression in an IPython cell [785b9b7]
> - Reworked bash command process completion and cleanup logic in `_prime_agent_bash_collect` function to terminate background descendants and handle output draining [6b696f8]
> - Reduced default bash command output capture limit and added explicit truncation notices to output representation [6b696f8]
> - Added test verification for truncation notices in bash command output [6b696f8]
> - Added test coverage for background process cleanup after shell exit [6b696f8]
> - Added explicit boolean type rejection to `max_output_bytes` parameter validation in the `bash()` function within the IPython kernel bootstrap code [9a05080]
> - Changed decoding of bounded bash output from UTF-8 to Latin-1 in IPython kernel [66902ff]
> - Added test case for malformed byte output handling in bash execution [66902ff]
> - Updated IPython shell execution guidance wording in prompt constant [66902ff]
> - Replaced timeout-based SIGKILL escalation with process-group-existence-based escalation in `_prime_agent_bash_collect` and `_prime_agent_bash_stop` functions [af69851]
> - Added test validating SIGKILL escalation for descendants that ignore SIGTERM and close pipes [af69851]
> - Changed `_PrimeAgentBoundedBytes.text` to decode captured bytes as UTF-8 with replacement for malformed sequences, then re-encode and head-trim the UTF-8 payload when it exceeds the byte limit, setting a new `render_truncated` instance attribute to track head-trimming, and updated `_PrimeAgentBoundedBytes.__init__` to initialize `render_truncated` to False [a15a9e0]
> - Updated `_PrimeAgentBashResult.__str__` to compute and display truncation byte counts using UTF-8 encoding instead of latin-1 for both stdout and stderr [a15a9e0]
> - Modified `bash` async function to combine raw capture truncation flags with the new `render_truncated` flag from `_PrimeAgentBoundedBytes` using OR operation for both stdout and stderr [a15a9e0]
> - Added test case verifying UTF-8 command output preservation through `bash` and modified existing truncation test to use UTF-8 byte lengths for assertions [a15a9e0]
> - Added platform-specific shell command execution guidance to `IPYTHON_CONTROL_PROMPT` constant [dd88b46]
> - Changed truncation message format in `_PrimeAgentBashResult.__str__` method for stdout and stderr [bdbd8ec]
> - Removed Windows platform restriction from `bash()` async function in IPython kernel bootstrap [02543c7]
> - Updated `IPYTHON_CONTROL_PROMPT` constant to provide platform-agnostic shell command guidance [02543c7]
> - Enabled Windows execution for IPython bash() integration tests [02543c7]
> - Updated `buildRlmPrompt` test assertions to expect platform-agnostic guidance [02543c7]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized ff15d7b.</sup>
> 
>


packages/coding-agent/src/core/prompts/rlm.ts
```diff
@@ -16,7 +16,7 @@ const IPYTHON_CONTROL_PROMPT = [
...
 	"",
-	"When running shell commands from IPython, use `%%bash` cells. If you use `%%bash`, it must be the first line of the code cell: no comments, spaces, blank lines, imports, or Python statements before it. Avoid `!cmd` shell escapes for project commands so shell behavior is explicit and multi-line commands share one shell context.",
+	'When running shell commands from IPython, prefer `await bash("...")` over `%%bash` cells: it runs the command via `asyncio.create_subprocess_exec` so the kernel event loop stays responsive to interrupts while the process runs. `%%bash` cells block the kernel until the command finishes. If you do use `%%bash`, it must be the first line of the code cell: no comments, spaces, blank lines, imports, or Python statements before it. Avoid `!cmd` shell escapes for project commands so shell behavior is explicit and multi-line commands share one shell context.',
```
sethkarten: should we mention python subprocess run here?

sethkarten: wait nvm im illiterate

packages/coding-agent/src/core/tools/ipython.ts
```diff
@@ -34,6 +34,77 @@ try:
...
+        cwd: Working directory (defaults to kernel cwd).
+
+    Returns:
+        _PrimeAgentBashResult with stdout, stderr, and returncode.
+    """
+    shell = _prime_agent_shutil.which("bash") or "/bin/bash"
+    env = dict(_prime_agent_bash_os.environ)
+    work_dir = cwd or _prime_agent_bash_os.getcwd()
+
+    proc = await _prime_agent_asyncio.create_subprocess_exec(
```
sethkarten, line 213: potential high ipython kernel memory usage issue here. critical to address

sethkarten (commented):
subprocess lifecycle clean-up and potential high memory usage needs addressing

packages/coding-agent/src/core/tools/ipython.ts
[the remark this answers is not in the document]
sethkarten: Fixed in ff15d7b06. `bash()` now drains stdout and stderr concurrently into fixed-size tail buffers (1 MiB per stream by default), exposes truncation and total-byte metadata, and includes a real-kernel regression that writes 200 KB to both streams with a 4 KB cap.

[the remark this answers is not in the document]
sethkarten: Fixed in ff15d7b06. POSIX commands now run in their own session and cleanup signals the entire process group on timeout before reaping all pipe tasks. Windows uses `taskkill /T /F`. The real-kernel timeout regression proves a background child cannot perform a delayed side effect.

packages/coding-agent/src/core/tools/ipython.ts
```diff
@@ -34,6 +34,77 @@ try:
...
+    )
+
+    try:
+        stdout_bytes, stderr_bytes = await _prime_agent_asyncio.wait_for(
+            proc.communicate(), timeout=timeout
+        )
+    except _prime_agent_asyncio.TimeoutError:
+        proc.kill()
+        await proc.wait()
+        raise TimeoutError(f"bash command timed out after {timeout}s: {command}")
```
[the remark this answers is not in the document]
sethkarten: Fixed in ff15d7b06. Timeout and any `BaseException` (including cancellation/interrupt) share process-tree cleanup; collector and pipe-drain tasks are awaited or cancelled. Real-kernel cancellation and timeout tests verify descendants do not survive.

packages/coding-agent/src/core/tools/ipython.ts
sethkarten: Fixed in ff15d7b06 with concurrent bounded capture for both streams, explicit truncation/total-byte metadata, and a real-kernel large-output regression.

packages/coding-agent/src/core/prompts/rlm.ts
sethkarten: No code change needed for this withdrawn prompt comment; `bash()` uses Python asyncio subprocess APIs internally and the prompt correctly documents the helper rather than asking the model to reimplement it.

packages/coding-agent/src/core/tools/ipython.ts
```diff
@@ -34,6 +34,201 @@ try:
...
+        if self.returncode != 0:
+            text += chr(10) + chr(10) + f"Command exited with code {self.returncode}"
+        return text
+
+    def __repr__(self) -> str:
+        truncated = self.stdout_truncated or self.stderr_truncated
+        return (
+            f"BashResult(returncode={self.returncode}, stdout={len(self.stdout)} chars, "
+            f"stderr={len(self.stderr)} chars, truncated={truncated})"
+        )
```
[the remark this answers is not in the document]
sethkarten: Valid finding. Fixed in 785b9b75f: `BashResult.__repr__` now renders the same captured output as `str(result)`, so a cell ending in `await bash(...)` exposes command output through IPython's execute result. Added a real-kernel regression for the standalone expression.

packages/coding-agent/src/core/tools/ipython.ts
```diff
@@ -34,6 +34,199 @@ try:
...
+    while True:
+        chunk = await stream.read(_PRIME_AGENT_BASH_READ_CHUNK)
+        if not chunk:
+            return
+        capture.append(chunk)
+
+
+async def _prime_agent_bash_collect(proc, stdout_task, stderr_task) -> None:
+    await proc.wait()
+    await _prime_agent_asyncio.gather(stdout_task, stderr_task)
```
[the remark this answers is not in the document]
sethkarten: Fixed in 6b696f8ae. Collection now watches the shell return code without blocking on inherited pipes, terminates remaining session descendants on normal shell exit, and bounds pipe shutdown. Added a real-kernel `sleep &` regression proving the delayed child side effect never occurs.

packages/coding-agent/src/core/tools/ipython.ts
```diff
@@ -34,6 +34,199 @@ try:
...
+            parts.append(self.stderr)
+        text = chr(10).join(parts)
+        if self.returncode != 0:
+            text += chr(10) + chr(10) + f"Command exited with code {self.returncode}"
+        return text
+
+    def __repr__(self) -> str:
+        # IPython renders a cell's final expression with repr(), so expose the
+        # captured command output instead of hiding it behind a metadata summary.
+        return str(self)
```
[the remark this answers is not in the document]
sethkarten: Fixed in 6b696f8ae. Captured output now defaults to 32 KiB per stream so the combined IPython execute result remains below the 64 KiB display cap, and visible truncation notices include retained and total byte counts. The large-output kernel test verifies both streams and notices.

packages/coding-agent/src/core/prompts/rlm.ts
```diff
@@ -16,7 +16,7 @@ const IPYTHON_CONTROL_PROMPT = [
...
 	"",
-	"When running shell commands from IPython, use `%%bash` cells. If you use `%%bash`, it must be the first line of the code cell: no comments, spaces, blank lines, imports, or Python statements before it. Avoid `!cmd` shell escapes for project commands so shell behavior is explicit and multi-line commands share one shell context.",
+	'When running shell commands from IPython, prefer `await bash("...")` over `%%bash` cells: it keeps the kernel event loop responsive to interrupts, cleans up its managed process group on interruption, and bounds captured stdout/stderr. `%%bash` cells block the kernel until the command finishes. If you do use `%%bash`, it must be the first line of the code cell: no comments, spaces, blank lines, imports, or Python statements before it. Avoid `!cmd` shell escapes for project commands so shell behavior is explicit and multi-line commands share one shell context.',
```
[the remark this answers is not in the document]
sethkarten: Valid finding. Fixed in dd88b4629: the control prompt now recommends `await bash(...)` only on POSIX and explicitly tells Windows agents to use `%%bash`, matching the helper's fail-closed platform behavior.

packages/coding-agent/src/core/tools/ipython.ts
```diff
@@ -34,6 +34,230 @@ try:
...
+    """Run a shell command without blocking the IPython kernel event loop.
+
+    At most max_output_bytes from the tail of each output stream is retained;
+    both pipes are continuously drained so large output cannot deadlock or grow
+    the persistent kernel without bound. Truncation and total byte counts are
+    reported on the result.
+    """
+    if isinstance(max_output_bytes, bool) or not isinstance(max_output_bytes, int) or max_output_bytes <= 0:
+        raise ValueError("max_output_bytes must be a positive integer")
+    shell = _prime_agent_shutil.which("bash") or "/bin/bash"
```
snimu, line 208: `bash()` hardcodes the shell and skips the user's shell settings. The `%%bash` path goes through `applyShellSettingsToBashMagicCell()`, which applies the configured `commandPrefix` and `shellPath`. Users who set those get them silently ignored on the now-preferred path. Suggest either applying the same settings here or noting in the prompt that `bash()` ignores them.

packages/coding-agent/src/core/tools/ipython.ts
```diff
@@ -34,6 +34,230 @@ try:
...
+            return
+        capture.append(chunk)
+
+
+async def _prime_agent_bash_collect(proc, stdout_task, stderr_task) -> None:
+    # asyncio's process wait can remain pending while descendants hold inherited
+    # pipes, even after the shell's returncode is set. Observe that state directly.
+    while proc.returncode is None:
+        await _prime_agent_asyncio.sleep(0.01)
+    # A successful shell may leave background descendants in its session. Stop
```
snimu, line 136: This kills background jobs even on success — by design, and the tests pin it. But that's a real behavior difference from `%%bash`: with `bash()` you can never start a long-lived background process (e.g. `nohup server &`). The prompt says "prefer `await bash(...)`" without mentioning this, so a model following the doctrine will launch a dev server and find it dead. One extra clause in the prompt line would prevent that.

packages/coding-agent/src/core/prompts/rlm.ts
snimu, line 19: No streaming: `%%bash` shows output as it runs, `bash()` buffers everything until the command exits. For a long build the model and the user watching see nothing until the end — and since only the tail is kept, the beginning is gone. Fine for short commands, but the prompt recommends `bash()` unconditionally. Worth either mentioning "use `%%bash` for long-running commands where you want live output" here, or treating streaming as a follow-up.

snimu (commented):
Took a close look at the final state — the process-group cleanup and bounded capture are careful and the real-kernel tests cover the hard cases well. Three things I think are worth addressing before (or right after) merge, since they're all on the path the prompt now recommends:

PR #1189, opened 2026-08-11
fix(coding-agent): stamp agent messages with the sender's compose time

### LINKED ISSUE #823: Queued agent mail is delivered one message per assistant turn and carries no compose time, so a recipient acts on stale instructions late, one at a time, and blind to the backlog
## Summary

Inter-agent messages (`agent_message` / `send`) always use steering delivery — `streamingBehavior` is hardcoded to `"steer"`. An idle target receives immediately (`deliveryStatus: "delivered"`), but a message sent to a busy target is queued (`deliveryStatus: "queued"`) and sits there until the target's turn reaches a boundary. Two properties of that queued path compound:

**A. The queue drains one message per assistant turn.** Under the shipped default `steeringMode: "one-at-a-time"`, exactly one queued message is delivered per assistant turn regardless of queue depth. Delivery is in queue order and nothing is silently discarded (sends past the per-session cap of 20 pending, `DEFAULT_AGENT_MESSAGE_MAX_PENDING_PER_SESSION`, are rejected at send time with an error to the sender rather than dropped). But N children handing back results at the same time cost the parent N full model round-trips instead of one. The single `steeringMode` setting governs both human keyboard steering (where one-at-a-time is the right ergonomic default) and machine-generated agent mail (where it arguably is not), and there is no per-source override — `settings.json` exposes only `steeringMode` and `followUpMode`.

**B. The delivered prompt contains no compose time.** `createAgentSessionMessagePrompt` renders sender, route, and message id, but no time of any kind. A time is already known to the system — the sender's receipt carries `queuedAt`/`deliveredAt`, and `createAgentSessionMessage` stamps a transcript-level `timestamp` — but neither value is ever rendered into `content`, so it never reaches the receiving model.

Together: a message can wait an unbounded number of turns behind other queued mail and long-running work, and when it finally lands the recipient cannot tell it waited, cannot tell how much more is behind it, and cannot order it against its own work. Both halves are fixed in the same place — the delivered prompt text and the drain site — and any fix to either must respect the same fixed-offset prompt parser (see "Placement is load-bearing" below), so they are filed together.

## Environment

- prime-agent 0.7.0 — npm package `@earendil-works/pi-coding-agent` (`piConfig.name: prime-agent`)
- Component: `packages/coding-agent` — session input pump / agent mail delivery (`src/core/agent-messages.ts`, `src/core/agent-session.ts`, `src/modes/daemon/daemon-mode.ts`)
- All line numbers below are against `main` at commit `b9a4461149419156599d60174dddf15458e2b9ee`. The shipped 0.7.0 `dist/core/agent-messages.js` and `dist/modes/daemon/daemon-mode.js` match that source, and `dist/core/agent-session.js` carries the identical drain logic.

### A. One message per turn

1. Start a parent session in daemon mode.
2. From the parent, spawn 4 children via `rlm()` and give each a task that takes a similar amount of time, so they finish at roughly the same moment.
3. Have the parent enter a long-running cell after spawning (e.g. a `time.sleep(600)` poll loop), so it is busy when the children finish. This forces `shouldQueue` in `acceptAgentSessionMessage`.
4. Each child calls `agent_message` / `send` to hand its result back to the parent. All 4 messages are accepted with `deliveryStatus: "queued"`.
5. Observe the parent's transcript.

Each queued message is delivered on its own assistant turn. The 4th message is not visible to the parent until 3 further model round-trips have completed, each of which must also wait out whatever long cell the parent is currently executing.

Setting `steeringMode: "all"` in `settings.json` coalesces the backlog, but it also changes how interactive keyboard steering behaves, and it only coalesces adjacent actions whose execution policies match (`turnExecutionPoliciesEqual`, agent-session.ts L5445).

### B. No compose time

The recipient-visible text is produced by a single exported pure function, so the core of this is deterministic and needs no daemon:

```ts
// packages/coding-agent/test/compose-time.test.ts
import { createAgentSessionMessagePrompt } from "../src/core/agent-messages.js";

const prompt = createAgentSessionMessagePrompt({
	id: "agentmsg_demo",
	source: "agent_message",
	message: "Please inspect the latest result.",
	from: { activeSessionId: "active-parent", sessionId: "parent-id", sessionName: "lead" },
	fromRelationship: "parent",
	target: { activeSessionId: "active-child", sessionId: "child-id", sessionName: "worker" },
});
console.log(prompt);
```

End-to-end version, which shows the delay the missing field hides:

1. Start a session A, spawn a child B via `rlm(...)`.
2. Give B a task long enough that it streams and runs tools continuously for many minutes.
3. From A: `await agent_message.send("Please inspect the latest result.", receiver_role="child", receiver_name="<B>")`.
4. The receipt returns immediately with `deliveryStatus: "queued"` and a `queuedAt` timestamp — visible only to A.
5. Let B keep working. The queued steering prompt is delivered when B's turn next permits a session-input selection.
6. Inspect B's context at the moment the message lands: the rendered block is byte-identical to the block in the unit-level output below. Nothing in it distinguishes a message composed 2 seconds ago from one composed 40 minutes ago.

## Expected

**A.** Machine-generated agent mail should be coalescable independently of human keyboard steering. When several messages are already sitting in the queue at a turn boundary, the agent should either receive them together, or at minimum be told that more are waiting so it can choose to take a cheap drain turn instead of starting expensive work on stale information.

**B.** The delivered prompt states when the message was composed, so the recipient can reason about staleness — e.g. re-read the referenced artifact rather than assuming "the latest result" still means what the sender meant, or ignore an instruction that a later message has superseded.

## Actual

**A.** Exactly one queued message is delivered per assistant turn, regardless of queue depth, and the receiving agent gets no signal that more mail is pending.

Measurements from one long-running orchestrator session: 87 of 87 inbound deliveries arrived exactly one per assistant turn; four completed-child handback messages waited 39-96 minutes behind the orchestrator's long sleep cells before being seen.

**B.** Output of the snippet above, produced by running the 0.7.0 build of `createAgentSessionMessagePrompt`. This string is verbatim the `content` of the `AgentSessionMessage` appended to the recipient's transcript:

```
[from parent]
Agent-to-agent message received.
Source: agent_message
From: lead, active active-parent, session parent-id
To: worker, active active-child, session child-id
Message id: agentmsg_demo

Please inspect the latest result.
```

Sender, relationship, route, message id — and no time. The only user-facing timestamps are on the sender's side (`deliveredAt`/`queuedAt` on the receipt) or in receiver-host transcript metadata that is never serialized into the prompt text. A recipient has no in-band signal of message age.

### Shared delivery path

**Agent mail always uses steering delivery** — `packages/coding-agent/src/modes/daemon/daemon-mode.ts`, `acceptAgentSessionMessage` (L5491). The behavior is hardcoded, not configurable:

```ts
// L5511
const streamingBehavior = "steer";
// ...
// L5516
const didQueue = await session.queueAgentMessagePrompt(prompt, streamingBehavior, message);
```

with an in-code note that delivery is unbounded in time:

```ts
			// Do not await delivery: a queued message delivers only when the target's
			// turn progresses, and the sender is blocked inside its own turn — awaiting
			// here deadlocks mutual sends between busy sessions.
			return { status: "queued" };
```

`shouldQueue` (L5503-5510) is true whenever the target is mid-accept or mid-prepare for another agent message, or `session.isStreaming || session.isCompacting || session.isRetrying || session.isBashRunning || session.unfinishedActionCount > 0` — i.e. for essentially any actively working target.

**`"steer"` maps to the next-turn-boundary queue** — `packages/coding-agent/src/core/agent-session.ts`, `queueAgentMessagePrompt` (L4470) → `_queuePreparedPrompt("steer", ...)` (L4477) → `_deliveryPolicy` (L5071):

```ts
private _deliveryPolicy(schedule: SessionInputSchedule): DeliveryPolicy {
    return schedule === "steer" ? "next_turn_boundary" : "when_run_idle";
}
```

So the wait is bounded only by the target's own progress. This is also stated in `packages/coding-agent/skills/agent-message/SKILL.md` L43-48:

```
Messages always use steering delivery so
a busy target sees them during its active run. Returns a receipt with a
`deliveryStatus` field: `"delivered"` means the message reached an idle target's
context; `"queued"` means a steering message was accepted and will deliver when
the target's current work allows (`send` does not block waiting for that).
Delivered receipts carry `deliveredAt`, queued receipts carry `queuedAt`.
```

### A. One-per-turn drain

**The drain coalesces only when the mode is `"all"`** — `packages/coding-agent/src/core/agent-session.ts`, `_pumpSessionInputs` (L5402). `first` is taken, and additional queued actions are appended *only* inside the `mode === "all"` loop (L5438-5451):

```ts
const mode = first.delivery === "next_turn_boundary" ? this.steeringMode : this.followUpMode;
const actions: QueuedSessionAction[] = [first];
while (!preselected && mode === "all") {
    const next = this._actionStore.queuedActions(first.delivery)[0];
    if (
        !next ||
        next.payload.kind !== "turn" ||
        !turnExecutionPoliciesEqual(first.payload.executionPolicy, next.payload.executionPolicy)
    ) {
        break;
    }
    this._actionStore.selectFirst();
    actions.push(next);
}
```

With `one-at-a-time`, `actions` stays `[first]` and the remaining mail waits for the next turn boundary.

**The default** — `packages/coding-agent/src/core/settings-manager.ts`, `getSteeringMode` (L704):

```ts
return this.settings.steeringMode || "one-at-a-time";
```

**The knob exists but is framed purely as keyboard steering.** `docs/settings.md` L138 documents `steeringMode` as "How steering messages are sent". `docs/usage.md` L77 introduces it directly under the **Enter** / **Alt+Enter** keybinding list. `README.md` L201 describes `"one-at-a-time"` as "(default, waits for response)". No doc states that `agent_message` / `send` rides this same queue and is governed by this same setting — which is what makes the coupling surprising in multi-agent use.

### B. Missing compose time

`packages/coding-agent/src/core/agent-messages.ts`

`AgentSessionMessagePayload` (L112-120) has no compose-time field:

```ts
export interface AgentSessionMessagePayload {
	id: string;
	source: typeof AGENT_MESSAGE_SOURCE;
	message: string;
	from?: AgentSessionMessageSender;
	/** Sender relationship from the receiver's point of view. */
	fromRelationship?: AgentFamilyRelationship;
	target: AgentSessionMessageEndpoint;
}
```

`AgentSessionMessageDetails` (L122-128) likewise carries only `id`, `message`, `from`, `fromRelationship`, `target`.

`createAgentSessionMessagePrompt` (L388-405) emits every line the recipient sees; there is no time in the list:

```ts
	lines.push(`To: ${formatAgentSessionMessageEndpoint(payload.target)}`);
	lines.push(`Message id: ${payload.id}`);
	lines.push("");
	lines.push(payload.message);
	return lines.join("\n");
```

`createAgentSessionMessage` (L407-425) does accept a timestamp, but it lands only in transcript metadata — `content` is built from `payload` alone, so the value is never visible to the model:

```ts
export function createAgentSessionMessage(
	payload: AgentSessionMessagePayload,
	timestamp = Date.now(),
): AgentSessionMessage {
	return {
		role: "custom",
		customType: AGENT_MESSAGE_CUSTOM_TYPE,
		content: createAgentSessionMessagePrompt(payload),
		...
		timestamp,
	};
}
```

`createAgentSessionMessageReceipt` (L440-455) is where the only user-facing time lives, and it is returned to the **sender**, never transmitted to the receiver:

```ts
		...(status === "delivered" ? { deliveredAt: at } : { queuedAt: at }),
```

### A. Coalesce machine mail

1. **A separate coalescing policy for agent mail.** Add e.g. `agentMessageMode: "all" | "one-at-a-time"` (default `"all"`) and thread it through `_pumpSessionInputs` so the mode selection at agent-session.ts L5438 distinguishes agent-mail-originated actions from keyboard steering, rather than keying only on `first.delivery`. Queued actions already carry the originating `AgentSessionMessage` via `queueAgentMessagePrompt`, so the source is available at selection time. This keeps `one-at-a-time` as the human default while letting machine mail batch.

2. **Deliver a queue-depth hint with each message.** When a steering action is delivered, append a line such as `(N more messages queued)` to the prompt content. The count is available at the drain site from the same `this._actionStore.queuedActions(first.delivery)` accessor already used at L5441 — so this is a small change that lets an agent decide to take a cheap drain turn instead of starting long work on a partial picture.

3. **Document the coupling.** Add a note in `docs/settings.md` and `skills/agent-message/SKILL.md` that agent messages are delivered on the steering queue and that `steeringMode` therefore controls inter-agent mail batching, not just keyboard steering.

(1) and (2) are complementary: (2) is useful even with per-source batching, since the 20-message pending cap means a backlog can still exceed what one turn coalesces.

### B. Stamp compose time and render it

4. Add `composedAt?: string` (ISO-8601) to `AgentSessionMessagePayload` and to `AgentSessionMessageDetails`.
5. In `createAgentSessionMessage`, default it from the timestamp already being passed: `const composedAt = payload.composedAt ?? new Date(timestamp).toISOString();` then build `content` with `createAgentSessionMessagePrompt({ ...payload, composedAt })` and set `details.composedAt`.
6. In `createAgentSessionMessagePrompt`, append conditionally:

```ts
	lines.push(`Message id: ${payload.id}`);
	if (payload.composedAt) {
		lines.push(`Composed at: ${payload.composedAt}`);
	}
```

7. Optionally also stamp the payload at the point of send in `daemon-mode.ts` rather than at `createAgentSessionMessage`. `createAgentSessionMessage(payload)` is called at daemon-mode.ts L5512 with no explicit timestamp, so it defaults to `Date.now()` at the receiving host's accept moment; stamping at send would instead reflect the sender's compose moment. The two differ by the remote round-trip in `sendRemoteAgentSessionMessage` (L5443).

### Placement is load-bearing (applies to both new lines)

`parseAgentSessionMessagePromptId` (agent-messages.ts L367-382) parses this prompt by fixed line offsets and matches `/^Message id: (agentmsg_[^\n]+)$/` at exactly `toLineIndex + 1`:

```ts
	const toLineIndex = lines[offset + 2]?.startsWith("From: ") ? offset + 3 : offset + 2;
	if (!lines[toLineIndex]?.startsWith("To: ")) {
		return undefined;
	}
	const match = /^Message id: (agentmsg_[^\n]+)$/.exec(lines[toLineIndex + 1] ?? "");
```

Confirmed empirically against the 0.7.0 build: inserting a `Composed at:` line **after** the `Message id:` line still parses (`parseAgentSessionMessagePromptId` returns `agentmsg_demo`); inserting the same line **before** `Message id:` makes it return `undefined`.

Both the `Composed at:` line (fix 6) and the `(N more messages queued)` hint (fix 2) must therefore go **after** the `Message id:` line, or the parser has to be updated in lockstep. Breaking it is not cosmetic: `isAgentSessionMessagePrompt` feeds `clearQueuedUserMessagesMatching` in `daemon-mode.ts` (L3970, L5174), and `agent-session.ts` (L4464, L4475) falls back to `parseAgentSessionMessagePromptId` for `agentMessageId` round-tripping. Existing coverage in `packages/coding-agent/test/fire-and-forget-protocol.test.ts` and `packages/coding-agent/test/agent-session-bus.test.ts` should catch a regression here; a test asserting the new lines survive a parse round-trip would be worth adding.

## Impact

Both halves matter specifically for long-running multi-agent operation, which is the case `agent_message` exists to serve, and they reinforce each other: A makes messages wait, and B makes the wait undetectable.

From A, the cost scales with fan-out: a parent coordinating N children pays N model round-trips to ingest a single wave of handbacks, and each round-trip is serialized behind whatever the parent is currently executing. For orchestrators that poll in long cells, the measured delay between a child reporting completion and the parent acting on it was 39-96 minutes. The failure mode is quiet — the parent has no way to know that three more results are already sitting in its queue, so it may commit to a next step based on the first message alone, then re-plan once the rest arrive.

From B:

- Inter-agent messages are frequently deictic — "the latest result", "the file I just wrote", "stop what you're doing and check X". Their correctness depends on when they were written. A recipient that reads such a message with no time has to assume it is current, which is exactly wrong when the message has been sitting in a steering queue across many turns.
- Queued delivery latency is bounded only by the target's own progress, and it is worst precisely when it hurts most: the busier the target, the longer the queue wait, and the more its own state has diverged from what the sender was looking at when composing.
- The recipient cannot easily order two messages relative to its own work. If a correction arrives after the instruction it corrects, both render identically, and nothing in the prompt text tells the model which reflects the sender's newer state — and under A the correction always arrives on a *later* turn than the instruction, never alongside it.
- The fix looks cheap: a timestamp already exists in the object graph at message-construction time; it is simply not serialized into the text the model reads.

Questions for maintainers: is the coupling of agent mail to `steeringMode` intentional, and is there an existing reason compose time was left out of the prompt (context-size budget, determinism of transcripts for replay/testing) that a `composedAt` line would conflict with?

Related: #758 and #759 (parent cannot reliably tell when a child has finished) describe the same coordination gap from the sender's side; #762 covers discoverability of the `agent_message` / `agent_observe` surface.

Hotragn:
Fixes #823 (part B in full; part A's stated minimum is discussed below and deliberately left as a maintainer decision).

## The gap

Agent mail always uses steering delivery, so a message sent to a busy target is queued and waits until that target's turn reaches a boundary. The wait is bounded only by the target's own progress — the issue reports measured waits of 39–96 minutes for child handback messages sitting behind an orchestrator's long sleep cells.

`createAgentSessionMessagePrompt` rendered sender, relationship, route, and message id, and **no time of any kind**:

```text
[from parent]
Agent-to-agent message received.
Source: agent_message
From: lead, active active-parent, session parent-id
To: worker, active active-child, session child-id
Message id: agentmsg_demo

Please inspect the latest result.
```

A time was already known to the system — the sender's receipt carries `queuedAt`/`deliveredAt`, and `createAgentSessionMessage` stamps a transcript-level `timestamp` — but neither value is ever rendered into `content`, so it never reaches the receiving model. The recipient could not distinguish a message composed seconds ago from one that had waited an hour, and so could not judge whether "the latest result" still referred to what the sender meant, or whether a later message had superseded it.

## Approach, and the prior art it follows

This is the [RFC 5322 §3.6.1](https://datatracker.ietf.org/doc/html/rfc5322#section-3.6.1) `Date:` origination header applied to agent mail. That field is mandatory in Internet messages precisely because store-and-forward delivery decouples send time from read time: the transport records when the message was written, and the reader decides how old is too old. Nothing here is invented — the queued-agent-mail path has exactly the store-and-forward shape the header exists for.

It also matches the in-repo precedent for the other injected-prompt surface: scheduled heartbeat prompts carry their delivery context in the prompt body rather than in details the model never sees.

`composedAt` is stamped once, where the payload is minted in `sendAgentSessionMessage`, and rendered as a `Composed:` line. It is also surfaced on `AgentSessionMessageDetails` so transcript consumers and UI have it without re-parsing text.

## Placement is load-bearing

`parseAgentSessionMessagePromptId` walks the header by fixed offsets — optional `[from …]`, then `Agent-to-agent message received.`, `Source:`, optional `From:`, `To:`, `Message id:` — and stops at the id line. The new line is appended **strictly after** the line the parser stops on, so:

- stamped and unstamped prompts both still parse to the correct id, including the no-`From:`/no-relationship shape;
- a payload without `composedAt` renders text byte-identical to before, so nothing changes for callers that do not stamp.

Both properties are asserted, including an exact full-string equality check on the unstamped form.

## What I did not do, and why

Part A of the issue — the queue draining one message per assistant turn — is **not** in this PR. The issue's own "Expected" allows either coalescing or "at minimum … told that more are waiting", and I could not implement either half to the standard the rest of this change meets:

- **Coalescing** is governed by a single `steeringMode` setting that deliberately covers both human keyboard steering (where one-at-a-time is the right ergonomic default) and machine-generated agent mail. Splitting them means adding a per-source override to `settings.json`. That is a product decision about the settings surface, not a bug fix, and it belongs to whoever owns that surface.
- **Reporting backlog depth** needs an accurate count of agent-message actions queued ahead of this one. `unfinishedActionCount` counts all unfinished actions, not agent mail, so using it would print a number that is wrong in the common case. Doing it properly needs a narrow queued-agent-message accessor on `AgentSession`, and I could not exercise the daemon queue end-to-end on this machine (Unix-socket tests do not run on Windows), so I would be shipping an unverified count. A misleading number is worse than none.

Happy to follow up on either with direction on the settings shape.

## Tests

`packages/coding-agent/test/suite/regressions/823-agent-message-compose-time.test.ts` (5 cases): the stamped prompt renders the compose time and still ends with the message body; the fixed-offset parser resolves the id for both stamped and unstamped prompts and the `Composed:` line is positioned after `Message id:`; the id parses for the bare no-sender/no-relationship shape; an unstamped payload renders byte-identical text to before; and the compose time reaches transcript details while an unstamped message has no `composedAt` key.

Verified: `npm run check` clean. Every test file in the repo that references `createAgentSessionMessagePrompt`, `Agent-to-agent message received`, or `Message id:` was run — 436 passed. The 5 failures are all in `daemon-mode.test.ts` (unix-socket bind and symlink cases) and reproduce identically on a clean `main` worktree on this Windows machine.

> [!NOTE]
> ### Stamp agent-to-agent messages with the sender's compose time
> - Adds an optional `composedAt` field (ISO 8601) to `AgentSessionMessagePayload` and `AgentSessionMessageDetails` in [agent-messages.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/1189/files#diff-d56c5dcb3c4410339ae008806859f45d787d326d94c56b2e7c839dae397048b6).
> - When `composedAt` is present, `createAgentSessionMessagePrompt` appends a `Composed:` header line after the `Message id:` line so recipients can judge queue wait time.
> - The daemon sets `composedAt: new Date().toISOString()` on every outgoing message in [daemon-mode.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/1189/files#diff-f657fb72071b9bcb67235932ddc24164a0376dc918944a8f58b857c2ac43e450).
> - A regression test suite in [823-agent-message-compose-time.test.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/1189/files#diff-521c2f210819b348bf1d9e0d70426336d1e85107a4c20f283594e5b24499e443) verifies ordering, backward compatibility, and correct omission when no compose time is provided.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 607f235.</sup>
> 
>


kevinjosethomas:

PR #1191, opened 2026-08-11
docs(coding-agent): document the three states of a subagent reply flag

### LINKED ISSUE #758: [rlm] repliedSinceTask flag unreliable for determining subagent completion
## Problem

The `repliedSinceTask` flag on subagent entries in the family roster is inconsistent:

- `phonetic-variants` replied correctly, flag showed `True`
- `phrase-variations-2` wrote findings file but reply flag showed `False` (eventually became idle)
- `confusion-variants` replied correctly, flag showed `True` on first check, `False` on second
- `visual-similar` replied, flag showed `True`

The flag appears to be non-deterministic or has race conditions, making it unreliable for determining whether a subagent has completed its task.

## Impact

Parent agents cannot rely on `repliedSinceTask` to gate their execution. Manual polling or file-based completion detection is required instead.

## Recommendation

1. Fix the `repliedSinceTask` flag to be deterministic
2. OR add a reliable completion status field
3. OR document that the flag should not be relied upon

Hotragn:
Fixes #758.

## It is not a race

The issue reports `repliedSinceTask` behaving non-deterministically — the same child showing `True` on one poll and a falsy value on the next — and concludes the flag cannot be trusted for gating. The observed behaviour is real, but the cause is not a race.

The flag is **three-valued**: replied, has not replied, and *unknown*. Only two of those states are representable in a boolean, so the third is encoded by omitting the key:

```ts
// buildAgentFamilyRoster
...(relationship === "child" && entry.repliedSinceTask !== undefined
    ? { repliedSinceTask: entry.repliedSinceTask }
    : {}),
```

and it becomes `undefined` upstream whenever the daemon cannot read the child's live session:

```ts
// daemon-mode, building the family catalog
const repliedSinceTask = this.findSessionBySessionFile(agent.sessionPath)?.runtime.session.repliedToParentSinceTask;
return repliedSinceTask === undefined ? {} : { repliedSinceTask };
```

`findSessionBySessionFile` only resolves *resident* sessions. It is also `undefined` by construction for a child resumed in a new process, which `AgentSession` sets deliberately:

```ts
// A resumed child may have replied before this process started; false would
// claim knowledge that is not present in the session transcript.
```

So the reported transition is: the child replied (`True`), was evicted or passivated between the two polls, and the second poll returned a row with **no key** — which `entry.get("repliedSinceTask")` renders as `None` and a falsy check renders as "has not replied". The flag did not flip; the child left residency.

The reporter's other observations fit the same model. `phrase-variations-2` wrote its findings file but never messaged the parent, so `False` was correct — the flag answers "has it replied since I last wrote to it", not "is it done".

## Why documentation is the fix

Nothing anywhere described the third state. `repliedSinceTask` was absent from the `list_agents()` API description in `SKILL.md` entirely, so there was no way to learn it existed short of reading daemon internals — which is exactly how an orchestrator ends up writing `if not entry["repliedSinceTask"]` and waiting forever on a child that already answered.

`SKILL.md` now documents the three states, how to branch on them, why the same child can look like `True` then not-replied, and that completion needs `agent_observe` or an agreed output artifact rather than this flag. The interface carries a matching doc comment.

## What I deliberately did not change

I first made the key always present with `null` for unknown, then reverted it. Two reasons:

1. `daemon-mode.test.ts:809` explicitly asserts a passive child's roster row does **not** have the property. That is a deliberate, tested invariant, not an oversight.
2. For the Python callers this issue is about, it would be cosmetic: `entry.get("repliedSinceTask")` returns `None` for an absent key and for an explicit `null` alike. It would break a tested contract to change nothing observable for the affected consumer.

The change that would make the flag *determinate* rather than merely legible is to derive it for non-resident children instead of reporting unknown — the parent→child and child→parent messages are both recorded in the child's transcript, so a bounded tail scan could answer it without a live session. That is a real behavioural change with a per-roster-call cost, it needs the daemon exercised end to end (the unix-socket suites do not run on Windows, so I could not validate it here), and it is worth its own issue rather than being smuggled into a documentation fix. Happy to take it on with direction.

## Tests

`packages/coding-agent/test/suite/regressions/758-subagent-reply-state.test.ts` (4 cases) pins the contract at the roster boundary: replied and not-replied stay distinct booleans; unknown is an absent key and is asserted `not.toBe(false)`; unknown and not-replied remain separable within a single roster via `Object.hasOwn`; and non-child rows carry no flag.

Verified: `npm run check` clean. Every test file referencing `repliedSinceTask` or `buildAgentFamilyRoster` was run. The only source change is a doc comment, so it cannot alter runtime behaviour — the 9 remaining failures in that sweep are the `agents-view-state` and `daemon-mode` symlink/unix-socket cases that fail identically on a clean `main` worktree on this Windows machine.

> [!NOTE]
> ### Document three-valued semantics of `repliedSinceTask` on subagent roster entries
> - Adds `repliedSinceTask?: boolean` to the `AgentFamilyRosterEntry` interface in [agent-messages.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/1191/files#diff-d56c5dcb3c4410339ae008806859f45d787d326d94c56b2e7c839dae397048b6) with JSDoc explaining `true` (replied), `false` (has not replied), and absent (unknown, when the daemon cannot read the child's live session).
> - Expands [SKILL.md](https://github.com/PrimeIntellect-ai/prime-agent/pull/1191/files#diff-1544a1dd3b2a523dabd1e2a923c3059d3e084a296aae8a421a0d69ff7b186c45) with a dedicated 'Reading repliedSinceTask' section covering three-valued semantics, roster iteration examples, and the distinction between acknowledgement and completion.
> - Adds a regression test in [758-subagent-reply-state.test.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/1191/files#diff-9eeab3678f2a67b9dd3979ab94026cb7319902be093a99d55ac30feab8974bb2) asserting that unknown state is encoded as a missing key (not `false`), keeping unknown and not-replied distinguishable.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized a4a64d0.</sup>
> 
>


kevinjosethomas:

PR #1210, opened 2026-08-11
fix(coding-agent): use Windows venv interpreter

### LINKED ISSUE #1194: Windows: kernel bootstrap fails - venv interpreter path hardcodes bin/python (bootstrap.ts:728)
# Windows: kernel bootstrap fails — venv interpreter path hardcodes `bin/python` (bootstrap.ts:728)

## Summary

On Windows, first-run kernel bootstrap always fails. `uv` creates the kernel venv correctly using the
Windows layout (`Scripts\python.exe`), but `bootstrap.ts` computes the interpreter path as
`<venv>/bin/python`, which never exists on Windows. `uv pip install --python <that path>` then exits 2
and setup aborts.

Everything else about Prime Agent works on Windows once this is worked around — including the native
`zeromq` binding, which was the risk I expected to be hardest.

## Environment

| | |
|---|---|
| Prime Agent | 0.7.1 (tarball from `stable` channel, SHA256 verified) |
| OS | Windows 10 Pro 19045, x64 |
| Node | v22.23.2 |
| npm | 11.19.0 |
| uv | 0.6.10 |
| Python | 3.11.11 (uv-managed, `cpython-3.11.11-windows-x86_64-none`) |

## Reproduction

`install.sh` cannot run on Windows (see "Secondary findings"), so install the tarball directly:

```bash
npm install -g --allow-scripts=prime-agent,zeromq,@google/genai,koffi,protobufjs prime-agent-0.7.1.tgz
```

with `PRIME_AGENT_BOOTSTRAP_KERNEL_ON_INSTALL=1` and `PRIME_AGENT_INSTALL_UV=1` set.

## Actual result

```
› setting up python kernel (one-time, ~30s)…
rebuilding kernel venv
prime-agent: postinstall setup skipped: Failed to set up the Python kernel runtime.
C:\Users\melve\.local\bin\uv.exe pip install
  --python C:\Users\melve\.prime\agent\kernel-venv\bin\python
  ipykernel ...\dist\prime-agent-runtime dill requests httpx pyyaml tomli
  python-dotenv pandas numpy scipy beautifulsoup4 lxml pydantic tyro
failed with exit code 2
```

## Root cause

`packages/coding-agent/src/core/kernel/bootstrap.ts`

```ts
// line 728
const python = path.join(venv, "bin", "python");
// line 734
await run(uv, ["venv", venv, "--python", PYTHON_VERSION, "--seed"]);
```

Line 734 creates the venv correctly. Line 728 assumes a POSIX layout for the interpreter inside it.

## Evidence

The venv `uv` produced is valid and complete — only the path used to address it is wrong:

```
C:\Users\melve\.prime\agent\kernel-venv\
  Lib\  Scripts\  pyvenv.cfg  CACHEDIR.TAG  .gitignore

Scripts\python.exe   -> exists
bin\python           -> does not exist
```

## Workaround

Populate the venv manually with the correct interpreter path, then set the documented escape hatch:

```powershell
uv pip install --python "$env:USERPROFILE\.prime\agent\kernel-venv\Scripts\python.exe" `
  ipykernel "<prefix>\node_modules\prime-agent\dist\prime-agent-runtime" `
  dill requests httpx pyyaml tomli python-dotenv pandas numpy scipy `
  beautifulsoup4 lxml pydantic tyro

$env:PRIME_AGENT_KERNEL_PYTHON = "$env:USERPROFILE\.prime\agent\kernel-venv\Scripts\python.exe"
```

After this, `prime-agent doctor` runs clean, `import rlm` resolves, and `ipykernel` launches.

## Proposed fix

Matches the idiom already used elsewhere in the same file (line 503 appends `.exe` on win32):

```ts
const python =
  process.platform === "win32"
```

Any other place that addresses venv-internal executables (e.g. resolving `ipython`) needs the same
treatment — on Windows those live in `Scripts\` with an `.exe` suffix.

## Secondary findings

These are separate from the bug above; happy to split them into their own issues.

1. **`install.sh` is POSIX-only.** Its `uname -s` case handles only `Darwin` and `Linux` and exits with
   "Unsupported operating system". But the OS gate only guards *auto-installing Node* — the actual
   install is `npm install -g <tarball>`, which works on Windows unchanged. A documented manual path
   for Windows would unblock users today, ahead of any installer work.

2. **npm 11 blocks lifecycle scripts by default.** Installs need
   `--allow-scripts=prime-agent,zeromq,@google/genai,koffi,protobufjs`. Without it the install
   silently succeeds while skipping `zeromq`'s native binding fetch and prime-agent's own
   `postinstall`, leaving a broken install with no error. Note that passing the root package name for
   a *tarball* install does not appear to cover its own `postinstall` — that had to be run manually.

3. **No Windows job in CI.** Every job across the workflows is `runs-on: ubuntu-latest`. A
   `windows-latest` entry in the test matrix would have caught this bug.

4. **Native deps are fine on Windows — no compiler needed.** `zeromq` 6.5.0 fetched its `win32`
   prebuild and loads cleanly (`zmq version 4.3.5`), and `koffi` resolved its Windows prebuild
   without incident.

5. **The codebase already has substantial deliberate Windows support**, which is why this looks so
   close to working: `utils/shell.ts` detects Git Bash and falls back to any `bash.exe` on PATH with a
   genuinely helpful error; `bootstrap.ts:503`/`:518` handle `.exe` suffixes and `uv.exe`;
   `cli/daemon-ps.ts` has several win32 branches; `config.ts` uses the `win32` path module;
   `keybindings.ts` sets Windows-specific bindings. `fork-server.ts` correctly degrades on non-Linux.
   The `bin/python` line looks like the main thing standing between this and a working Windows install.

alloutflo:
## Summary

Fix Windows kernel bootstrap by resolving the virtual-environment interpreter from the platform-specific layout.

## Root cause

The bootstrap path always used `<venv>/bin/python`. `uv` creates `<venv>/Scripts/python.exe` on Windows, so both the initial package installation and later readiness checks addressed a path that does not exist.

## Changes

- Add one platform-aware kernel-venv interpreter path helper.
- Use it when bootstrapping the venv and when checking or reusing an existing venv.
- Add focused coverage for Windows and POSIX layouts.
- Document the fix in the coding-agent changelog.

## Verification

- `npx tsx ../../node_modules/vitest/dist/cli.js --run test/kernel-bootstrap.test.ts` — 22 tests passed.
- `npm run check` — formatting, linting, type checking, installer rendering, and browser smoke checks passed.

## Risk

Low. POSIX keeps the existing `bin/python` path. Windows now uses the standard `Scripts/python.exe` path. The platform selection is covered without requiring a Windows host, but this PR does not claim a full Windows end-to-end install run.

Fixes #1194

> [!NOTE]
> ### Fix kernel bootstrap to use Windows venv interpreter path
> The kernel bootstrap always used the POSIX `bin/python` path, causing failures on Windows where the interpreter lives at `Scripts/python.exe`.
>
> - Introduces `getKernelVenvPythonPath` in [bootstrap.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/1210/files#diff-e87b9a04ab9b4ac02e8dd33af2f2f58320d8b7f6be6853acf4a3e05a34623cae) to return the correct interpreter path based on platform (`Scripts/python.exe` on Windows, `bin/python` on POSIX).
> - Updates `bootstrapVenv` and `ensureKernelPythonUncached` to use this utility for dependency installation, readiness checks, and the final returned interpreter path.
> - Adds a unit test in [kernel-bootstrap.test.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/1210/files#diff-41f56d719cdb305c346ca238c1ec65971766309e330e3afc2812b0df018b63ab) covering both platform layouts.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 93b6ee7.</sup>
> 
>


kevinjosethomas:

PR #1216, opened 2026-08-11
Fix/operational recovery journal hardening

gonos2k:
### Harden operational recovery journals with tail repair, strict validation, and atomic writes
- Adds tail-repair logic to orphan process, worker recovery, and command recovery journals: on startup, any incomplete final append is either truncated or newline-terminated before new records are written.
- Strict JSON schema validation now rejects malformed or logically inconsistent complete records with a descriptive error (line number + reason) instead of silently skipping them.
- Introduces `writeAllSync` to guarantee full buffer writes by looping on short-write OS reports, used by all journal append and compact paths.
- Append paths now open with `'a+'`, capture original file size, fsync after write, and truncate back to the prior size on failure; compact paths use atomic temp-file rename with fsync.
- Adds automatic continuation of token-truncated assistant responses (up to 3 hidden prompts) in `AgentSession`, with a user-visible exhaustion notice when the cap is reached.
- Post-compaction maintenance is now time-bounded; failures surface as a warning message rather than stalling compaction. Scheduler suspension ownership is tracked through manual compaction and abort flows to prevent unintended turn starts.
- Risk: token estimation for non-ASCII and punctuation-heavy content is now higher than the previous `chars/4` heuristic, which shifts compaction thresholds and cut-point selection.

<sup><a href="https://app.macroscope.com">Macroscope</a> summarized fc9b6e4.</sup>


kevinjosethomas:

PR #1220, opened 2026-08-11
fix(kernel): serialize prewarm with root turns

paralin:
Kernel prewarm starts outside the root foreground context. A root turn can acquire the foreground lease while kernel startup is still preparing its bootstrap cell. Bootstrap then waits for the root-cell lease, while the turn's first IPython call waits for bootstrap to finish.

Acquire the root-cell lease before prewarm starts kernel initialization. An immediately following root turn now waits behind bootstrap instead of overtaking it. Abort the admission when the provisioner is disposed so a queued prewarm cannot outlive its session.

Retain a deterministic readiness-gate regression that proves the root turn queues behind prewarm without relying on scheduler timing.

> [!NOTE]
> ### Serialize kernel prewarm with root turns in `IpythonKernelProvisioner`
> When a `foregroundLease` is passed to `prewarm()`, the kernel bootstrap now runs under that lease (labeled `"root-cell"`), blocking any subsequent root turn from starting until bootstrap completes or is aborted. Without a lease, the prior background behavior is unchanged. A new test in [ipython-provisioner.test.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/1220/files#diff-acde112b894c2a7c08a290667fa9aa34fccd3aeb6a7044304afdd76c5a945ae0) validates that a root turn queued immediately after `prewarm()` waits until the readyGate is released.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 7e24f73.</sup>
> 
>

### DIFF
--- a/packages/coding-agent/src/core/tools/ipython.ts
+++ b/packages/coding-agent/src/core/tools/ipython.ts
@@ -353,9 +353,19 @@ export class IpythonKernelProvisioner {
 		return this._lastRestore;
 	}
 
-	/** Start the kernel in the background. Failures are swallowed here and surface on the next ensure(). */
+	/**
+	 * Start the kernel in the background. Foreground admission prevents the first root turn from
+	 * overtaking bootstrap and then waiting on it. Failures surface on the next ensure().
+	 */
 	prewarm(): void {
-		void this.ensure().catch(() => {});
+		const foreground = this.options?.foregroundLease;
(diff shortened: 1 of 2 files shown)


snimu:
Uses a hallucinated API, doesn't compile

paralin:
Ah, it's not hallucinated, just from my fork. Apologies.

PR #1225, opened 2026-08-11
fix(coding-agent): restore externally deleted supervisor owner records

irl-dan:
## Summary

A daemon that runs for several days can stop accepting work. Every new CLI launch then fails within seconds. The daemon looks healthy from the outside, so nothing replaces it. Only a manual shutdown clears the state. This change lets the supervisor repair itself and keep serving.

## Crash signature

Every CLI launch exits with code 1 after 1-2 seconds:

```
Error: Daemon supervisor generation <uuid> no longer owns its registry entry
    at deserializeDaemonError (packages/coding-agent/src/modes/daemon/daemon-errors.ts:34:9)
    at createDaemonClientConnection (packages/coding-agent/src/main.ts:1002:10)
```

## Root cause

The supervisor writes an owner record to the registry. The registry is in the OS temporary directory: `os.tmpdir()/prime-agent-<uid>/supervisor-owners/<generation>.owner/`. The owner record has two files: `owner.json` and `scope.json`.

OS cleaners delete old files in temporary directories. The macOS `/var/folders` cleaner removes files after about three days. It deletes the two record files in place. It keeps the record directory, the socket, and the lockfile directories. We observed this on a supervisor with four days of uptime. The same sweep emptied hundreds of unrelated temporary directories.

The supervisor validates the owner record for each command (`daemon-supervisor-ownership.ts:130-138`). After the deletion, validation always fails. The supervisor rejects every command with the error above.

The supervisor keeps its socket. It still sends a current-version `daemon_hello`. Client probes report the daemon as current. So no client starts a replacement. The daemon stays in this state until a person stops it.

Workers validate the same owner record every 250 ms (`daemon-mode.ts`). They drop their supervisor connections for the same reason.

## Reproduction

The steps are deterministic and need no API keys:

```bash
SANDBOX=$(mktemp -d /tmp/pa-repro.XXXX)
export PRIME_AGENT_INTERNAL_DAEMON_SUPERVISOR_REGISTRY_DIR=$SANDBOX/registry
export PRIME_AGENT_CODING_AGENT_DIR=$SANDBOX/agent
# 1. Start a sandbox supervisor.
prime-agent --mode daemon --daemon-socket $SANDBOX/daemon.sock --offline &
sleep 2
# 2. Delete the owner record files in place, as the cleaner does.
rm $SANDBOX/registry/*.owner/owner.json $SANDBOX/registry/*.owner/scope.json
# 3. Each client command now fails with the signature.
prime-agent --daemon-socket $SANDBOX/daemon.sock -p --no-skills --no-context-files -- 'hi'
```

On `main` (14d6e74), 15 of 15 concurrent CLI launches fail with the signature. With this fix, 0 of 15 fail. The supervisor logs one line: `Restored supervisor owner record for generation <uuid> after external removal`.

## Fix

The fix has two goals. A live supervisor must survive the loss of its owner record. A superseded generation must not replace its live successor.

- New method `DaemonSupervisorOwnership.restoreIfUnowned()`. It runs under the registry guard. It confirms that the owner record is absent. It then runs the same conflict scan as acquisition. If no live conflicting owner exists, it writes the owner record again with the same generation and token. If a live conflicting owner exists, it refuses. If a different record is present, it also refuses.
- `DaemonSupervisor.assertCurrentOwnership()` calls the restore when validation fails with `supervisor_generation_stale`. It then validates again. The restore is single-flight: concurrent commands share one scan. If the restore refuses, the original error propagates.
- The idle-eviction sweep runs the same ownership check. An idle supervisor heals without client traffic. Workers re-fence within 250 ms after the owner record returns.

## Tests

Three new cases in `test/suite/regressions/4600-supervisor-singleton.test.ts`:

- Restore succeeds after the record files are deleted in place. Restore also succeeds after the record directory is removed. A released ownership refuses to restore.
- A generation with a deleted record does not restore over a live successor.
- End to end: a supervisor with a deleted record continues to serve `list`. It writes the owner record again with the same generation and token.

Results: `4600-supervisor-singleton` 18/18 pass. `daemon-supervisor-process` 8 pass, 8 tag-skipped. `daemon-supervisor-eviction` 13/13 pass. `daemon-socket` 7/7 pass. `npm run check` passes.

### DIFF
--- a/packages/coding-agent/src/modes/daemon/daemon-supervisor-ownership.ts
+++ b/packages/coding-agent/src/modes/daemon/daemon-supervisor-ownership.ts
@@ -137,6 +137,46 @@ class DaemonSupervisorOwnership {
 		}
 	}
 
+	/**
+	 * Rewrite this owner's durable registry entry after an external removal.
+	 *
+	 * A long-running supervisor's owner.json lives under the OS temporary
+	 * directory, where system cleaners (for example the macOS /var/folders
+	 * reaper) can delete files out from under a live process. Losing the record
+	 * must not permanently poison the supervisor: it still holds the socket and
+	 * greets clients as current, so nothing would ever replace it.
+	 *
+	 * The record is restored only when it is genuinely gone and no other live
+	 * supervisor owns a conflicting scope, so a superseded generation can never
+	 * resurrect itself past its successor. Returns whether the durable record
+	 * once again matches this ownership.
+	 */
+	async restoreIfUnowned(): Promise<boolean> {
+		if (this.released || !matchesExactProcessIdentity(this.record)) {
+			return false;
+		}
+		return withDaemonSupervisorRegistryGuard(this.registryDir, () => {
+			const current = readOwnerRecord(this.ownerDirectory);
+			if (current) {
+				return sameOwnerRecord(current, this.record);
+			}
+			for (const directory of listOwnerDirectories(this.registryDir)) {
+				if (directory === this.ownerDirectory) {
+					continue;
+				}
+				const owner = readOwnerRecordForScope(directory, (scope) => ownerConflicts(scope, this.record));
+				if (owner && ownerConflicts(owner, this.record) && isProcessIdentityAlive(owner)) {
+					return false;
+				}
+			}
+			this.record.updatedAt = new Date().toISOString();
+			mkdirSync(this.ownerDirectory, { recursive: true, mode: 0o700 });
+			writeOwnerScope(this.ownerDirectory, this.record);
+			writeOwnerRecord(this.ownerDirectory, this.record);
+			return true;
+		});
+	}
+
 	async updatePhase(phase: DaemonSupervisorOwnerPhase): Promise<void> {
 		if (this.released) {
 			return;
--- a/packages/coding-agent/src/modes/daemon/daemon-supervisor.ts
+++ b/packages/coding-agent/src/modes/daemon/daemon-supervisor.ts
@@ -590,6 +590,7 @@ export class DaemonSupervisor {
 	private socketLease?: DaemonSocketPathLease;
 	private ownership?: Awaited<ReturnType<typeof acquireDaemonSupervisorOwnership>>;
 	private cleanupPromise?: Promise<void>;
+	private ownershipRestore?: Promise<boolean>;
 	private shuttingDown = false;
 	private updateRestartPhase?: "draining" | "fencing" | "prepared";
 	private readonly mutationDrain = new MutationDrainLatch();
@@ -792,6 +793,10 @@ export class DaemonSupervisor {
 
 	private async runIdleEvictionSweep(now = Date.now()): Promise<void> {
 		if (this.shuttingDown || this.updateRestartPhase !== undefined || this.idleEvictionFence) return;
+		// Piggyback an ownership self-check on the periodic sweep so an idle
+		// supervisor whose owner record was externally removed heals without
+		// waiting for the next client command.
+		await this.assertCurrentOwnership().catch(() => undefined);
 		await this.settingsManager.reload();
 		if (this.shuttingDown || this.updateRestartPhase !== undefined) return;
 		const idleEvictionMinutes = this.settingsManager.getIdleEvictionMinutes();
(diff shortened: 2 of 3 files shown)


packages/coding-agent/src/modes/daemon/daemon-supervisor-ownership.ts
```diff
@@ -137,6 +137,46 @@ class DaemonSupervisorOwnership {
...
+	 * The record is restored only when it is genuinely gone and no other live
+	 * supervisor owns a conflicting scope, so a superseded generation can never
+	 * resurrect itself past its successor. Returns whether the durable record
+	 * once again matches this ownership.
+	 */
+	async restoreIfUnowned(): Promise<boolean> {
+		if (this.released || !matchesExactProcessIdentity(this.record)) {
+			return false;
+		}
+		return withDaemonSupervisorRegistryGuard(this.registryDir, () => {
```
varindersingh83, line 158: Thanks for keeping this fix focused. There is a shutdown race here: `this.released` is checked before waiting for the registry guard. If `release()` removes the owner record while restoration is waiting, restoration can subsequently acquire the guard and recreate the record for the exiting supervisor.

Could we mark the ownership as `releasing` before `release()` waits for the guard, recheck that state inside this callback, and add a regression covering release while restoration is queued?

cc @irl-dan @snimu

nnunley:
These tests may be useful to you. No claim on the fix; they were written against `main` before I found this PR, and they cover two paths your change already handles but does not pin.

`nnunley/prime-agent@test/supervisor-registry-absence-proofs`, three files, tests only, no dependencies added.

**The coverage gap.** `readOwnerRecord` (`daemon-supervisor-ownership.ts:639-643`) returns `undefined` for three distinct causes: the file is missing, `JSON.parse` throws, and `isDaemonSupervisorOwnerRecord` rejects the shape. `restoreIfUnowned`'s `if (current)` is therefore false for all three, so your fix repairs a truncated or schema-invalid record as well as a deleted one. Your three new cases exercise deletion only, so the other two are fixed but unpinned. An exhaustive decision table over `assertCurrent` confirms they are indistinguishable today:

```
collapsed-causes (all three yield the identical observed outcome under a live owner):
  absent      -> DaemonSupervisorOwnershipLostError / supervisor_generation_stale
  unparseable -> DaemonSupervisorOwnershipLostError / supervisor_generation_stale
  wrong-shape -> DaemonSupervisorOwnershipLostError / supervisor_generation_stale
```

That table enumerates 210 fact-shapes over record state, every subset of the five fields `sameOwnerRecord` compares, pid liveness, and start-id agreement, constructed against a real on-disk registry and run against the real `assertCurrent`. It carries a reachability column: 6 of the 10 divergences are reachable, and they are exactly the three read failures under a live owner. The other 4 require assigning over `ownership.record.pid` after acquisition, which no production path can do, since the record is minted from `process.pid` at :309/:316 and the class is constructed only at :362.

**An end-to-end regression for the blast radius.** `1291-supervisor-registry-absence-e2e.test.ts` forks a real `DaemonSupervisor` via the existing `eng-4600-supervisor-fixture`, creates a real session over its socket with the faux provider, runs one prompt round trip, then issues a real `prepare_update_restart`:

```
healthy: {"success":true,  "total":1}
reaped:  {"success":false, "total":0, "error":"...no longer owns its registry entry"}
```

`total` is computed the way `package-manager-cli.ts` computes it during self-update, so it is the same quantity as the `{total:0,restored:0,resumed:0,failed:0}` in the field reports. The round trip matters: `createUpdateRestartSession()` discards an empty-draft session, so without it the healthy case also reports 0.

**Also on that branch**, aimed at the fence path rather than at your change: a 291-shape table over `assertDaemonSupervisorOwnerCurrent` (:365-390). 36 of its shapes are unreachable today and become reachable under any design that lets a rebuilt entry describe a process that is no longer alive — a record whose identity is provably dead currently raises ownership-lost rather than permitting takeover of a stranded socket. Your conflict scan refuses to restore over a live conflicting owner, which is the same question from the other side.

Take, adapt, or ignore. If any of it is useful I can rebase onto this branch and open it as a separate tests-only PR, or hand you the files to fold in.

kevinjosethomas:

PR #1234, opened 2026-08-11
fix(coding-agent): pin a stable locale for ps-based process start identities

itay-na:
## Problem

`getProcessStartId` reads the macOS/BSD start time from `ps -p <pid> -o lstart=`, which formats the date in the active locale. A start id recorded under one locale (or from a supervisor spawned in a different environment) no longer matches the same live process under another locale, so `processIdentity` returns `"replaced"` for a worker that is still running.

Recovery then refuses to signal the live worker (a replaced pid is treated as a recycled pid and never SIGKILLed), the worker keeps its socket, and any replacement fails with `Daemon socket already in use`. The session stays unavailable until the worker is killed by hand. A system locale change, or a supervisor relaunched from a shell with a different locale, is enough to trigger this.

## Fix

Run the `ps` query under a pinned `C` locale (`LC_ALL`, `LC_TIME`, `LANG`) so the start id is reproducible regardless of the system or caller locale. `process.env` is still spread so `PATH` (and the rest of the environment) is preserved.

The POSIX query is made injectable, mirroring `getWindowsProcessStartId`, so the locale pinning is covered by a unit test.

## What this does not change

- The Linux `/proc/<pid>/stat` path is already locale-independent and untouched.
- The Windows path is untouched.
- No stop, recovery, or ownership behavior changes. This only makes the identity stable so the existing identity checks stop misclassifying live workers as replaced.

## Note

Start ids recorded before this change under a non-C locale will not match ids computed after this change, so a worker that outlives an upgrade under a non-English locale is seen as replaced once and recovered (or relaunched from its saved transcript). English-locale installs already produce C-equivalent output and are unaffected. Per the project rule, no backward-compatibility shim is added.

## Validation

- `npm run check` passed (biome, tsgo, installer, browser-smoke).
- `npx tsx ../../node_modules/vitest/dist/cli.js --run test/session-lease.test.ts`: 9 passed (was 8; added a POSIX locale-stable case, skipped on Windows where the POSIX path is unreachable).

> [!NOTE]
> ### Pin C locale for `ps`-based process start identities in `getProcessStartId`
> On non-Windows systems, `ps -o lstart` output varies by locale, causing inconsistent process start identities. The fix passes `LC_ALL=C`, `LC_TIME=C`, and `LANG=C` via a new `POSIX_PROCESS_QUERY_ENV` constant when invoking `ps`, ensuring stable output regardless of the host locale. `runProcessQuery` is extended to accept an optional `env` parameter, and `getProcessStartId` supports dependency injection of the query function for testability.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 0c1e413.</sup>
> 
>

### DIFF
--- a/packages/coding-agent/src/core/session-lease.ts
+++ b/packages/coding-agent/src/core/session-lease.ts
@@ -112,10 +112,11 @@ function isProcessAlive(pid: number): boolean {
 
 type ProcessQuery = (command: string, args: string[]) => string;
 
-function runProcessQuery(command: string, args: string[]): string {
+function runProcessQuery(command: string, args: string[], env?: NodeJS.ProcessEnv): string {
 	return execFileSync(command, args, {
 		encoding: "utf8",
 		stdio: ["ignore", "pipe", "ignore"],
+		...(env ? { env } : {}),
 	});
 }
 
@@ -137,7 +138,13 @@ export function getWindowsProcessStartId(pid: number, query: ProcessQuery = runP
 	}
 }
 
-export function getProcessStartId(pid: number): string | undefined {
+type PosixProcessQuery = (command: string, args: string[], env: NodeJS.ProcessEnv) => string;
+
+// `ps -o lstart` formats start times in the active locale, so an identity recorded
+// under one locale can read as "replaced" under another. Pin a stable C locale.
+const POSIX_PROCESS_QUERY_ENV: NodeJS.ProcessEnv = { ...process.env, LC_ALL: "C", LC_TIME: "C", LANG: "C" };
+
+export function getProcessStartId(pid: number, query: PosixProcessQuery = runProcessQuery): string | undefined {
 	if (!Number.isInteger(pid) || pid <= 0) {
 		return undefined;
 	}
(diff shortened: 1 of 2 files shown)


packages/coding-agent/src/core/session-lease.ts
```diff
@@ -137,7 +138,13 @@ export function getWindowsProcessStartId(pid: number, query: ProcessQuery = runP
 	}
 }
 
-export function getProcessStartId(pid: number): string | undefined {
+type PosixProcessQuery = (command: string, args: string[], env: NodeJS.ProcessEnv) => string;
+
+// `ps -o lstart` formats start times in the active locale, so an identity recorded
+// under one locale can read as "replaced" under another. Pin a stable C locale.
+const POSIX_PROCESS_QUERY_ENV: NodeJS.ProcessEnv = { ...process.env, LC_ALL: "C", LC_TIME: "C", LANG: "C" };
```
varindersingh83, line 145: This stabilizes the locale, but `ps -o lstart` is still formatted using the active timezone. The same process can therefore produce a different identity after `TZ` changes.

Could we add `TZ: "UTC"` here and extend the regression to compare the identity under two different parent timezone values?

cc @itay-na @snimu

kevinjosethomas:

PR #1241, opened 2026-08-11
fix: fail closed on silent follow-ups and incomplete streams

ProDrifterDK:
## Summary

- Track every accepted parent-directed child task with a monotonic work epoch, including messages queued while the child is busy.
- Emit exactly one automatic notice only for the latest silent epoch, after session input and scheduled post-compaction continuation work are quiescent; explicit replies still suppress it.
- Reject OpenAI-compatible streams that reach EOF without a terminal `finish_reason`, while preserving partial content for diagnostics.

## Why

The retained-child notice previously observed one action-completion deferred, not the child session's logical quiescence. A newer parent message could be accepted and queued while the original task was streaming; when the original action settled, its completion callback sent `completed without sending a reply` even though the child was already processing the newer work. Busy-path messages also lacked equivalent monitoring, so merely suppressing the old notice could leave the parent waiting forever.

The child now binds each accepted direct, queued, or spawn task to a monotonic epoch. A completion/failure handler awaits the event-driven session idle barrier plus any timer-scheduled post-compaction continuation, then re-checks epoch, explicit-reply count, and disposal state. Completion monitoring is registered before queued admission so an immediately completed action cannot lose its settlement; rejected admission does not advance the epoch. The original spawn notice uses the same epoch/quiescence fence.

Separately, the OpenAI-compatible adapter initialized responses as successful stops and accepted clean iterator EOF even when no terminal chunk framed the completion. The incident transcript did not retain raw provider chunks, so this remains independent fail-closed hardening rather than a claimed incident trigger.

No daemon protocol, registry schema, or production polling behavior changes.

## TDD evidence

- Original coding-agent RED: silent retained follow-up had 0 terminal notices (97 passed, 1 failed).
- Exact incident RED: the original task emitted its notice while newer queued parent work was still streaming.
- AI RED: unterminated reasoning stream incorrectly emitted `done` (16 passed, 1 failed).
- Coding-agent GREEN after rebase: `agent-session-recursion.test.ts` — 102 passed, covering exact queued-work deferral, multiple queued epochs, immediate completion registration, post-compaction continuation quiescence, and explicit-reply suppression.
- AI GREEN: `openai-completions-tool-choice.test.ts` — 18 passed.
- Full gate after rebase onto `a3b3e7534`: `npm run check` passed, including Biome, `tsgo --noEmit`, installer render, and browser smoke checks.

> [!NOTE]
> ### Fix fail-closed behavior for silent RLM child follow-ups and incomplete OpenAI streams
> - OpenAI-compatible streams that end without any terminal `finish_reason` now emit an error event with `stopReason: 'error'` instead of being recorded as successful completions, tracked in [`openai-completions.ts`](https://github.com/PrimeIntellect-ai/prime-agent/pull/1241/files#diff-d3f9c03940767ab76191513078367a64d9e38e58fbdff99f5ad7bf65786b4246).
> - Retained RLM child follow-up notices are deferred until true session quiescence and suppressed if newer parent work arrives or the child replies, preventing out-of-order or duplicate notices in [`agent-session.ts`](https://github.com/PrimeIntellect-ai/prime-agent/pull/1241/files#diff-a72e8a0b5f514ca1def756ee863a6e8a556df115e7167a50ca4f5e66b14448ae).
> - Queued parent-directed messages to a retained child are also monitored; only the latest silent parent task emits one terminal/failure notice after the child is idle.
> - Post-compaction continuation cancellation now wakes quiescence waiters so notice deferral logic resolves correctly.
> - Behavioral Change: OpenAI-compatible streams that previously silently succeeded without a terminal `finish_reason` will now surface as errors.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized f9d2b29.</sup>
> 
>

### DIFF
--- a/docs/superpowers/plans/2026-08-11-retained-followup-incomplete-stream.md
+++ b/docs/superpowers/plans/2026-08-11-retained-followup-incomplete-stream.md
@@ -1,0 +1,135 @@
+# Retained Follow-up and Incomplete Stream Implementation Plan
+
+> Design: `docs/superpowers/specs/2026-08-11-retained-followup-incomplete-stream-design.md`
+
+## Goal
+
+Ship two independent fail-closed fixes on current `upstream/main`, preserve fire-and-forget agent messaging, open one minimal upstream PR, then cherry-pick and activate the functional commits on the existing local Prime Agent stack.
+
+## Fixed Point and Commit Shape
+
+- Base: `upstream/main` at `14d6e74919a5fba4916e5cc04b4f439a09a3750a`
+- Documentation commits already precede implementation.
+- Functional commit 1: `fix(coding-agent): report silent retained follow-ups`
+- Functional commit 2: `fix(ai): reject unterminated completion streams`
+- Keep the two functional fixes independently revertible.
+
+## Task 1: Silent Retained Follow-up Regression
+
+**Files**
+- Modify: `packages/coding-agent/test/agent-session-recursion.test.ts`
+- Read fully before editing: `packages/coding-agent/src/core/agent-session.ts`
+
+**Public seam**
+- `AgentSession.acceptAgentMessagePrompt`
+
+**RED**
+
+1. Create an RLM-depth-1 child with an agent-message controller whose `sendAgentMessage` calls are observable.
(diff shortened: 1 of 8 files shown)


kevinjosethomas:

PR #1269, opened 2026-08-12
fix(coding-agent): report session id alongside path to herdr

### LINKED ISSUE #1260: coding-agent: builtin herdr reporter does not register a resumable session ref
Package: coding-agent

**Description**

prime-agent's builtin herdr reporter (`packages/coding-agent/src/core/extensions/builtin/herdr-agent-state.ts`) reports agent lifecycle state to herdr (working/idle/blocked) via `pane.report_agent`, but never conveys the resumable session identity. As a result herdr cannot `prime-agent --resume` the session after a herdr-server restart; panes come back as fresh shells with no agent session to resume.

**Root cause (investigated, not yet fixed)**

The reporter reads the session ref from `ctx.sessionManager.getSessionFile()` / `getSessionId()` at `session_start`, but that value never resolves in practice, even after a session file has been persisted under the session store. Two attempts were made:
- refresh the ref before every `report_agent`, and
- poll after `session_start` and publish once a ref is available.

Both confirm lifecycle state reaches herdr, but the session ref never does (herdr's `agent_session` stays null on a pane despite a persisted session file). This points to the reporter being unable to obtain the active session file from its extension context in this runtime, not a herdr-side rejection.

**Steps to reproduce**

1. Run herdr (a terminal multiplexer) with a shell pane.
2. `herdr agent start caretaker --kind prime-agent --pane <pane-id>`
3. Send a trivial prompt so prime-agent persists a session file: `herdr agent prompt caretaker "Reply with exactly: OK" --wait`
4. Query herdr: `herdr pane get <pane-id>` — `agent_session` is null.
5. Restart the herdr server: the pane returns as a fresh shell; no `prime-agent --resume` is issued.

**Expected behavior**

The herdr reporter registers the active session file (or id) with herdr so `agent_session` is populated, and herdr relaunches `prime-agent --resume <ref>` on server restart.

**Actual behavior**

Lifecycle state is reported, but no session ref. herdr's `agent_session` remains null even with a persisted session file.

**Prime Agent version**

0.7.2

**Operating system**

Linux

**Additional context**

herdr's side is ready: its fork treats `(herdr:pi, prime-agent)` as an official resumable source and the resume planner would run `prime-agent --resume <ref>`. The missing link is prime-agent reporting the session ref. The reporter needs a reliable way to obtain and report the active session file once it is persisted (e.g. fixed at the first message persist, or via the session context).

witt3rd:
## Summary

The builtin herdr reporter captured the resumable session ref correctly but sent only `agent_session_path` when the session was persisted. Herdr's report handler accepts paths only for the `pi`/`omp` labels; for the `prime-agent` label it reads `agent_session_id`, so the ref was silently dropped and panes never became resumable (`agent_session` stayed null, no `prime-agent --resume` on herdr-server restart).

The reporter now sends both `agent_session_id` and `agent_session_path` on every report, so the session registers on current herdr (via id) and on path-accepting herdr (via path).

Fixes #1260.

## Root cause notes

- The ref itself resolves fine: daemon-created sessions are persisted via `SessionManager.create`, which assigns the session file at construction, before `session_start` fires. A socket tap on `HERDR_SOCKET_PATH` shows the first `pane.report_agent` already carrying `agent_session_path`. The drop happens herdr-side, in `session_ref_from_report`.
- prime-agent support in herdr currently lives on a fork (witt3rd/herdr); upstream herdr has no `("herdr:pi","prime-agent")` official source. The companion herdr-side fix (accept prime-agent paths, matching the snapshot/planner handling) is witt3rd/herdr@d524e4f1. This change stands alone: sending the id makes registration work regardless.

## Test plan

- New regression test `packages/coding-agent/test/suite/regressions/1260-herdr-session-ref.test.ts` drives the builtin extension against a real persisted `SessionManager` and a fake herdr socket, asserting both identity forms on the report (fails without the fix).
- Extended `test/herdr-agent-state.test.ts` assertions for both forms.
- `npm run check` clean.
- Verified end-to-end on an isolated herdr session: agent start → prompt → `agent_session` registered → herdr-server restart → pane relaunches `prime-agent --resume <ref>` with session content intact.

> [!NOTE]
> ### Fix herdr reporter to include session id alongside session path in reports
> Previously, `withSessionRef` in [herdr-agent-state.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/1269/files#diff-70837eb42752c61bc0c74d4ac79ed6c43a9d5c1cd8f2d90ea742d729f7dbbaa0) sent either `agent_session_path` or `agent_session_id`, but not both. Now both fields are included in `pane.report_agent` requests when available, allowing herdr to register a resumable session correctly.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized dafafe8.</sup>
> 
>

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## [Unreleased]
 
+- Fixed the builtin herdr reporter never registering a resumable session ref: reports now carry both `agent_session_id` and `agent_session_path`, since herdr only accepts the id for the `prime-agent` label ([#1260](https://github.com/PrimeIntellect-ai/prime-agent/issues/1260)).
+
 ## [0.7.2] - 2026-08-11
 
 - Fixed Down Arrow focusing the Agents View entry before moving a nonempty prompt cursor to the end ([ENG-5147](https://linear.app/primeintellect/issue/ENG-5147/keep-down-arrow-in-the-prompt-until-the-cursor-reaches-the-end)).
--- a/packages/coding-agent/src/core/extensions/builtin/herdr-agent-state.ts
+++ b/packages/coding-agent/src/core/extensions/builtin/herdr-agent-state.ts
@@ -205,13 +205,14 @@ function herdrAgentStateExtensionImpl(pi: ExtensionAPI, getLoadedExtensionPaths:
 	}
 
 	function withSessionRef(params: Record<string, unknown>): Record<string, unknown> {
-		if (currentAgentSessionPath) {
-			return { ...params, agent_session_path: currentAgentSessionPath };
-		}
-		if (currentAgentSessionId) {
-			return { ...params, agent_session_id: currentAgentSessionId };
-		}
-		return params;
+		// Report both identity forms. Herdr's report handler only accepts
+		// agent_session_id for the prime-agent label (path acceptance is
+		// pi/omp-only), so a path-only report never registered the session.
+		return {
+			...params,
+			...(currentAgentSessionId ? { agent_session_id: currentAgentSessionId } : {}),
+			...(currentAgentSessionPath ? { agent_session_path: currentAgentSessionPath } : {}),
+		};
 	}
 
 	function sendState(state: AgentState, message?: string, seq = nextReportSeq()): Promise<void> {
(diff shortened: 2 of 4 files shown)


jonaowen (changes requested):
The product change is reasonable, but the new regression is not cross-platform and will fail on native Windows. `startFakeHerdrServer(socketPath)` passes a normal `C:\...\h.sock` filesystem path to `net.Server.listen(path)`. Windows Node treats string IPC endpoints as named-pipe names and requires the `\\.\pipe\...` namespace; it does not create Unix-domain sockets at a temp filesystem path. The test has no `skipIf(win32)` and no platform endpoint helper. Please construct a unique `\\.\pipe\...` endpoint on Windows (and avoid `mkdir/rm` assumptions for it), or explicitly POSIX-gate this test while retaining platform-independent unit coverage. Add/execute the native-Windows focused test before merge.

kevinjosethomas:

PR #1278, opened 2026-08-12
fix(coding-agent): pickle callable Python skill modules by reference

### LINKED ISSUE #1211: Kernel state snapshot silently drops 5 built-in skills: cannot pickle '_PrimeAgentCallableSkillModule'
## Summary

The kernel state snapshot (`state-snapshot.ts` → `kernel-state.json` manifest) silently skips 5 built-in Python-backed skills because their module wrapper cannot be pickled by dill. These skills are therefore absent from saved/restored kernel state.

## Evidence (v0.7.1, WSL2/Ubuntu 24.04, kernel venv)

`~/.prime/agent/session-artifacts/<session>/kernel-state.json`:

```json
{
  savedNames: [agent_message, agent_observe, goal, rlm_heartbeat],
  skipped: [
    {name: attach_image, reason: TypeError:cannotpickle_PrimeAgentCallableSkillModuleobject},
    {name: compact, reason: TypeError:cannotpickle_PrimeAgentCallableSkillModuleobject},
    {name: edit, reason: TypeError:cannotpickle_PrimeAgentCallableSkillModuleobject},
    {name: refine, reason: TypeError:cannotpickle_PrimeAgentCallableSkillModuleobject},
    {name: websearch, reason: TypeError:cannotpickle_PrimeAgentCallableSkillModuleobject}
  ]
}
```

`state-snapshot.ts` iterates the kernel namespace and calls `dill.dumps(value)`, recording `skipped` with the reason. The `_PrimeAgentCallableSkillModule` wrapper (used for `compact`, `edit`, `refine`, `websearch`, `attach_image`) is not dill-serializable.

## Impact

- Skills are usable in the live session (they are imported into the running kernel), but they are dropped from the snapshot used for session restore/fork.
- `refine` being skipped is notable: the self-improvement loop's own kernel-side trigger is not part of the persisted kernel state.
- The skip is silent — nothing in the TUI or RPC event stream indicates that 5 built-in skills were not saved.

## Suggested fix directions

1. Provide a `__reduce__` / `__getstate__` implementation on `_PrimeAgentCallableSkillModule` so dill can serialize it (restoring by re-importing the skill module).
2. Or, treat built-in skill modules as re-importable by reference (like regular modules are) instead of value-serializing them.
3. At minimum, surface the `skipped` list somewhere observable (e.g. a session warning or RPC event) so users know the snapshot is incomplete.

15297839035:
Fixes #1211.

## Root cause

The kernel bootstrap wraps every Python skill that exposes `run()` in `_PrimeAgentCallableSkillModule`, a `types.ModuleType` subclass (`packages/coding-agent/src/core/tools/ipython.ts`). dill saves modules by reference, but pickle's dispatch table is keyed on the exact type, so the subclass fell through to the generic reduce path and raised `TypeError: cannot pickle '_PrimeAgentCallableSkillModule' object`.

This predicts the reported split exactly: the five skills listed as skipped (`attach-image`, `compact`, `edit`, `refine`, `websearch`) are the ones defining `run`; the four listed as saved (`agent-message`, `agent-observe`, `goal`, `rlm-heartbeat`) do not, so they stayed plain modules and pickled by reference.

## The part that actually loses data

The skill globals themselves are rebuilt on resume regardless — `restoreState()` runs before the bootstrap specifically so the bootstrap overwrites them with fresh live handles.

The real loss is that an unpicklable wrapper also poisons any user variable holding a reference to it. Reproduced in a real kernel:

```python
tools = {'doubler': demo_skill}
```

`tools` was reported in `skipped` and dropped, with no bootstrap to rebuild it.

## Fix

Give the wrapper a `__reduce__` that restores through `importlib.import_module`. The reducer names a stdlib function, so it pickles by reference rather than dragging the wrapper class into the payload — which would reintroduce the same failure.

## Tests

- `test/kernel-state-roundtrip.test.ts`: real-kernel round-trip covering both the skill module and a user variable referencing it. Written first and confirmed failing against the unfixed code, with both `demo_skill` and `tools` in `skipped`.
- `test/suite/regressions/1211-callable-skill-module-pickle.test.ts`: contract test on the generated bootstrap, so the guarantee also holds in the default shards where `kernel-heavy` is filtered out.

## Verification

- `npm run test:kernel` scope: 7/7 in `kernel-state-roundtrip.test.ts`.
- 117 tests pass across 9 related kernel and skill test files.
- `npm run check` passes.

Four failures in `kernel-goal-skill.test.ts` and `acp-kernel-features.test.ts` are pre-existing in my environment, not from this change: the same four fail at `HEAD` with the patch reverted. My kernel venv's `rlm` lacks the host bridge.

## Known limitation

A variable restored before the bootstrap holds the unwrapped module, so `tools['doubler'].run(21)` works but calling `tools['doubler'](21)` directly does not. Closing that gap means switching the bootstrap from copy-wrapping to an in-place `__class__` upgrade, which makes every existing reference callable and drops the `__dict__` copy. I verified that works but left it out — it changes module-wrapping semantics for all users and is beyond this issue. Happy to send it separately if you want it.

> [!NOTE]
> ### Fix pickling of callable Python skill modules in kernel state snapshots
> - Callable Python skill module wrappers (used to expose `run()` from Python skills) were silently dropped during kernel state snapshots due to a `TypeError` when `dill` tried to pickle them.
> - Adds a `__reduce__` method to the wrapper class in [ipython.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/1278/files#diff-812e64dd0562a31155c537ae8f665280f5201d993eea105d5066d0e7d02ee36d) so instances are pickled by reference via `_prime_agent_importlib.import_module(self.__name__)` instead of via the generic reduce path.
> - Adds an integration test in [kernel-state-roundtrip.test.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/1278/files#diff-0384e2afc0ba33cb48d0c8ec3a51a50c408444c4f9ae6418bf6ca83864801e68) covering snapshot/restore of callable skill modules and user variables referencing them.
> - Adds a regression test in [1211-callable-skill-module-pickle.test.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/1278/files#diff-12d21b9112d5772b26d6c7d6ebcf0eb97bef6c27961102c7522e55886ec86bac) asserting the reducer emits correctly and is absent when no Python skills are present.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized e026085.</sup>
> 
>

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## [Unreleased]
 
+- Fixed the kernel state snapshot silently dropping every Python skill that exposes `run()`, and any user variable referencing one, with `cannot pickle '_PrimeAgentCallableSkillModule' object` ([#1211](https://github.com/PrimeIntellect-ai/prime-agent/issues/1211)).
+
 ## [0.7.2] - 2026-08-11
 
 - Fixed Down Arrow focusing the Agents View entry before moving a nonempty prompt cursor to the end ([ENG-5147](https://linear.app/primeintellect/issue/ENG-5147/keep-down-arrow-in-the-prompt-until-the-cursor-reaches-the-end)).
--- a/packages/coding-agent/src/core/tools/ipython.ts
+++ b/packages/coding-agent/src/core/tools/ipython.ts
@@ -88,6 +88,13 @@ class _PrimeAgentCallableSkillModule(_prime_agent_types.ModuleType):
             return await result
         return result
 
+    # dill pickles modules by reference, but its dispatch is keyed on the exact
+    # type, so this subclass would otherwise fall through to the generic reduce
+    # path and raise TypeError - dropping the skill from the kernel state
+    # snapshot along with any user variable that references it.
+    def __reduce__(self):
+        return (_prime_agent_importlib.import_module, (self.__name__,))
+
 class _PrimeAgentUnavailableSkill:
     def __init__(self, name, error):
         self.__name__ = name
--- a/packages/coding-agent/test/kernel-state-roundtrip.test.ts
+++ b/packages/coding-agent/test/kernel-state-roundtrip.test.ts
@@ -1,9 +1,10 @@
 import { spawnSync } from "node:child_process";
-import { existsSync, mkdtempSync, rmSync } from "node:fs";
+import { existsSync, mkdirSync, mkdtempSync, rmSync, writeFileSync } from "node:fs";
 import { homedir, tmpdir } from "node:os";
 import { join } from "node:path";
 import { afterAll, beforeAll, describe, expect, it } from "vitest";
 import { KernelManager } from "../src/core/kernel/index.js";
+import { buildRlmBootstrapCode } from "../src/core/tools/ipython.js";
 
 /** Find a python that can launch an ipykernel and has dill, or null to skip. */
 function resolveKernelPython(): string | null {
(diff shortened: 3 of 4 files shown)


jonaowen (approved):
The exact-head change is narrowly scoped and matches Python pickle semantics for a `ModuleType` subclass: reconstructing through `importlib.import_module(module.__name__)` preserves by-reference identity without serializing the kernel-local wrapper class. The real-kernel round trip gates both the skill global and a container holding its reference across production-order restore then bootstrap, while the static regression keeps the reducer from drifting back to a local helper. I found no correctness or lifecycle blocker at `f5c7aa58b218cf3151241b065574ce12415f89b5`.

kevinjosethomas:

PR #1283, opened 2026-08-12
fix: PrimeIntellect-ai/prime-agent#1271

### LINKED ISSUE #1271: Bug: invalid interactive OAuth provider input crashes instead of reporting “Invalid selection”

Zewang0217:
Closes #1271

## What

Fixes the interactive provider selection prompt in the CLI so non-numeric input no longer crashes the process.

## Why

When a user typed something that was not a number at the provider prompt, `parseInt` returned `NaN`, which slipped past the existing range check (`index < 0 || index >= PROVIDERS.length`). The code then fell through to `PROVIDERS[NaN].id` and threw, instead of printing the intended "Invalid selection" message and exiting cleanly.

## How

Added a `Number.isInteger(index)` check to the validation condition so `NaN` (and any non-integer) is rejected up front, alongside the existing bounds check. Non-numeric or out-of-range input now prints "Invalid selection" and exits with code 1.

## Verification

- New Vitest suite `packages/ai/test/1271-cli-invalid-selection.test.ts` covers non-numeric, empty, and out-of-range inputs.
- Ran the package checks locally.

> [!NOTE]
> ### Fix CLI login to reject non-numeric provider selection input
> Adds a `Number.isInteger(index)` check in [cli.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/1283/files#diff-d407c2a03328fecec709113b81652fbf80f40a7b49ea75e3fa78c8cd0be2d320) alongside the existing range check, so non-numeric input (e.g. `abc`) now exits with code 1 and prints `Invalid selection` instead of crashing. A new Vitest suite in [1271-cli-invalid-selection.test.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/1283/files#diff-62a5afbe6f15d777b7fd96aa00a4be85def1085c2eb89227f4951d0c4c273db1) covers both non-numeric and out-of-range inputs.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized d6a973b.</sup>
> 
>

### DIFF
--- a/packages/ai/src/cli.ts
+++ b/packages/ai/src/cli.ts
@@ -104,7 +104,7 @@ Examples:
 			rl.close();
 
 			const index = parseInt(choice, 10) - 1;
-			if (index < 0 || index >= PROVIDERS.length) {
+			if (!Number.isInteger(index) || index < 0 || index >= PROVIDERS.length) {
 				console.error("Invalid selection");
 				process.exit(1);
 			}
--- a/packages/ai/test/1271-cli-invalid-selection.test.ts
+++ b/packages/ai/test/1271-cli-invalid-selection.test.ts
@@ -1,0 +1,33 @@
+import { spawnSync } from "node:child_process";
+import { createRequire } from "node:module";
+import { dirname, resolve } from "node:path";
+import { fileURLToPath } from "node:url";
+import { describe, expect, it } from "vitest";
+
+const require = createRequire(import.meta.url);
+const tsxLoader = require.resolve("tsx/esm");
+const packageRoot = resolve(dirname(fileURLToPath(import.meta.url)), "..");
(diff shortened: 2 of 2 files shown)


Zewang0217:
Closing: another contributor (kannishk) already has an open PR (#1282) for this issue. Avoiding duplicate work — our change was rejected by our own human review of the conflict.

PR #1287, opened 2026-08-12
fix(ai): reject invalid OAuth provider selections

### LINKED ISSUE #1271: Bug: invalid interactive OAuth provider input crashes instead of reporting “Invalid selection”

MJ-CJM:
## Summary

- Reject non-numeric interactive OAuth provider selections before indexing the provider list.
- Report `Invalid selection` instead of throwing a property access error.
- Add an isolated CLI regression test and changelog entry.

## Root cause

`parseInt()` returned `NaN` for non-numeric input. Because comparisons with `NaN` are false, the existing range check was bypassed and the CLI attempted to read `PROVIDERS[index].id`.

## Behavior

```text
Before: Error: Cannot read properties of undefined (reading 'id')
After:  Invalid selection
Exit code: 1
```

## Testing

- `npx tsx ../../node_modules/vitest/dist/cli.js --run test/cli.test.ts`
- `npm run check`
- Manually ran the built CLI, entered `abc`, and verified `Invalid selection` with exit code 1.

## Related

PR #1282 was opened while this fix was being prepared and also touches #1271 as part of a multi-issue change. This PR is intentionally limited to #1271 and includes focused regression coverage.

Fixes #1271

> [!NOTE]
> ### Fix `login` command to reject non-numeric OAuth provider selections
> Replaces `parseInt` with `Number()` and adds a `Number.isInteger` guard in [cli.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/1287/files#diff-d407c2a03328fecec709113b81652fbf80f40a7b49ea75e3fa78c8cd0be2d320) to validate provider selection input. Out-of-range or non-integer input now exits with code 1 and prints 'Invalid selection' to stderr instead of throwing a TypeError.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized a6e37ac.</sup>
> 
>

### DIFF
--- a/packages/ai/CHANGELOG.md
+++ b/packages/ai/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## [Unreleased]
 
+- Fixed interactive OAuth provider selection crashing on non-numeric input ([#1271](https://github.com/PrimeIntellect-ai/prime-agent/issues/1271)).
+
 ## [0.7.2] - 2026-08-11
 
 ## [0.7.1] - 2026-08-07
--- a/packages/ai/src/cli.ts
+++ b/packages/ai/src/cli.ts
@@ -103,8 +103,8 @@ Examples:
 			const choice = await prompt(rl, `Enter number (1-${PROVIDERS.length}): `);
 			rl.close();
 
-			const index = parseInt(choice, 10) - 1;
-			if (index < 0 || index >= PROVIDERS.length) {
+			const index = Number(choice) - 1;
+			if (!Number.isInteger(index) || index < 0 || index >= PROVIDERS.length) {
 				console.error("Invalid selection");
 				process.exit(1);
 			}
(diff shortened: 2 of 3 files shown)


jonaowen (approved):
The integer/range guard closes the NaN comparison bypass before indexing the provider list, preserves valid numbered selections, and the isolated CLI regression proves the exact exit/status/no-auth-write behavior. The scope is narrow and avoids the unrelated multi-issue tranche. No blocker found.

kevinjosethomas:

PR #1293, opened 2026-08-12
fix: extension setStatus dropped when it races the client's own attach

JackDanger:
Ran into this while writing a status-bar extension that shows live tok/s in the footer. `ctx.ui.setStatus(...)` called from `session_start` — the pattern the docs and `examples/extensions/status-line.ts` both show — basically never showed up. Not "sometimes flaky," just never.

Added some trace logging to find out why, and it's a straight race: `session_start` fires inside the worker and the extension calls `setStatus` before the client that's causing all this has finished attaching. `broadcastToSession` fans the update out over `state.clients`, which is still empty at that exact moment, so it just gets dropped. No error, nothing queued, no retry — the next tick of whatever timer/interval is driving the extension is the only thing that saves you, and if the extension only calls `setStatus` once (also a documented pattern), it's gone for good.

Logged timestamps to confirm: the broadcast with `clients=0` and the client's `attach` finishing are about 30ms apart, broadcast always loses.

**Fix**: keep the last value passed to `setStatus` per key on `ActiveSessionState` (`extensionStatusByKey`), and replay it to a client right after it's added to `state.clients` — both places that happens (`worker_subscribe`, and the real `attach` handler). Cleared on rebind/reload too, so a stale value from a previous extension instance can't get replayed to someone after a `/reload` swaps it out.

Left `notify()` alone on purpose — a one-shot toast doesn't mean anything replayed to a client that wasn't there when it fired. `setStatus` is different, it's meant to represent current state, so replaying the last value on attach is just... correct, not a workaround.

**Tests**: added to `daemon-mode.test.ts` (replay-on-attach, and that a rebind clears the recorded value) and `daemon-extension-binding.test.ts` (setStatus actually records what it's given). All existing tests still pass, `npm run check` is clean.

> [!NOTE]
> ### Fix extension `setStatus` being dropped when a client attaches after the call was made
> - When `ctx.ui.setStatus(key, text)` is called, the key/value pair is now recorded on `ActiveSessionState.extensionStatusByKey` before broadcasting.
> - When a client subscribes or attaches to an active session, `replayExtensionStatusToClient` sends all recorded `setStatus` values immediately, so the client sees current status without waiting for the next update.
> - On session rebind/reload, `extensionStatusByKey` is cleared so stale statuses are not replayed to future clients.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 3ea84d1.</sup>
> 
>

### DIFF
--- a/packages/coding-agent/src/modes/daemon/active-session-state.ts
+++ b/packages/coding-agent/src/modes/daemon/active-session-state.ts
@@ -41,6 +41,16 @@ export interface ActiveSessionState {
 	eventGeneration: string;
 	lastEventSequence: DaemonEventSequence;
 	unsubscribe?: () => void;
+	/**
+	 * Last value an extension passed to ctx.ui.setStatus(key, text) for this
+	 * session, keyed by status key (undefined entries mark a cleared key so a
+	 * newly attaching client doesn't resurrect it). A setStatus call made
+	 * synchronously from session_start races the calling client's own attach —
+	 * broadcastToSession sees zero clients yet and the update is dropped with
+	 * no retry. Recording it here lets a newly attached client be caught up
+	 * immediately instead of waiting on the extension's next unrelated update.
+	 */
+	extensionStatusByKey?: Map<string, string | undefined>;
 	/** Latest background status summary, surfaced in the agents view. */
 	summaryState?: AgentStatus;
 	/**
--- a/packages/coding-agent/src/modes/daemon/daemon-extension-binding.ts
+++ b/packages/coding-agent/src/modes/daemon/daemon-extension-binding.ts
@@ -199,7 +199,13 @@ function createExtensionUIContext(
 			),
 		notify: (message, notifyType) => emitUiRequest("notify", { message, notifyType }),
 		onTerminalInput: () => () => {},
-		setStatus: (key, text) => emitUiRequest("setStatus", { statusKey: key, statusText: text }),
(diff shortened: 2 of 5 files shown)


jonaowen (changes requested):
This moves the race rather than establishing attach ordering. Both insertion sites call `replayExtensionStatusToClient()` **before** writing/returning the successful attach response. A client that creates/routes its per-session UI only after attach succeeds can drop these replay events for exactly the same reason the original pre-attach broadcast was dropped. The tests call the private helper directly and never assert wire order through a real client.

Complete attach first, then deliver a generation-stamped snapshot/catch-up under the existing event-ordering contract (including backpressure and disconnect handling). Gate the exact wire sequence: attach response before status replay; no replay after failed/closed/replaced attach; concurrent `setStatus` during attach yields the latest value exactly once; snapshot/backpressure preserves ordering; and rebind generation prevents old values crossing into the new binding.

`extensionStatusByKey` is also unbounded in key count and text size. An extension can generate unique keys forever and retain arbitrary strings even with zero clients. Apply the public status key/value limits, cap aggregate state, delete cleared keys once all relevant clients have observed the clear (or use a bounded generation snapshot), and define collision ownership if multiple extensions use the same key. Durable current-state semantics are appropriate, but they need bounded, generation-bound delivery rather than pre-response best effort.

kevinjosethomas:

PR #1294, opened 2026-08-12
Add named sessions with --name, --list-sessions and --delete-session

porky11:
Adds a way to open or resume a session by name instead of by id or path.

- `--name <name>` opens the session with that name, resuming it if it exists and creating it otherwise.
- `--list-sessions` lists the sessions of the current directory with id, date and name.
- `--delete-session <name>` drops a name mapping without touching the session file.

Name lookup first checks the mapping file in the session directory, then falls back to scanning the sessions of the current directory for a stored name, so sessions named through `/name` are found as well. A stale id in the mapping falls back to the scan instead of failing.

Tests: `packages/coding-agent/test/named-sessions.test.ts`, four cases covering resolve, set, delete and auto-naming, isolated through a temporary session directory.

> [!NOTE]
> ### Add named sessions with `--name`, `--list-sessions`, and `--delete-session` CLI flags
> - `--name`/`-n` opens an existing session by name in the current directory, or creates a new one with that name if none exists.
> - `--list-sessions` prints all sessions for the working directory (id, modified date, name, path) and exits.
> - `--delete-session <name>` resolves a session by name and deletes it, reporting errors for ambiguous or missing names.
> - `findSessionByName` in [named-sessions.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/1294/files#diff-e3fd319b69ac8918f16862348a3eeab51e814eb587495b893517b5570fc6e9cb) filters sessions by name and throws `AmbiguousSessionNameError` when multiple sessions share the same name in one directory.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 2bfe652.</sup>
> 
>

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [Unreleased]
 
 - Fixed URLs not opening on click in fullscreen mode on terminals such as Ghostty; clicking a link in the transcript, dock, or overlays now opens it in the browser.
+- Added `--name` to open or resume a session by name within the current directory, plus `--list-sessions` to list that directory's sessions and `--delete-session` to delete a named session ([#1294](https://github.com/PrimeIntellect-ai/prime-agent/pull/1294) by [@porky11](https://github.com/porky11)).
 
 ## [0.7.2] - 2026-08-11
 
--- a/packages/coding-agent/src/cli/args.ts
+++ b/packages/coding-agent/src/cli/args.ts
@@ -22,6 +22,9 @@ export interface Args {
 	mode?: Mode;
 	daemonSocket?: string;
 	noSession?: boolean;
+	sessionName?: string;
+	listSessions?: boolean;
+	deleteSession?: string;
 	fork?: string;
 	sessionDir?: string;
 	models?: string[];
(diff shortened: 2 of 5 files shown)


jonaowen (changes requested):
The mapping is global (`getSessionsDir()/session-names.json`) but keyed only by name, while the user contract says sessions are scoped to the current directory. `--name demo` in project B first resolves the global mapping created in project A and `resolveSessionPath()` may return/open that foreign session (the `global` branch explicitly adopts its cwd). This silently crosses project/session authority. `listNamedSessions(_cwd)` even ignores cwd and admits the gap in a comment.

The store is also a racy read-modify-`writeFileSync` with no lock, atomic replace, schema validation, permissions, or corruption preservation. Concurrent processes can lose mappings; a crash can truncate the whole file; arbitrary JSON values are cast to strings. Names are not validated/bounded. Auto naming splits only `/`, so Windows `C:\repo` becomes the entire path and different semantics.

Use the existing session catalog as the single name authority rather than a second global index, or key an atomic locked index by canonical project identity + session directory + name. Reject ambiguous duplicate stored names rather than `find()`-selecting one; compare exact canonical cwd/session ownership before open. Add cross-project same-name, duplicate names, stale foreign mapping, Windows paths, symlink/case canonicalization, concurrent writers, partial/corrupt file, permission, and crash tests. `--delete-session` wording must say it deletes an alias (not a session), or actually perform the promised session lifecycle safely.

kevinjosethomas:

PR #1298, opened 2026-08-12
fix: PrimeIntellect-ai/prime-agent#1285

### LINKED ISSUE #1285: [pi-tui] KKP: ESC[?0u (unsupported) response still sets  _kittyProtocolActive, breaking legacy ESC-prefixed Alt keys
### Package

tui

### Description

Alt+letter keybindings (e.g. alt+t, alt+o) are silently dropped inside
 zellij, because the bundled TUI library (pi-tui) misinterprets the Kitty
 Keyboard Protocol (KKP) capability response.

 When prime-agent starts, pi-tui sends a KKP capability query (ESC[?u). In
 environments where KKP is not supported (here: zellij 0.44.3 with
 support_kitty_keyboard_protocol false), the terminal replies ESC[?0u (N=0 =
 unsupported). pi-tui enables KKP mode regardless of the N value —
 _kittyProtocolActive becomes true even for ?0u.

 Once KKP mode is (wrongly) active, parseKey() stops recognizing legacy
 ESC-prefixed sequences such as ESC t (Alt+t). Since zellij forwards
 physical Alt+letter presses as the legacy ESC <char> encoding, every
 Alt+letter shortcut is silently discarded. Plain Ctrl+letter shortcuts
 still work because single-byte Ctrl sequences (0x01–0x1a) have a separate
 parse path.

### Steps to reproduce

1. Run prime-agent 0.7.2 inside zellij 0.44.3 (config with
 support_kitty_keyboard_protocol false)
 2. Add an Alt+letter binding, e.g.:
     { "app.thinking.toggle": "alt+t" }
 3. Focus the input box, press Alt+t
 4. Nothing happens — the thinking blocks never toggle

### Expected behavior

alt+t toggles thinking blocks (and any other Alt+letter binding works as
 configured).

### Actual behavior

 The keystroke is silently dropped. Verified evidence:

 - A minimal probe sending ESC[?u in the same zellij pane receives ESC[?0u
 (KKP unsupported), yet prime-agent still enters KKP mode — confirmed by
 injecting ESC\r into prime-agent's stdin, which is only interpreted as
 shift+enter when _kittyProtocolActive === true
 - zellij forwards physical Alt+t to panes as the legacy sequence ESC t
 (verified with cat -v and a byte-level probe: 1b 74)
 - Injecting legacy ESC t directly into prime-agent's stdin does nothing
 (expected: it should map to alt+t)
 - Injecting the KKP-encoded ESC[116;3u does produce alt+t (pi-tui's
 modValue - 1 maps 3 → Alt)
 - Injecting ESC[116;2u (standard KKP: modifier 2 = Alt) yields shift+t
 instead of alt+t — the modValue - 1 in parseKittySequence shifts standard
 KKP modifier bits (1=Shift, 2=Alt, 4=Ctrl), so Alt becomes Shift, Ctrl
 becomes Alt+Shift, etc.

### Prime Agent version

0.7.2

### Operating system

Other

### Additional context

 Operating system
 WSL2 (Ubuntu) + Windows Terminal; zellij 0.44.3 as the terminal multiplexer

 Root cause is in the bundled pi-tui (TUI bundle, chunk-PMFPRFOT.js):

 1. queryAndEnableKittyProtocol() unconditionally writes ESC[?u, and the
 response handler enables KKP without checking the capability value:
     const kittyResponsePattern = /^\x1b\[\?(\d+)u$/;
     // ...
     if (match) {              // matches both ?0u and ?1u
       this._kittyProtocolActive = true;
       // ...
     }
   Fix: only enable when match[1] === "1".
 2. parseKey() only maps legacy ESC-prefixed Alt sequences while KKP is
 inactive:
     if (!_kittyProtocolActive && data.length === 2 && data[0] === "\x1B") {
       // ESC + letter → alt+letter
     }
   So the wrongly-enabled KKP mode disables the legacy path that zellij
 actually uses.
 3. parseKittySequence() returns modifier: modValue - 1, which is incorrect
 for standard KKP modifiers (1=Shift, 2=Alt, 4=Ctrl, 8=Super) — this breaks
 all CSI-u modifier combos independently.

 Workaround (applied locally, verified working): patch the bundled bundle to
 (a) not send the KKP query / only enable KKP when N === 1, and (b) remove
 the modifyOtherKeys fallback (ESC[>4;2m). The env var / config currently
 offers no way to disable the KKP query from prime-agent's side.

 Related upstream discussions (zellij side): zellij-org/zellij#5017,
 zellij-org/zellij#5191 (KKP modifier re-encoding).

Zewang0217:
## Fix Alt+letter shortcuts dropped inside zellij

Closes #1285

### Summary

Alt+letter bindings (e.g. `alt+t`) are silently ignored when running inside zellij (and similar terminals that don't support the Kitty Keyboard Protocol). This fixes two bugs in the bundled TUI key handling that caused it.

### Why it happens

Two separate bugs in `packages/tui`:

1. **KKP was enabled even when unsupported.** On startup we send the Kitty capability query (`ESC[?u`) and unconditionally entered Kitty mode on *any* response, including `CSI ? 0 u` (N=0 means unsupported — this is what zellij replies when `support_kitty_keyboard_protocol` is off). Once Kitty mode was wrongly active, `parseKey()` stopped recognizing the legacy `ESC <char>` sequences that zellij actually sends for physical Alt+letter presses, so every Alt+letter keystroke was dropped. Plain Ctrl+letter shortcuts were unaffected because single-byte Ctrl sequences take a separate parse path.

2. **CSI-u modifiers were decoded with the wrong encoding.** Kitty protocol reports modifiers as a bitmask (`shift=1, alt=2, ctrl=4, super=8`), but the code applied the xterm `modValue - 1` normalization used by `CSI 1;<mod>` modified arrows. The result: `CSI 116;2u` (standard KKP for Alt+t) parsed as Shift+t, and Ctrl became Alt+Shift, etc.

### Changes

- `packages/tui/src/terminal.ts` — only enable Kitty mode when the terminal replies `?1u`; a `?0u` reply is treated as unsupported and we keep the legacy parse path.
- `packages/tui/src/keys.ts` — decode CSI-u modifier values as a raw bitmask in both `parseKittySequence` and `decodeKittyPrintable`.
- Tests for both behaviors in `packages/tui/test/keys.test.ts` and `packages/tui/test/terminal.test.ts`.

### Verification

- New unit tests covering: `?0u` does not enable Kitty mode nor push flags, `?1u` does; standard KKP sequences parse correctly (`CSI 116;2u` → `alt+t`, `;4u` → `ctrl+t`, `;1u` → `shift+t`, combined bitmasks, `ctrl+-`).
- Existing test that relied on the old (wrong) modifier encoding was updated (`ctrl+shift+c` cyrillic sequence now uses modifier `4`).
- Full `packages/tui` test suite passes (`bun test`).

> [!NOTE]
> ### Fix `parseKey` and `matchesKey` to recognize legacy ESC-prefixed Alt sequences when Kitty protocol is active
> Some terminals advertise Kitty protocol support but still emit legacy ESC-prefixed sequences for Alt+letter/digit and Ctrl+Alt+letter combinations. Previously, `_kittyProtocolActive` suppressed parsing and matching of these sequences, causing those key combinations to go unrecognized.
>
> - [`keys.ts`](https://github.com/PrimeIntellect-ai/prime-agent/pull/1298/files#diff-c4967c48e0748404d8098865b5c26f3cde0386f443cdc0876af23a893ce7774a): Removes the `_kittyProtocolActive` guard from the ESC+control-char (Ctrl+Alt) and ESC+letter/digit (Alt) matching and parsing paths so these sequences are always accepted.
> - Alt+space and Alt+arrow sequences remain suppressed under Kitty to avoid ambiguity with other Kitty-encoded keys.
> - [`keys.test.ts`](https://github.com/PrimeIntellect-ai/prime-agent/pull/1298/files#diff-01758e1bdaec75181fe122dbda7b6a7a63579730dac019f7b5d688b6b21f9980): Adds a regression test verifying both legacy and CSI-u paths for Alt+letter/digit when Kitty is active.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 79a6154.</sup>
> 
>

### DIFF
--- a/packages/tui/src/keys.ts
+++ b/packages/tui/src/keys.ts
@@ -1172,15 +1172,17 @@ export function matchesKey(data: string, keyId: KeyId): boolean {
 		const isLetter = key >= "a" && key <= "z";
 		const isDigit = isDigitKey(key);
 
-		if (modifier === MODIFIERS.ctrl + MODIFIERS.alt && !_kittyProtocolActive && rawCtrl) {
+		if (modifier === MODIFIERS.ctrl + MODIFIERS.alt && rawCtrl) {
 			// Legacy: ctrl+alt+key is ESC followed by the control character.
-			// If that legacy form does not match, continue so CSI-u and
-			// modifyOtherKeys sequences from tmux can still be recognized.
+			// Kept even with Kitty protocol active — some terminals (e.g. zellij)
+			// answer the capability probe but still send legacy ESC+char.
 			if (data === `\x1b${rawCtrl}`) return true;
 		}
 
-		if (modifier === MODIFIERS.alt && !_kittyProtocolActive && (isLetter || isDigit)) {
-			// Legacy: alt+letter/digit is ESC followed by the key
+		if (modifier === MODIFIERS.alt && (isLetter || isDigit)) {
+			// Legacy: alt+letter/digit is ESC followed by the key.
+			// Kept even with Kitty protocol active — some terminals answer the
+			// capability probe but still send legacy ESC+char for Alt combos.
 			if (data === `\x1b${key}`) return true;
 		}
 
@@ -1311,12 +1313,14 @@ export function parseKey(data: string): string | undefined {
 	if (data === "\x1b\x7f" || data === "\x1b\b") return "alt+backspace";
 	if (!_kittyProtocolActive && data === "\x1bB") return "alt+left";
 	if (!_kittyProtocolActive && data === "\x1bF") return "alt+right";
-	if (!_kittyProtocolActive && data.length === 2 && data[0] === "\x1b") {
+	if (data.length === 2 && data[0] === "\x1b") {
 		const code = data.charCodeAt(1);
 		if (code >= 1 && code <= 26) {
 			return `ctrl+alt+${String.fromCharCode(code + 96)}`;
 		}
-		// Legacy alt+letter/digit (ESC followed by the key)
+		// Legacy alt+letter/digit (ESC followed by the key).
+		// Kept even with Kitty protocol active — some terminals answer the
+		// capability probe but still send legacy ESC+char for Alt combos.
 		if ((code >= 97 && code <= 122) || (code >= 48 && code <= 57)) {
 			return `alt+${String.fromCharCode(code)}`;
 		}
--- a/packages/tui/test/keys.test.ts
+++ b/packages/tui/test/keys.test.ts
@@ -415,7 +415,7 @@ describe("matchesKey", () => {
 			);
 		});
 
-		it("should parse legacy alt-prefixed sequences when kitty inactive", () => {
+		it("should parse legacy alt-prefixed sequences regardless of Kitty protocol state", () => {
 			setKittyProtocolActive(false);
 			assert.strictEqual(matchesKey("\x1b ", "alt+space"), true);
 			assert.strictEqual(parseKey("\x1b "), "alt+space");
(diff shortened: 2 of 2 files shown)


jonaowen (changes requested):
`CSI ? 0 u` does **not** mean unsupported. Under the Kitty keyboard protocol, `CSI ? flags u` reports the currently enabled flags; a compliant terminal with no flags enabled normally reports 0, which is exactly when the application should push its requested flags. Treating every 0 as unsupported disables Kitty negotiation on fresh conforming terminals and loses disambiguation/release/base-layout behavior. Zellij’s nontransparent synthetic response cannot be distinguished from a genuine zero-flags response by this value alone.

Also, the PR description claims a CSI-u modifier-decoding fix in `keys.ts`, but the actual head changes only `terminal.ts` and terminal tests; the advertised `CSI 116;2u`/combined-modifier regressions are absent.

Use a capability/transport strategy that does not reinterpret the protocol (for example, zellij-aware explicit configuration/probe behavior, tolerant legacy ESC parsing when safe, or upstream zellij support), and preserve zero as a valid flags state. Add real parser/integration gates for zero-flags Kitty terminals, zellij enabled/disabled, timeout/no response, split/coalesced responses, legacy Alt sequences, raw Kitty bitmasks, and stop/pop symmetry. Current tests merely encode the incorrect assumption.

Zewang0217:
Thanks for the thorough review — you're right on all points. I've rewritten the fix.

**On `?0u`:** Correct, `CSI ? flags u` reports the currently-enabled flags, not a support boolean. My `>0` heuristic conflated "zero flags pushed" with "unsupported" and would have broken Kitty negotiation on any conforming terminal that hasn't pushed flags yet. Reverted.

**On the CSI-u modifier claim in the old PR description:** That was a misdiagnosis on my part. The spec encodes modifiers as `1 + bitmask` (`shift+a → CSI 97;2u`), so the existing `modValue - 1` in `parseKittySequence` is correct. That change was already reverted in an earlier commit, but I failed to update the description. Sorry for the confusion.

**Root cause (was in the wrong file):** The actual bug is in `keys.ts`, not `terminal.ts`. When Kitty mode activates (correctly — on any capability response), the legacy `ESC + <char>` path for Alt+letter/digit is gated behind `!_kittyProtocolActive` and gets skipped entirely. Terminals like zellij answer the capability query (turning Kitty mode on) but still send legacy `ESC+char` for Alt combinations — those sequences are then dropped.

**New fix:** Remove the `!_kittyProtocolActive` gate from the legacy Alt+letter/digit parsing in both `matchesKey` and `parseKey`, so legacy `ESC+char` stays as a fallback even when Kitty mode is active. Real Kitty terminals send CSI-u for Alt combos and never produce bare `ESC+char`, so this doesn't interfere with them. `terminal.ts` is left unchanged from upstream — the capability negotiation (unconditional push on response) was correct as-is.

kevinjosethomas:

PR #1303, opened 2026-08-12
Fix transcript scrolling twice per PageUp/PageDown press

porky11:
Pressing PageUp scrolls the transcript up, and releasing the key scrolls it up again, so one keypress moves two pages. Same for PageDown.

Key release events are filtered only on the path to the focused component:

```ts
if (isKeyRelease(data) && !this.focusedComponent.wantsKeyRelease) return;
```

`handleFullscreenInput`, which evaluates `tui.viewport.pageUp`, `tui.viewport.pageDown`, `tui.viewport.top` and `tui.viewport.follow`, had no such guard. With the Kitty keyboard protocol enabled the terminal also sends a release event, it matches the same binding, and the viewport scrolls a second time. Terminals without the protocol are unaffected, which is why this is easy to miss. Reproduced on Alacritty.

The guard is placed after the mouse handling so mouse reports keep being consumed as before.

The regression test in `packages/tui/test/fullscreen.test.ts` was verified against the unpatched code: it fails without the fix and passes with it.

> [!NOTE]
> ### Fix transcript scrolling twice per PageUp/PageDown press on Kitty protocol terminals
> Terminals using the Kitty keyboard protocol send both key press and key release events, causing PageUp/PageDown to scroll the transcript twice. A guard in the [TUI fullscreen input handler](https://github.com/PrimeIntellect-ai/prime-agent/pull/1303/files#diff-0aed3a4b69095fa2ae0cffa8689c7dad409e8b36ab77544b0defb1474fde7311) now calls `isKeyRelease()` and returns early for release events, so each keypress scrolls exactly once.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized a0bff26.</sup>
> 
>

### DIFF
--- a/packages/tui/CHANGELOG.md
+++ b/packages/tui/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [Unreleased]
 
 - Fixed hyperlinks not being clickable in fullscreen mode on terminals that gate native link handling while mouse reporting is active (e.g. Ghostty); left-clicking a link now opens it directly.
+- Fixed the transcript scrolling twice for a single PageUp or PageDown press on terminals using the Kitty keyboard protocol.
 
 ## [0.7.2] - 2026-08-11
 
--- a/packages/tui/src/tui.ts
+++ b/packages/tui/src/tui.ts
@@ -994,6 +994,10 @@ export class TUI extends Container {
 
 		if (overlayFocused || !fullscreen.viewportControls) return false;
 
+		// A key release repeats the same binding under the Kitty protocol and would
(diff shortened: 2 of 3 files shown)


jonaowen (approved):
The guard is correctly placed after mouse handling and before every fullscreen viewport keybinding, so Kitty release reports cannot repeat page/top/follow actions while focused-component release policy remains unchanged. The virtual-terminal regression drives the real release sequence and demonstrates one scroll per press. No blocker found.

kevinjosethomas:

PR #1304, opened 2026-08-12
fix(goals): park after empty continuation turns

nnunley:
## Summary

- park an active goal when an assistant turn has no non-whitespace text or tool call
- keep the goal active so later external input resumes it normally
- hide repeated continuation prompt panels and show the existing continuation count in the fixed lower tray

## Why

An active goal currently requests another model turn after every natural stop. If the model returns an empty/whitespace-only assistant message while waiting for input, Prime Agent immediately injects another visible `<goal_context>`. This creates a visible wake loop and advances the transcript despite no actionable result.

This change treats that empty result as a quiescent boundary. It does not pause, clear, complete, or error the goal, and it does not change persisted or wire schemas.

## Semantics

- Quiescent: no trimmed text and no tool call (thinking-only is also quiescent).
- Terminal error/abort behavior is unchanged.
- `continuationsUsed` increments only when an actual continuation prompt is issued.
- Continuation contexts remain in model/session history with `display: false`.
- Budget-limit and objective-update contexts remain visible.
- Later user/agent input starts a normal run against the still-active goal.

## Tests

- focused Vitest: 196 passed
- TypeScript: `tsgo --noEmit`
- Biome 2.5.5 on all changed TS files
- installer render check
- browser smoke check

`npm run check` initially resolved a stale hoisted Biome 2.3.5 from the parent checkout, which cannot parse this branch's 2.5.5 config. Running the pinned 2.5.5 CLI directly passed.

> [!NOTE]
> ### Park active goals after empty assistant continuation turns
> - `AgentSession._getContinuationMessages` now returns `[]` when the latest assistant message has no tool calls and no non-empty text (detected via new `_isQuiescentGoalBoundary` helper), preventing active goals from looping on empty turns.
> - Continuation goal context messages are hidden from display; only actionable goal contexts remain visible.
> - The lower-tray goal label now shows a wake count (e.g. `· 2 wakes`) for active goals using `continuationsUsed`.
> - Behavioral Change: active goals that previously auto-continued after empty or thinking-only assistant turns will now park until new input arrives.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized f5a2b1c.</sup>
> 
>

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [Unreleased]
 
+- Fixed active goals repeatedly waking after an empty assistant turn; goals now remain active until new input arrives, with continuation wakes coalesced into the lower-tray counter.
 - Fixed URLs not opening on click in fullscreen mode on terminals such as Ghostty; clicking a link in the transcript, dock, or overlays now opens it in the browser.
 
 ## [0.7.2] - 2026-08-11
--- a/packages/coding-agent/src/core/agent-session.ts
+++ b/packages/coding-agent/src/core/agent-session.ts
@@ -3195,11 +3195,23 @@ export class AgentSession {
 		}
 	}
 
+	private _isQuiescentGoalBoundary(message: AssistantMessage): boolean {
+		if (message.stopReason === "error" || message.stopReason === "aborted") {
+			return false;
+		}
+		return !message.content.some(
+			(content) => (content.type === "text" && content.text.trim().length > 0) || content.type === "toolCall",
+		);
+	}
+
 	private async _getContinuationMessages(
 		context: GetContinuationMessagesContext,
 		signal?: AbortSignal,
 	): Promise<AgentMessage[]> {
 		if (this.queuedActionCount > 0) {
+			return [];
+		}
+		if (this._goalState.status === "active" && this._isQuiescentGoalBoundary(context.message)) {
 			return [];
 		}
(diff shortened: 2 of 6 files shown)


jonaowen (changes requested):
This stops the immediate empty-output continuation loop, but as written it converts it into a silent indefinite park while the goal still reports `active`. That is a semantic-liveness regression for unattended goals: status heartbeats do not wake it, persistent transcript activity can still look healthy, and restart has no durable park reason or wake policy. It also does not address retry-exhausted errors, which still deactivate persistent goals.

Specific issues:

- `continuationsUsed` is displayed as “wakes,” but external/user/heartbeat wake runs do not increment that counter. The label is inaccurate.
- The initial `/goal` context also uses `kind="continuation"`, so this hides the first objective-bearing context, not only repeated panels.
- Thinking-only and anomalous normal-stop empty responses are both treated as intentional quiescence despite different recovery needs.
- There is no durable `quiescent` state/reason, last substantive progress, delayed-retry marker, or restart/idempotent wake contract.

Please preserve the loop breaker but model the park explicitly, or distinguish intentional wait from anomalous empty output and give the latter a bounded durable continuation. Rename the counter to “continuations” unless a true wake count is added, keep the initial goal context visible, and add restart plus status-heartbeat/scheduled-heartbeat/external-message exactly-once wake coverage. This should be scoped as resource-loop suppression rather than unattended-liveness improvement.

kevinjosethomas:

PR #1309, opened 2026-08-12
fix(acp): title IPython tool calls by their cell and show the source

### LINKED ISSUE #1307: ACP: every IPython tool call is titled "IPython cell" and shows no cell source
### Summary

In an ACP client every IPython tool call is titled `IPython cell`, and the cell source is only in `rawInput`. A client that renders content blocks rather than raw input shows a column of identical, empty rows — there is no way to tell which cell is which, or what any of them ran.

### Details

`packages/coding-agent/src/modes/acp/acp-events.ts` maps `tool_execution_start` with a constant title:

```ts
title: event.toolName === IPYTHON_TOOL_NAME ? "IPython cell" : event.toolName,
```

Since IPython is the model-facing tool, that constant is the title of essentially every tool call in the session. The source goes out only as `rawInput`, which ACP does not require a client to render, and `tool_execution_end` sends just the result text.

There is a second, subtler effect: a client replaces a tool call's content on update instead of appending, so even a client that does show the start content loses it as soon as the call completes.

### Proposal

- Title the call by the cell's first meaningful line, appending `· +N lines` for the rest. Pair a leading cell magic with the following line, since `%%bash` alone names the interpreter rather than the work.
- Send the cell source as a fenced content block on `tool_call`, and repeat it on `tool_call_update` so it survives completion.

This is a display change in the ACP mapping only; nothing in the tool contract changes.

### Environment

prime-agent 0.7.1 and `main` at 0987c1b, Zed as the ACP client, macOS.

AndriyPytel:
## Context

Discussion: #1355. Issue: #1307.

Every IPython tool call reaches an ACP client titled `IPython cell`, with the source only in `rawInput`. Since IPython is the model-facing tool, that is the title of nearly every row in the session, and a client that renders content blocks rather than raw input shows those rows empty.

## Changes

- Title the call by the cell's first meaningful line, with `· +N lines` for the rest. A leading cell magic is paired with the following line, because `%%bash` on its own names the interpreter rather than the work.
- Send the source as a fenced content block on `tool_call`, and repeat it on `tool_call_update`: a client replaces a tool call's content on update instead of appending, so without the repeat the cell disappears the moment the call completes.
- Hold the in-flight cell in `AcpEventMappingState`, which already exists per session for bash correlation, and delete the entry when the call ends so the map cannot grow.

## Validation

- `test/acp-events.test.ts` covers the title for a plain cell, a `%%bash` cell, a blank cell, and a 200-character line, plus completion carrying both the cell and the result with the state entry released.
- `npm run check` passes.

fixes #1307

> [!NOTE]
> ### Title IPython tool calls by their cell and include cell source as fenced content
> - Adds `ipythonCellTitle` to derive a sanitized, truncated title from the first meaningful line of a cell, with a `+N lines` suffix when additional lines exist.
> - Adds `ipythonCellContent` to wrap cell source in a fenced code block, using a backtick fence longer than any run in the source to prevent fence escapes.
> - Cell source is stored in mapper state on `tool_execution_start` and included as fenced content on both start and completion events; state is cleared on `agent_end`.
> - Behavioral Change: IPython `tool_call` updates now carry a derived title and fenced content block instead of generic defaults; clients that replace content on update (rather than append) will see cell source on completion.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 086b88d.</sup>
> 
>

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## [Unreleased]
 
-- Changed ACP IPython tool calls to be titled by their cell and to carry the cell source as content, so a client no longer shows a column of identical "IPython cell" rows with nothing in them.
+- Changed ACP IPython tool calls to be titled by their cell and to carry the cell source as content, so a client no longer shows a column of identical "IPython cell" rows with nothing in them ([#1309](https://github.com/PrimeIntellect-ai/prime-agent/pull/1309) by [@AndriyPytel](https://github.com/AndriyPytel)).
 - Fixed URLs not opening on click in fullscreen mode on terminals such as Ghostty; clicking a link in the transcript, dock, or overlays now opens it in the browser.
 
 ## [0.7.2] - 2026-08-11
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [Unreleased]
 
+- Changed ACP IPython tool calls to be titled by their cell and to carry the cell source as content, so a client no longer shows a column of identical "IPython cell" rows with nothing in them.
 - Fixed URLs not opening on click in fullscreen mode on terminals such as Ghostty; clicking a link in the transcript, dock, or overlays now opens it in the browser.
 
 ## [0.7.2] - 2026-08-11
--- a/packages/coding-agent/docs/acp.md
+++ b/packages/coding-agent/docs/acp.md
@@ -40,7 +40,7 @@ Session activity arrives as `session/update` notifications:
 | tool finishes | `tool_call_update` (`completed` / `failed`) |
 | shell output | `tool_call` plus incremental `tool_call_update` |
 
-IPython is Prime Agent's model-facing tool, so a cell is a `tool_call` of kind `execute` whose `rawInput` carries the cell source.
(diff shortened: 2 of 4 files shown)


jonaowen (changes requested):
The source should not be wrapped in a fixed triple-backtick fence without escaping. An IPython cell may legally contain ``` (including Markdown strings or copied untrusted text); that closes the fence and lets subsequent source render as arbitrary ACP Markdown outside the code block, so the visible “source” can be structurally deceptive. Choose a fence longer than every backtick run, or use an ACP content representation that is explicitly plain/code text. Add regressions for triple/long backtick runs, tildes, and HTML/Markdown payloads.

Also, `ipythonCells` is cleaned only on `tool_execution_end`. Interrupted/cancelled sessions or a missing terminal event retain complete cell source for the life of `AcpEventMappingState`, contradicting “cannot grow.” Clear per-call entries on every terminal/cancel/session-reset path and bound the map before insertion; gate a start-without-end followed by teardown. Cell titles should normalize control characters/bidi isolates before placing source-derived text into a UI title. The underlying goal of showing source is sound, but the rendering and lifecycle boundaries need to be explicit.

kevinjosethomas:

Muzosh:
Hello, @kevinjosethomas 

Could you please review discussion https://github.com/PrimeIntellect-ai/prime-agent/discussions/1355, which introduces the problem this PR is trying to solve?

PR #1310, opened 2026-08-12
Fetch the live OpenRouter model catalog

kristianernst:
OpenRouter adds and drops models between releases, but the picker still came from the snapshot baked in at generate time. This fetches the live catalog when you open the model picker or run `model list`, and keeps the snapshot as the offline / failure / first-load fallback.

## Why this is a new PR

[#804](https://github.com/PrimeIntellect-ai/prime-agent/pull/804) did the same job on an older main, and it also tried to invent live reasoning-effort maps. That second part landed independently in [#1258](https://github.com/PrimeIntellect-ai/prime-agent/pull/1258) (`getOpenRouterReasoningCapabilities`). Rebasing #804 as-is would have reintroduced a weaker parser: omitted `supported_efforts` meant “every gateway effort,” while #1258 treats that as toggle-only and `null` as every effort.

This branch is based on current main and uses that helper. The remaining user-visible change is live catalog membership (new models appear, removed ones leave).

## Behavior

- Live fetch on catalog queries, session restore of an OpenRouter id, saved OpenRouter defaults, and `--model openrouter/...`. Other startup paths do not wait on the network.
- 5 minute cache, 3 second timeout, one in-flight request, 30 second backoff after failure. Restore of a missing OpenRouter id bypasses TTL and backoff so a stale cache cannot hide a newly listed model.
- For ids already in the snapshot, live overlay keeps snapshot `cost`, `maxTokens`, `contextWindow`, and `thinkingLevelMap` (the generate-models corrections such as Kimi K2.5’s 4096 output cap). Snapshot `featured` / `headers` / extra compat are merged. Virtual aliases (`auto`, `~*-latest`) stay even when the filtered endpoint omits them.
- `models.json` overrides still win.
- Generator `fetchOpenRouterModels()` now calls the shared parser so the next snapshot regen cannot drift from runtime parsing. This PR does not regenerate `models.generated.ts`.

Closes #804.

> [!NOTE]
> ### Fetch the live OpenRouter model catalog and merge it with the snapshot registry
> - Adds [openrouter-model-catalog.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/1310/files#diff-826a7bde17d30273d5f74ccb977e7c9735958ff4dc6cc6c05b71d559433ee627) to fetch and cache the live OpenRouter catalog with a 5-minute TTL, in-flight deduplication, 3-second timeout, and 30-second failure backoff.
> - Adds `resolveOpenRouterCatalog` in [model-registry.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/1310/files#diff-2fd8c4a2672b4ce27e73d494479df612c8b30ebac3dda5a160e2d861c612f453) to merge the live catalog with the generated snapshot, using live membership/capabilities while preserving snapshot cost, limits, and virtual aliases as fallback.
> - Extracts a shared [openrouter-models.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/1310/files#diff-72b94a81b18b69003032ae3e45c50805864bb85275242ccd014789a860b797ce) parser (`parseOpenRouterModels`) that normalizes catalog entries: filters non-tool and batch variants, converts pricing to $/million, clamps `maxTokens` to `contextWindow`, and applies DeepSeek-specific compat overrides.
> - Session restore, startup model resolution, daemon rehydration, and the `list-models` CLI command all call `findOrFetch` or `refreshOpenRouterModels` to opportunistically load OpenRouter models missing from the snapshot.
> - Risk: live network requests are made during model registry initialization and session restore; failures are swallowed and fall back to the snapshot, but may add latency on slow or offline connections.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 2b95c10.</sup>
> 
>

### DIFF
--- a/packages/ai/CHANGELOG.md
+++ b/packages/ai/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## [Unreleased]
 
+- Added a shared OpenRouter catalog parser used by live discovery and model generation.
+
 ## [0.7.2] - 2026-08-11
 
 ## [0.7.1] - 2026-08-07
--- a/packages/ai/scripts/generate-models.ts
+++ b/packages/ai/scripts/generate-models.ts
@@ -5,6 +5,7 @@ import { homedir } from "os";
 import { dirname, join } from "path";
 import { fileURLToPath } from "url";
 import { getAnthropicCacheCosts } from "../src/cache-pricing.js";
+import { parseOpenRouterModels } from "../src/openrouter-models.js";
 import { getOpenRouterReasoningCapabilities } from "../src/openrouter-reasoning.js";
 import {
 	CLOUDFLARE_AI_GATEWAY_ANTHROPIC_BASE_URL,
@@ -15,7 +16,6 @@ import {
 import {
 	Api,
 	type AnthropicMessagesCompat,
-	KnownProvider,
 	Model,
 	type OpenAICompletionsCompat,
 } from "../src/types.js";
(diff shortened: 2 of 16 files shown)


jonaowen (changes requested):
Live membership is reasonable, but new-model economic authority fails open. `toCost()` maps missing, malformed, zero, and OpenRouter’s negative “unknown” sentinels all to `0`, then the registry publishes that live-only model as costing $0. The picker/usage accounting cannot distinguish a genuinely free route from unknown pricing. A partial/migrating catalog response can therefore make a paid model appear free and drive selection/spend under false authority.

For live-only models, require complete finite nonnegative provider pricing with an explicit known/free distinction before publication, or carry `unknown` through the model contract and render/block cost-dependent selection accordingly. Preserve the last complete catalog when a response is only partially valid rather than treating any one valid entry as an authoritative full replacement that drops every other model.

There is also a `requireId` race: if a general refresh is already `inFlight`, `getOpenRouterModels(..., requireId)` returns that promise even if its result lacks the required id, so exact session/default restoration does not perform the promised bypass/refetch. Coalesce requirements or re-check/refetch after the shared request. Add gates for unknown/negative/missing pricing, mixed partial catalog responses, and concurrent general-refresh plus exact-ID restore. The snapshot merge policy should be documented as intentionally preferring stale generated economic/limit metadata for known IDs.

kevinjosethomas:

PR #1315, opened 2026-08-12
fix(daemon): close interrupted tool calls

ProDrifterDK:
## Summary

- close unresolved tool calls in the active branch's terminal `toolUse` turn during daemon recovery
- persist synthetic error results before the existing worker-recovery marker without replaying uncertain work
- keep partial persistence resumable and retain busy recovery-journal records when catalog persistence fails

The synthetic result preserves the original tool call ID and name and states that execution, its result, and external side effects are unknown. Recovery does not automatically resume or retry the operation.

## Validation

- `npm run check`
- `npx tsx ../../node_modules/vitest/dist/cli.js --run test/session-manager/interrupted-tool-results.test.ts test/daemon-catalog-process.test.ts test/daemon-supervisor-monitor.test.ts`
- 92 focused tests passed

## Scope

No daemon protocol/schema changes, provider transformation changes, automatic resume, or tool replay.

> [!NOTE]
> ### Close interrupted tool calls during worker session recovery
> - Adds exact-authority worker recovery: when a daemon worker is killed mid-tool-use, unresolved tool calls in the terminal assistant turn are closed by appending synthetic error `toolResult` messages before the session is marked recovered.
> - Introduces `ExactRecoveryAuthority` snapshots in `SessionManager` capturing session lineage, head entry, assistant entry, and unresolved tool call identities, journaled as v2 busy checkpoints in `WorkerRecoveryJournal`.
> - Catalog recovery (`markSessionInterrupted`) is now lease-fenced, serialized per session path, and returns a structured `CatalogRecoveryResult` (applied / already_applied / stale) with idempotency keyed on `operationId`.
> - Supervisor shutdown and restart now require a `DaemonShutdownAuthority` derived from the connection's `daemon_hello` handshake; commands missing or mismatching authority are rejected with a typed `shutdown_authority_rejected` error.
> - `shutdownDaemonAndWait` and `runShutdownAllConverging` in `daemon-ps` now distinguish authority-rejected outcomes from stopped/unavailable, preserving protected supervisor scopes instead of force-killing them.
> - Risk: schema revision bumped from 16 to 18; supervisors at revision ≥17 require authority on all public shutdown/restart commands, breaking legacy clients that send bare `{ type: 'shutdown' }`.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized c29d8e6.</sup>
> 
>

### DIFF
--- a/docs/superpowers/plans/2026-08-13-exact-turn-worker-recovery-plan.md
+++ b/docs/superpowers/plans/2026-08-13-exact-turn-worker-recovery-plan.md
@@ -1,0 +1,99 @@
+# Exact-turn worker recovery implementation plan
+
+**Design:** `docs/superpowers/specs/2026-08-13-exact-turn-worker-recovery-design.md`
+
+## Test seams
+
+The approved design fixes three observable seams:
+
+1. `WorkerRecoveryJournal`: durable v2 authority epochs and conditional completion.
+2. `SessionManager`: exact-authority, resumable closure of one assistant tool-use turn.
+3. `DaemonCatalogClient.markInterrupted`: lease-fenced, idempotent catalog recovery result consumed by the supervisor.
+
+Tests exercise these public module boundaries and persisted JSONL output rather than private helpers.
+
+## Commit 1 — journal exact recovery authority
+
+Files:
+
+- `packages/coding-agent/src/modes/daemon/worker-recovery-journal.ts`
+- `packages/coding-agent/src/modes/daemon/daemon-mode.ts`
+- `packages/coding-agent/test/worker-recovery-journal.test.ts`
+- focused daemon-mode tests only if needed
+
+Red/green slices:
+
+1. Parse v2 records and ignore legacy busy records for mutation authority.
+2. Keep one stable `operationId` during a busy epoch; allocate a new id after idle.
+3. Deduplicate only identical authority, including session generation, head, assistant, tool calls, and lineage.
+4. Capture authority from one active-branch snapshot without discarding tool event identity.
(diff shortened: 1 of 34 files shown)


jonaowen (changes requested):
Recovery is not bound to the interrupted turn. The journal/request supplies only `sessionPath`, `activeSessionId`, and operation names; `markSessionInterrupted()` reopens whatever active branch exists at handling time and `closeUnresolvedToolCalls()` searches its **latest** assistant turn. If the session was resumed, branched, or advanced after the worker died but before/during catalog recovery, this can close a new owner’s live tool calls as “interrupted.” `activeSessionId` is merely copied into marker details and never validated against an exact session epoch/head/tool-turn identity.

The marker append is also not idempotent. If catalog persistence succeeds but its response is lost, the supervisor retains the busy journal and retries; the second call sees no unresolved calls but appends another `prime-agent.worker_recovery` marker. The tests cover result→marker partial failure, not committed-request/response-loss replay.

Record the exact active lineage/session generation and assistant entry/head (plus a stable recovery operation ID) in the worker journal before execution. At recovery, compare-and-append only against that exact unchanged turn; mismatch must stale nothing and must not mutate newer work. Make the whole recovery operation idempotent by operation ID, including the marker. Gate advance/branch/session-replacement races, same-path newer tool-use, response loss followed by replay, concurrent duplicate requests, and partial multi-result persistence. Unknown side-effect wording is good, but it cannot compensate for missing mutation authority.

kevinjosethomas:

PR #1316, opened 2026-08-12
fix(coding-agent): preserve goal continuation after compaction

samsja:
## Summary

- Retarget this PR from the retired async-bash stack to current `main`.
- Fix active goals stalling when another threshold compaction starts while the prior goal continuation is still running.
- Keep deduplication for pending continuation actions, but allow a running continuation to queue its successor.
- Stabilize the regression fixture with separate threshold and overflow boundaries and replace fixed sleeps with a bounded condition wait.

## Root cause

`_queuedGoalThresholdContinuation` can refer to the continuation that is currently running. If that continuation crosses the compaction threshold, the scheduler treated its running action as an already-pending successor and did not queue another goal continuation. After compaction, the running action completed, leaving no queued work. `agent.continue()` could not resume the assistant-last stop, so the active goal stalled.

The fix only deduplicates undelivered continuation states: `queued`, `selected`, `preparing`, and `committing`. A `running` continuation may now enqueue its successor.

## Validation

- Exact failing regression: 20/20 consecutive passes
- Full `agent-session-compaction-continuation.test.ts`: 7/7 passed
- `npm run check`
- `git diff --check`
- Full CI workflow: [passed](https://github.com/PrimeIntellect-ai/prime-agent/actions/runs/32085810365)


jonaowen (changes requested):
This prompt tells the agent to end the turn while an unowned `asyncio.Task` runs only in the kernel. That is not a persistent/durable task contract: kernel restart, session rotation, process teardown, or task GC/cancellation loses the command and its result. `asyncio.create_task()` also leaves exceptions unobserved until a later turn that may never occur, and there is no cleanup/adoption registry.

Moreover the stacked base #1187’s `bash()` explicitly terminates background descendants and drains pipes as part of collection. Turning its coroutine into a detached task does not provide receipt paths, durable process identity, restart reconstruction, or cancellation ownership—the properties needed for an actual long job. Do not describe this as “persistent.” Either scope the guidance to short same-kernel work and require an immediate owner plus exception-safe callback/finally cleanup, or land a real managed task registry/detached-job substrate with durable receipts, restart/adoption semantics, bounded output, terminal observation, and cancellation tests. It should also remain stacked until #1187’s contract is accepted.

sethkarten (approved):

Issue #1290, opened 2026-08-12
refine: refiner receives entries truncated to 240 chars with no truncation marker and no way to request more


bilelrais:
*(This issue was rewritten. The original text claimed the harness store had an append/replace bug. That was wrong — replacement is deliberate and covered by tests: `refinement.test.ts` asserts `content).toBe("second")` after two sequential updates. The defect is in what the refiner is shown, not in how writes are applied. Earlier comments are superseded.)*

## Summary

`overviewForPrompt` renders each harness entry as `entry.content.replace(/\s+/g," ").slice(0, 240)` — no ellipsis, no length, no marker. That string is the entire `<current_harness_state>` block in the refiner's prompt. The refiner is then asked to emit complete replacement `content` for entries it may have seen 2% of, and it runs as a single `completeSimple` call with no `tools`, so it cannot request more. `REFINEMENT_SYSTEM_PROMPT` never states that `content` replaces.

Contrast: `formatHarnessStateForPrompt` (the system-prompt renderer) uses `compactText`, which does append `...` when it truncates. The refiner's view is the one without the signal.

## Measured over one session

89 refinements, 154 applied `update` edits, snapshots read from the session JSONL:

- **0 of 154** updates preserved any of the prior text
- **90 of 154** made the entry shorter
- **136 of 152** opened with the literal string `"APPEND — prior clauses stand"`
- one prompt note reached **v98**; 78 prior versions existed only in the session JSONL
- largest affected entry: a subagent spec of 14,572 chars, of which the refiner saw 240 (1.6%)

## Representative case

The refiner emitted, as the entry's new complete content:

> "(Existing sections of this note — pipeline, brief/prompt split, subagent lifecycle, context/compaction policy, git and backlog routing, reviewer adjudication — remain in force. The sections below are appended…)"

It named six sections it could not see and replaced them with a pointer to themselves. It knew they existed because they appeared in `<conversation>` and in past summaries in `<refinement_history>`; it could not reproduce them because they were never in its context.

## Suggested fixes, in order of value

1. **Mark truncation and show the real length** in `overviewForPrompt` — `compactText`, or `...(+N chars not shown)`. Without this the refiner cannot distinguish a complete short entry from a fragment of a long one.

[...]

PR #1317, opened 2026-08-13
fix(coding-agent): mark truncated entries in the refiner overview

bilelrais:
Refs #1290. **Draft — not ready for review.**

`overviewForPrompt` cut each harness entry to a fixed window with a bare `.slice()` — no ellipsis, no length — and that string is the entire `<current_harness_state>` block the refiner sees. It is then asked to emit complete replacement `content`, with no way to see the rest and nothing stating that an update replaces what is not reproduced.

Measured over one session: **154 applied `update` edits, 0 preserved any prior text, 90 made the entry shorter, and 136 opened with the literal string `"APPEND — prior clauses stand"`.** The largest affected entry was 14,572 chars, of which the refiner saw 240 (1.6%). One representative reply named six sections it could not see and replaced them with a pointer asserting they "remain in force".

## Changes

- `overviewForPrompt` marks truncation and reports the hidden length: `... (+N chars not shown)`, and returns the set of entries it rendered whole.
- `REFINEMENT_SYSTEM_PROMPT` gains a three-line `Editing model:` section: an `update` replaces the entry, what the marker means, and how to ask for more.
- **Expansion round-trip.** The refiner may reply `{"expand": ["id", ...]}`; those entries are injected in full and it may then emit edits or expand again. `completeSimple` returns an `AssistantMessage`, so the prior turn is pushed back verbatim — no `tools` array, no reconstructed turn, and any thinking blocks survive intact. Bounded at 3 expansion rounds (4 calls) and an 80,000-char budget; the final round says so, and withheld or unknown ids are reported rather than silently omitted.
- **Enforcement, not instruction.** `applyRefinementProposal` accepts a `fullyVisibleIds` set and refuses `update` for any id outside it, alongside the existing `"entry changed during refinement planning"` guard. The guarantee does not depend on model compliance.
- Entries short enough to render whole are fully visible by definition, so a harness of short entries never triggers a round-trip and behaves exactly as before.
- Display-prefixed ids (`local:foo`) resolve, since that is the form the overview renders and the form models actually emit.
- The two magic numbers in `overviewForPrompt` are named.

## Testing

`test/suite/regressions/1290-refiner-truncation-marker.test.ts` — 9 tests covering the marker, the prompt text, single and multi-entry expansion, expansion across rounds, the prefixed-id form, the round limit, budget exhaustion, refusal of an unseen entry, and a short entry updating with no round-trip. With `test/refinement.test.ts`: **68 passing.** `npm run check` clean.

Verified end to end against a real model (`claude-opus-5`) on a real 14,572-char entry: the refiner requested expansion, received the entry in full, found the lesson already present at char 10,801, and correctly declined to edit — three runs out of three. From 240 chars that judgement is not available.

## Open

- `REFINER_MAX_EXPANSION_ROUNDS = 3` and `REFINER_EXPANSION_BUDGET = 80_000` are my values with no evidence behind them; maintainers may prefer different ones.
- `overviewForPrompt` now returns `{ text, fullyVisible }` where `historyForPrompt` still returns a string. The set is load-bearing, but the asymmetry is worth a view.
- The control arm is unmeasured: I can show the new path works, not the size of the delta against the old one.

> [!NOTE]
> ### Mark truncated entries in the refiner overview and refuse updates to entries not seen in full
> - `overviewForPrompt` in [`refinement.ts`](https://github.com/PrimeIntellect-ai/prime-agent/pull/1317/files#diff-17b614d876315ad8ac0165562a8f07a5c4ada3b28928eb02d90f6b93a96ea91d) now annotates truncated entries with `... (+N chars not shown)` and returns a structured result tracking which entries were fully visible.
> - The refiner can request full entry content by canonical identity (`scope:kind:id`) across up to 3 expansion rounds before producing edits, bounded by an 80,000-char budget.
> - `applyRefinementProposal` rejects update edits for entries the refiner did not see in full, recording the edit as unapplied with an explicit error message.
> - Regression tests in [`1290-refiner-truncation-marker.test.ts`](https://github.com/PrimeIntellect-ai/prime-agent/pull/1317/files#diff-6b029a95ccaa7b291ca23b99f65582fa00fe38b98040678a741c0271d4bd54a3) and [`1317-refiner-visibility-identity.test.ts`](https://github.com/PrimeIntellect-ai/prime-agent/pull/1317/files#diff-5ba46ef379d51d9167ed62858a01bcbaaff6772ab61d70365709f0f1d635d759) cover truncation markers, expansion flow, budget limits, and cross-kind/cross-scope identity isolation.
> - Behavioral Change: refinement proposals that update large entries the model did not explicitly expand will now be silently refused rather than applied.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 1aa3f36.</sup>
> 
>

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## [Unreleased]
 
-- Fixed the continual harness refiner rewriting entries it was only shown a fragment of; it now sees how much is hidden, can request entries in full, and updates to entries it has not seen are refused ([#1317](https://github.com/PrimeIntellect-ai/prime-agent/pull/1317) by [@bilelrais](https://github.com/bilelrais)).
+- Fixed the continual harness refiner rewriting entries it was only shown a fragment of; it now sees how much is hidden, can request entries in full, and updates to entries it has not seen are refused; its editing rules also now state that the overview and the system prompt render only an entry's leading characters, so an entry's opening must say what it is and when to open it ([#1317](https://github.com/PrimeIntellect-ai/prime-agent/pull/1317) by [@bilelrais](https://github.com/bilelrais)).
 - Fixed URLs not opening on click in fullscreen mode on terminals such as Ghostty; clicking a link in the transcript, dock, or overlays now opens it in the browser.
 
 ## [0.7.2] - 2026-08-11
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [Unreleased]
 
+- Fixed the continual harness refiner rewriting entries it was only shown a fragment of; it now sees how much is hidden, can request entries in full, and updates to entries it has not seen are refused ([#1317](https://github.com/PrimeIntellect-ai/prime-agent/pull/1317) by [@bilelrais](https://github.com/bilelrais)).
 - Fixed URLs not opening on click in fullscreen mode on terminals such as Ghostty; clicking a link in the transcript, dock, or overlays now opens it in the browser.
 
 ## [0.7.2] - 2026-08-11
--- a/packages/coding-agent/src/core/agent-session.ts
+++ b/packages/coding-agent/src/core/agent-session.ts
@@ -7926,6 +7926,7 @@ export class AgentSession {
 				rollbackOf: plan.rollbackOf,
 				scope: targetScope,
 				baselineState: plan.baselineState,
+				fullyVisibleIds: plan.fullyVisibleIds,
 			});
 			result.harnessStatePath = saveHarnessState(targetHarnessStateDir, state);
 			if (targetScope === "global") {
--- a/packages/coding-agent/src/core/refinement/refinement.ts
+++ b/packages/coding-agent/src/core/refinement/refinement.ts
@@ -144,6 +144,7 @@ Continual harness components:
 
 Editing model:
 - An \`update\` replaces the entry's entire \`content\`; text you do not reproduce is removed.
+- The harness overview and every agent's system prompt render only the leading characters of an entry's \`content\` — a mechanical slice, not a summary — so the opening of \`content\` is the entry's interface: begin it with what the entry is and when to open it, never with a date, a header, or narrative.
 - \`... (+N chars not shown)\` marks an entry you were not shown in full; an update to one is refused.
 - Each overview entry has a canonical \`scope:kind:id\` identifier. To see entries in full, reply with \`{"expand": ["scope:kind:id", ...]}\` and nothing else. Copy each identifier exactly; you may expand again in a later reply.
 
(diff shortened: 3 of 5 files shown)


jonaowen (changes requested):
The visibility capability is keyed only by bare `id`, but IDs are not globally unique across harness kinds. `state.entries.prompt.card` and `state.entries.memory.card` are distinct valid entries. If the memory is short/expanded, `fullyVisible.add("card")` authorizes an update to the hidden prompt with the same ID. `findEntryById()` is likewise ambiguous: it walks kinds and returns the first matching ID, so `{"expand":["card"]}` may disclose/authorize a different kind from the one intended.

This defeats the load-bearing enforcement. Key authority by the exact `(kind,id)` identity (and scope if merged state can contain both), render/request an unambiguous canonical identifier, and check that exact key in `applyRefinementProposal`. Add a regression with same ID in two kinds where one is fully visible and one truncated, proving neither expansion nor update authority crosses entries. Also preserve the exact baseline identity through the expansion rounds; display-prefix stripping alone is not sufficient.

bilelrais:
Thanks — the collision is real and the fix is in `9025c090b`. Chasing it also turned up a worse defect in the same enforcement, which is described below.

## Identity

Visibility authority is now keyed by the canonical `scope:kind:id` identity (`local:prompt:card`) rather than a bare id, so a short `memory:card` no longer authorizes a truncated `prompt:card`.

Scope is required, not defensive: merged state can hold a global and a local entry with the same kind and id, so the local record is re-keyed for identity while retaining its original internal id. `findEntryById()`'s ambiguous kind walk is gone. Expansion resolves against an immutable baseline map captured from the initial overview, so identity is carried through the rounds rather than reconstructed by stripping a display prefix. `applyRefinementProposal` checks the target's actual scope, kind, and internal id.

The overview and the expansion contract both render the canonical identifier, and the system prompt instructs the model to copy it exactly.

## The enforcement was inert in production

`AgentSession` never passed `fullyVisibleIds` into `applyRefinementProposal`, so the refusal this PR exists to provide never ran outside tests. Every update was applied regardless of whether the refiner had seen the entry. One line, now fixed.

The existing tests could not catch it because they called `applyRefinementProposal(...)` directly and supplied `fullyVisibleIds` themselves — they proved the function honours the option, not that anything provides it. The new regressions drive `session.refine(...)` end to end; removing that line again fails both.

## Regressions

`test/suite/regressions/1317-refiner-visibility-identity.test.ts`

- same id in two kinds, one fully visible and one truncated: expansion discloses only the requested entry, the update to the truncated entry is refused, and its content is unchanged
- same kind and id in both scopes: same guarantees across the scope boundary

## Validation

`npm run check` clean. With the repo's launcher from `packages/coding-agent`: `test/refinement.test.ts` 59 passed, `1290-refiner-truncation-marker.test.ts` 9 passed, `1317-refiner-visibility-identity.test.ts` 2 passed.

Running all of `test/suite/` shows 17 failures in five daemon, supervisor and worker-recovery files. They reproduce identically on the base commit on this machine and are timing-sensitive against live daemon processes; none touches refinement.

Not addressed here: `REFINER_MAX_EXPANSION_ROUNDS` and `REFINER_EXPANSION_BUDGET` remain arbitrary values, and discovery-empty is still reported as an auth failure at the call site.

kevinjosethomas:

PR #1323, opened 2026-08-13
fix(coding-agent): list client-owned workers

### LINKED ISSUE #1215: `list` reports "No active agents" while agents are running: isVisibleWorker hides any worker with an ownerClientId
`prime-agent list` printed `No active agents.` while five prime-agent processes and an ipykernel were demonstrably running (confirmed with `ps`), and the agent was actively writing to its session JSONL.

### Root cause

`packages/coding-agent/src/modes/daemon/daemon-supervisor.ts`:

```typescript
private isVisibleWorker(worker: ResidentWorker): boolean {
    return worker.descriptor.ownerClientId === undefined;
}
```

A worker is visible only if it has **no owning client**. A run started from the CLI has an `ownerClientId` set, so it is filtered out of the `list` response entirely (`case "list"` → `cli/daemon-command.ts` `runList` → prints `No active agents.` when the filtered list is empty).

### Why it matters

The operator-facing answer to "what is running?" is wrong, and wrong in the dangerous direction — it under-reports. Practical consequences we hit:

- A hung agent that the tooling insists does not exist, which turned a quick diagnosis into a long one.
- The documented cleanup discipline (stop/shutdown detached agents between runs) is hard to follow, because you cannot stop what you cannot see. A leftover daemon then interferes with the next run.

### Suggested direction

Please **don't** simply widen `isVisibleWorker` — the same predicate also gates attach/first-worker selection in at least three other call sites in the same file, and loosening it there would change attach semantics.

Fixing on the **list path only** seems right:

- include owned workers in `list`, with an explicit `owner`/`attached` column so they remain distinguishable from unowned background agents, and
- never print "No active agents" while any worker exists — if the only workers are owned, say so rather than reporting nothing.

Happy to send a PR if that shape works for you.

veerareddyvishal144:
## Summary

Fix `prime-agent list` under-reporting live workers that are owned by another CLI client.

The supervisor previously filtered the list response through `isVisibleWorker()`, which intentionally excludes client-owned workers for attach/selection semantics. Because `prime-agent list` uses a fresh daemon connection, it could not see those workers and incorrectly printed `No active agents.` even while they were running.

This change keeps `isVisibleWorker()` unchanged and fixes only the list path:

- include every resident worker in the daemon list response, regardless of ownership
- expose worker ownership on the public session summary
- add an `owner` column to the CLI table (`attached` for client-owned workers)
- add coverage for owned and unowned list rows

## Validation

- `npx vitest run packages/coding-agent/test/daemon-list-format.test.ts`
- `npx tsgo --noEmit`
- `npx biome check --write --error-on-warnings .`
- repository pre-commit checks passed
- `git diff --check`

Fixes #1215

> [!NOTE]
> ### Fix daemon list command to include client-owned worker sessions
> - `DaemonSupervisor.handleList` now returns all workers regardless of ownership, removing the previous filter that excluded client-owned sessions from other clients.
> - `publicSummary` is extended to include `ownerClientId` on each session summary when set on the worker descriptor.
> - The session list table gains an `owner` column that displays `attached` for client-owned sessions and is empty for resident sessions.
> - Risk: clients that previously relied on the list being scoped to visible/accessible workers will now see all sessions.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized f97d00a.</sup>
> 
>

### DIFF
--- a/packages/coding-agent/src/cli/daemon-command.ts
+++ b/packages/coding-agent/src/cli/daemon-command.ts
@@ -1644,6 +1644,7 @@ function isSessionSummary(value: unknown): value is SessionSummary {
 		typeof candidate.attachedClients === "number" &&
 		typeof candidate.messageCount === "number" &&
 		(candidate.unfinishedActionCount === undefined || typeof candidate.unfinishedActionCount === "number") &&
+		(candidate.ownerClientId === undefined || typeof candidate.ownerClientId === "string") &&
 		typeof candidate.sessionActions === "object" &&
 		candidate.sessionActions !== null &&
 		typeof candidate.sessionActions.queuedCount === "number" &&
--- a/packages/coding-agent/src/cli/daemon-list-format.ts
+++ b/packages/coding-agent/src/cli/daemon-list-format.ts
@@ -26,6 +26,7 @@ type ListRow = {
 	model: string;
 	messages: string;
 	clients: string;
+	owner: string;
 };
 
 export function formatSessionListTable(sessions: readonly SessionSummary[], nowMs = Date.now()): string {
@@ -37,8 +38,9 @@ export function formatSessionListTable(sessions: readonly SessionSummary[], nowM
 		model: formatSessionModel(session.model),
 		messages: String(session.messageCount),
 		clients: String(session.attachedClients),
+		owner: session.ownerClientId ? "attached" : "",
 	}));
-	return formatTable(["name", "id", "status", "age", "model", "messages", "clients"], rows, formatListCell);
+	return formatTable(["name", "id", "status", "age", "model", "messages", "clients", "owner"], rows, formatListCell);
 }
 
 function sortSessionsForList(sessions: readonly SessionSummary[]): SessionSummary[] {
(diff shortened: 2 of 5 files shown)


jonaowen (changes requested):
The list-path direction is right, but this needs two changes before merge.

1. Do not add raw `ownerClientId` to the public `SessionSummary` returned to every daemon client. The UI only needs an ownership classification (`attached`/resident), not the internal client capability/identity. Expose a boolean or narrow enum computed by the supervisor. This is also a daemon response-shape change: update the protocol/schema compatibility contract and old/new peer tests per repository doctrine, or place it behind an existing negotiated shape that is demonstrably compatible.
2. The added test constructs summaries and exercises only table formatting. It does not prove `handleList` includes a worker owned by a different client while attach/selection remains filtered, nor that `busyClientOwnedSessionCount` and stop safety remain accurate. Add a supervisor/daemon behavioral regression with owned and unowned workers and two client identities.

Please also remove or redefine the now-dead `includeClientOwned` request semantics rather than silently accepting and ignoring it; retaining a flag with behavior different from its negotiated contract is misleading.

kevinjosethomas:

PR #1325, opened 2026-08-13
test(coding-agent): make the harness-state atomicity test pass on Windows

Hotragn:
`test/refinement.test.ts > atomically replaces harness state without leaving temporary files` fails on Windows for two POSIX assumptions. Neither is a product bug — the test is wrong, not the code.

**1. Path separator**

```ts
expect(readdirSync(harnessStateDir)).toEqual([statePath.split("/").at(-1)]);
```

`saveHarnessState` builds `statePath` with `path.join`, so on Windows it contains `\`. Splitting on `/` returns the whole path rather than the filename:

```
- Expected  [ "C:\Users\...\prime-agent-refinement-test-NbADsg\harness_state.json" ]
+ Received  [ "harness_state.json" ]
```

`basename` is separator-correct on every platform.

**2. File mode**

```ts
chmodSync(statePath, 0o600);
saveHarnessState(harnessStateDir, state);
expect(statSync(statePath).mode & 0o777).toBe(0o600);
```

Windows `chmod` only toggles the read-only bit; Node reports `0o666` (438), so this asserts something the platform cannot express:

```
AssertionError: expected 438 to be 384
```

The behaviour under test — `saveHarnessState` reading the existing mode and preserving it across the atomic rename — is genuinely POSIX-only, so the mode assertions now run on POSIX only. The save and re-save still execute on every platform, so the atomicity half of the test is unchanged everywhere. Coverage on Linux and macOS is identical to before.

## Scope

This does not fix a red build: CI is ubuntu-only today, where the test already passed. It removes a false failure for anyone running the suite on Windows, and it is a prerequisite for the Windows matrix leg proposed in #781. There are ~20 open Windows issues, so contributors landing there currently have to distinguish this failure from their own.

Verified: `packages/coding-agent` `test/refinement.test.ts` **59/59 on Windows** (was 58/59 with this one failing), and `npm run check` clean.

> [!NOTE]
> ### Fix harness-state atomicity test to pass on Windows
> Updates the atomicity test in [refinement.test.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/1325/files#diff-00863e8a97f5558182ff8ff3d737756693af93b3c0467130f90b4739f4298195) to use `basename(statePath)` instead of a Unix-style string split, and skips `chmod`/`stat` permission assertions on Windows since the platform does not support POSIX file modes.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized b74f8c4.</sup>
> 
>

### DIFF
--- a/packages/coding-agent/test/refinement.test.ts
+++ b/packages/coding-agent/test/refinement.test.ts
@@ -1,6 +1,6 @@
 import { appendFileSync, chmodSync, mkdtempSync, readdirSync, rmSync, statSync, writeFileSync } from "node:fs";
 import { tmpdir } from "node:os";
-import { join } from "node:path";
+import { basename, join } from "node:path";
 import type { AgentMessage } from "@earendil-works/pi-agent-core";
 import type * as PiAi from "@earendil-works/pi-ai";
 import type { AssistantMessage, Model } from "@earendil-works/pi-ai";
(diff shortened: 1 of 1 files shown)


jonaowen (approved):
`basename` is the correct cross-platform filename projection. The mode-preservation assertion is specifically about POSIX permission bits, which Windows cannot represent through Node `chmod`; retaining both save operations on Windows while gating only that assertion preserves the platform-independent atomic replacement coverage.

kevinjosethomas:

PR #1327, opened 2026-08-13
Fix Windows supervisor replacement launch lock

### LINKED ISSUE #1286: Windows supervisor self-replacement always fails: lock directory is created under the named-pipe namespace
### Prime Agent version

0.7.2 (the same code path was present in 0.7.1)

### Platform

Windows 11, native Windows daemon/supervisor using the named-pipe socket.

### Problem

`launchReplacementSupervisor` derives its filesystem lock directory from `dirname(supervisorSocketPath)`. On Windows the supervisor socket is a named pipe such as `\\.\pipe\prime-agent-daemon`, so `dirname(...)` resolves to the pipe namespace rather than a real filesystem directory. The subsequent `mkdirSync(candidateDirectory, ...)` fails with `EPERM` on every replacement attempt.

The effective path is shaped like:

```text
\\.\pipe\.supervisor-launch-<hash>.lock.candidate-<pid>-<token>
```

Representative error (identifiers removed):

```text
failed to launch replacement supervisor: Error: EPERM: operation not permitted,
mkdir '\\.\pipe\.supervisor-launch-<hash>.lock.candidate-<pid>-<token>'
```

### Impact

A worker can remain alive while its supervisor is unavailable, but the supervisor cannot self-replace. On one installation the agent log contains 675 `failed to launch replacement supervisor` records with this exact pipe-namespace `EPERM` over six days. External recovery has been the only successful exit.

This is especially risky because `prime-agent list` talks to the supervisor: supervisor-unavailable and worker-absent look identical through that channel, although they require different recovery decisions.

### Suggested fix

Derive the launch lock from a real filesystem state/registry directory (for example the directory selected by `PRIME_AGENT_INTERNAL_DAEMON_SUPERVISOR_REGISTRY_DIR`), never from the socket path. The named-pipe address should remain only the IPC endpoint.

Please also add a Windows test that asserts the replacement lock path is a normal filesystem path and that a dead supervisor can be replaced while an existing worker/session descriptor remains present.

### Reproduction

1. Run Prime Agent 0.7.2 natively on Windows.
2. Start a daemon worker/session.
3. Make the supervisor unavailable while leaving the worker/session state present.
4. Trigger the replacement-supervisor path.
5. Observe three `mkdir` failures under `\\.\pipe` and no replacement supervisor.

jonaowen:
## Summary

- place the replacement-supervisor launch lock in the existing real filesystem supervisor registry instead of the daemon endpoint directory
- handle Windows directory-rename collision semantics without treating a live contender as a launch failure
- add a cross-platform regression using a literal Windows named-pipe endpoint and the real replacement method

## Validation

- focused regression: 1 passed
- `npm run check`: green (Biome, tsgo, installer render, browser smoke)
- fresh adversarial review: GO after one correction moved registry creation under the existing reset guard

The patch does not change daemon wire/schema/protocol behavior. It preserves the existing socket hash and stale-lock algorithm; PID-reuse/token/ABA hardening is separate scope.

Fixes #1286.

> [!NOTE]
> ### Fix Windows supervisor replacement launch lock in `AgentDaemon`
> - Changes the lock directory base in `launchReplacementSupervisor` from `dirname(supervisorSocketPath)` to a global registry directory from `defaultDaemonSupervisorRegistryDir()`, fixing failures when `supervisorSocketPath` is a Windows named-pipe path with no valid filesystem directory.
> - Creates the registry directory with mode `0o700` before use and treats `EPERM` on rename (when the destination already exists) as equivalent to a pre-existing lock rather than throwing.
> - Exports `defaultDaemonSupervisorRegistryDir` from [daemon-supervisor-ownership.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/1327/files#diff-6f9ef6ce6a9d2bc51c61fe94d59310efffd26d8f7b3966f5355606b2847d06a5) so it can be imported by the daemon mode module.
> - Adds a regression test in [1286-windows-supervisor-lock.test.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/1327/files#diff-14278dca027dc3aa482182f7f40b9d772d616f8ad11275cccc17997074bdb9ad) covering the named-pipe path scenario.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized a3ddbe9.</sup>
> 
>

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -6,6 +6,7 @@
 - Fixed `openai-codex` models being invisible to `rlm` subagents and `find_models` because model discovery reported Prime Agent's own version as the Codex client version ([#1375](https://github.com/PrimeIntellect-ai/prime-agent/pull/1375) by [@bilelrais](https://github.com/bilelrais)).
 - Added a working hint that recommends sharing traces with Prime Intellect to help train open-source LLMs.
 - Restored bare `prime-agent --resume` opening the agents view and the `/resume [id|path]` slash command; bare commands open the agents view and an argument resumes that session in place.
+- Fixed Windows daemon workers failing to coordinate replacement supervisor launches through named-pipe paths ([#1286](https://github.com/PrimeIntellect-ai/prime-agent/issues/1286)).
 - Fixed URLs not opening on click in fullscreen mode on terminals such as Ghostty; clicking a link in the transcript, dock, or overlays now opens it in the browser.
 - Fixed ctrl+p ("Toggle agent message expansion") only toggling received agent messages; it now expands and collapses sent agent messages together with received ones.
 
--- a/packages/coding-agent/src/modes/daemon/daemon-mode.ts
+++ b/packages/coding-agent/src/modes/daemon/daemon-mode.ts
@@ -187,7 +187,11 @@ import {
 	prepareDaemonSocketPath,
 	restrictDaemonSocketPath,
 } from "./daemon-socket.js";
-import { assertDaemonSupervisorOwnerCurrent, isDaemonShutdownAdmissionActive } from "./daemon-supervisor-ownership.js";
+import {
+	assertDaemonSupervisorOwnerCurrent,
+	defaultDaemonSupervisorRegistryDir,
+	isDaemonShutdownAdmissionActive,
+} from "./daemon-supervisor-ownership.js";
 import {
 	DAEMON_WORKER_ACTIVE_SESSION_ID_ENV,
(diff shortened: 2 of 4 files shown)


jonaowen:
Rebased cleanly onto current main `5e268e28f`. Post-rebase focused #1286 regression: 1/1 passed. Post-rebase `npm run check`: green; 910 files checked, no fixes, tsgo/installer/browser smoke passed.

jonaowen:
Rebased onto current main `7787f0741`. Post-rebase focused #1286 regression: 1/1 passed. `npm run check`: green; 910 files checked, no fixes, tsgo/installer/browser smoke passed.

kevinjosethomas:

PR #1341, opened 2026-08-13
fix(coding-agent): isolate invalid provider registrations

### LINKED ISSUE #1230: prime-agent: one provider with missing apiKey blocks ALL providers at extension load
- Репозиторий: PrimeIntellect-ai/prime-agent
- Тип: bug (UX / error isolation)
- Статус: готов к отправке (2026-08-11)
- Воспроизведено на: prime-agent 0.7.1 (release tarball), mode json

## Title

A single extension that registers a provider without a resolvable apiKey prevents every other provider from loading ("Unknown provider" for unrelated providers)

## Body

**Context.** We run prime-agent as a multica runtime with two provider extensions in `~/.prime/agent/extensions/`: `deepseek-provider.ts` (apiKey from `process.env.DEEPSEEK_API_KEY`) and `codex-provider.ts` (apiKey from `process.env.CODEX_API_KEY`).

**Bug.** At process start, *both* extensions are evaluated. If only `CODEX_API_KEY` is set and `DEEPSEEK_API_KEY` is missing, startup fails for **both** providers:

```
Error: Extension ".../deepseek-provider.ts" error: Provider deepseek: "apiKey" or "oauth" is required when defining models.
Error: Unknown provider "codex". Use "prime-agent model list" to see available providers/models.
```

Even though the codex provider has a valid key and the user only wants to use codex, the *unused* deepseek extension's missing key poisons the whole startup — the codex provider never registers.

**Expected.** A provider whose apiKey can't be resolved at extension-load time should either (a) register lazily and error only when actually invoked, or (b) log a warning and skip just that provider, or (c) at minimum not prevent other providers from registering. One misconfigured/unused provider should not take down the entire agent.

**Notes.**
- The extension pattern `apiKey: process.env.X ?? ""` is the documented way to source keys (`docs/custom-provider.md`), so this is a common configuration shape.
- Same applies to any provider registered with `models` where the env var is absent at load time.
- Workaround for us: always export all keys (even unused ones), or keep each provider in its own extension and delete unused files — but both are fragile in multi-tenant/daemon setups where keys are injected per-session.

mrrobotke:
## Summary

- treat failed extension provider registrations as non-fatal warnings
- keep valid providers available when another extension has no resolvable credential
- add focused regression coverage for mixed unavailable and valid providers

## Behavior

An invalid provider is skipped with a diagnostic that identifies its extension and validation error. Other provider registrations continue loading, while selecting the skipped provider still follows the existing unavailable-provider path.

## Validation

- `npx tsx ../../node_modules/vitest/dist/cli.js --run test/agent-session-services.test.ts test/suite/regressions/1230-provider-registration-isolation.test.ts`
- `npm run check`

Fixes #1230

> [!NOTE]
> ### Isolate failed provider registrations in coding agent so valid providers still load
> Previously, a single extension with an unavailable provider credential could block other providers from loading. Now, failed provider registrations in `createAgentSessionServices` are caught individually and reported as warnings ("Skipped provider from extension...") instead of errors, leaving valid providers unaffected. A regression test in [1230-provider-registration-isolation.test.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/1341/files#diff-82ac8acca2b6663fbad6d1fd2200383af9b13c1ddfc824ec59a55f19b1a4ae2d) verifies that a missing-credential provider is skipped while a valid provider registers successfully.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 52add7f.</sup>
> 
>

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [Unreleased]
 
+- Fixed an extension with an unavailable provider credential preventing other providers from loading ([#1230](https://github.com/PrimeIntellect-ai/prime-agent/issues/1230)).
 - Fixed URLs not opening on click in fullscreen mode on terminals such as Ghostty; clicking a link in the transcript, dock, or overlays now opens it in the browser.
 
 ## [0.7.2] - 2026-08-11
--- a/packages/coding-agent/src/core/agent-session-services.ts
+++ b/packages/coding-agent/src/core/agent-session-services.ts
@@ -234,8 +234,8 @@ export async function createAgentSessionServices(
 		} catch (error) {
 			const message = error instanceof Error ? error.message : String(error);
 			diagnostics.push({
-				type: "error",
-				message: `Extension "${extensionPath}" error: ${message}`,
+				type: "warning",
+				message: `Skipped provider from extension "${extensionPath}": ${message}`,
 			});
 		}
 	}
--- a/packages/coding-agent/test/suite/regressions/1230-provider-registration-isolation.test.ts
+++ b/packages/coding-agent/test/suite/regressions/1230-provider-registration-isolation.test.ts
@@ -1,0 +1,82 @@
+import { existsSync, mkdirSync, rmSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+import { registerFauxProvider } from "@earendil-works/pi-ai";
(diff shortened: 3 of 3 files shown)


jonaowen (approved):
The registration loop already isolates each extension provider structurally; changing a rejected provider registration from a process-fatal runtime diagnostic to a warning completes that isolation. The regression proves the invalid model is absent, a later valid provider remains registered, and no fatal diagnostic survives. Other extension load errors remain unchanged.

kevinjosethomas:

PR #1342, opened 2026-08-13
fix(coding-agent): make daemon snapshot ids unique with a per-session serial

### LINKED ISSUE #1229: v0.7.1: snapshot position is reused as transfer identity; byte mismatch disconnects a healthy worker
## Human Notes

Below was researched and generated by Sol/high + several Terra/xhigh and Luna/xhigh subagents. The corruption issues are getting very annoying to me, so I had Sol come up with a new design that's dramatically cleaner and should permanently fix all related issues.

I'm creating an issue instead of a PR because I've found for my own repos that good AI-written issues are easier to work with than PRs. I'm happy to send a PR if you'd prefer.

## Summary

Two ordinary chunked snapshots can use the same `(activeSessionId, eventGeneration, lastEventSequence)` ID but contain different bytes. The supervisor treats that as a transport/protocol violation, closes the supervisor↔worker control socket, and starts worker recovery.

This turns a recoverable UI snapshot race into an interruption of active work. The durable error is:

```text
Duplicate snapshot <activeSessionId>-<eventGeneration>-<lastEventSequence> did not match cached bytes
```

This has occurred repeatedly without a machine sleep or a known network transition. It is in the local Unix-socket snapshot path, not an HTTPS provider path.

## Environment

- Prime Agent: `0.7.1` (`95afd319`, global install)
- OS: macOS / Darwin, Apple Silicon
- Node: `v23.7.0`
- Workload: several long-lived root sessions plus recursive children; some transcripts are large

## Local evidence

The structured supervisor logs contain **31** occurrences between August 5 and August 10 across five active-session IDs:

| active session | count | public daemon client count |
|---|---:|---:|
| `59dface50d55` | 12 | 1 |
| `984502b61a86` | 12 | 1 |
| `fffe5d3a9b34` | 4 | 1 |
| `c423d0db8fc3` | 2 | 1 |
| `395d6526ccac` | 1 | 1 |

Each affected session used one public client for all of its recorded failures. Multiple windows are therefore not required; one backpressured client repeatedly entering catch-up is enough.

The mismatch is followed by failures such as:

```text
Daemon worker client closed
Session worker is recovering
Session worker is not connected
```

For example, the first mismatch for `59dface50d55` was immediately followed by `get_connection_state`, `get_messages`, and `get_session_context` failing because the worker was recovering. There was no earlier recovery event for that worker in the available structured log.

A worker launched under `0.7.1` later recovered to `lifecycle: "ready"`, but its descriptor still preserves another recurrence:

```json
{
  "lifecycle": "ready",
  "consecutiveFailures": 0,
  "lastError": "Duplicate snapshot 89f45adf1a01-c814fbea45ac-26654 did not match cached bytes"
}
```

One larger incident crossed this boundary during active work. Recovery reported that model/tool/child operations were uncertain and were not replayed. Later turns could not resume because provider history contained an orphaned function-call output. That incident also hit `ENOSPC`, which likely made recovery worse, but the duplicate-snapshot failures began before that disk-full event and also occur in other sessions.

## Snapshot identity invariant and possible causes

What the exception proves is narrow: two byte-different frames were associated with one `(worker, activeSessionId, snapshotId)`. It does not by itself prove which producer-side race made the bytes differ.

In `v0.7.1`, snapshot IDs use only the active session ID and event cursor:

```ts
const snapshotId = `${activeSessionId}-${state.eventGeneration}-${state.lastEventSequence}`;
```

Credible causes include:

1. **Capture/serialization race.** `createSessionSnapshot()` returns `messages: session.messages`, which is live state. `createSnapshotTranscriptChunks()` copies only the array:

   ```ts
   const messages = [...options.messages];
   ```

   It serializes the still-referenced message objects later, when the iterable is consumed. `createSessionSnapshot()` also awaits child-snapshot discovery but does not fence the whole payload to one transcript/event revision.
2. **Overlapping or replayed transfers.** Duplicate validation has one `generation.duplicateChunkIndex`. Late or overlapping frames such as `A:chunk0, B:chunk0` can make a valid replay look inconsistent unless each transfer is separately correlated. This is the failure family that #397, #442, and #580 intended to close, so the local evidence may be a remaining edge or regression.
3. **A stale frame from a prior worker connection.** The frame handler should be correlated to the originating worker-connection generation.
4. **Cache corruption/bookkeeping.** This was plausible during the one `ENOSPC` incident, but does not explain the separate `0.7.1` recurrence with successful disk writes.

The small demonstration below proves only that the shipped chunk source can produce different bytes for the same caller-supplied ID; it does not identify which of these races occurred in production. Content hashes, transfer IDs, and chunk-index logging would distinguish them.

## Fatal handling

When the supervisor receives a second transfer with the same ID/metadata, it requires every chunk to be byte-identical. A difference calls:

```ts
failWorkerSnapshotCache(worker, activeSessionId, error, true, snapshotId)
```

The `true` argument closes the worker control channel. That is the direct transition from a display-snapshot mismatch to worker recovery.

The same sequence-only identity and fatal mismatch behavior remain on current `main` (`14d6e749`) as of August 11.

## Small source-level demonstration

This does not reproduce the complete daemon timing, but it demonstrates the unsafe identity/serialization invariant in the shipped chunker:

```js
import { createSnapshotTranscriptChunks } from
  "/opt/homebrew/lib/node_modules/prime-agent/dist/modes/daemon/snapshot-transcript-cache.js";

const message = {
  role: "assistant",
  content: [{ type: "text", text: "before" }],
  timestamp: 1,
};
const options = {
  activeSessionId: "a",
  snapshotId: "a-generation-1",
  messages: [message],
};

const first = createSnapshotTranscriptChunks(options);
const second = createSnapshotTranscriptChunks(options);
const firstBytes = Buffer.concat([...first]);
message.content[0].text = "after";
const secondBytes = Buffer.concat([...second]);

console.log(firstBytes.equals(secondBytes)); // false; same snapshotId
```

## Suggested regression test

1. Pause `_processAgentEvent` after live message state changes but before `broadcastToSession()` advances `lastEventSequence`.
2. Request two attach/catch-up snapshots for the same session/cursor, with one transcript difference.
3. Release the event.
4. Assert that the worker control channel remains connected and the client receives a fresh replacement/resync snapshot.

A second test should use one backpressured client, matching the local logs, rather than requiring two clients.

## Proposed fix pattern: make one transfer an immutable value

The smallest design with a strong guarantee is not a daemon rewrite. It is to stop using a logical session position as both byte identity and transfer identity.

The protocol should enforce this invariant:

> Once a snapshot transfer ID is published, it maps to exactly one frozen byte stream for its entire lifetime.

### Minimal implementation

1. **Make the existing `snapshotId` an opaque, unique transfer ID.** Allocate a new nonce/counter/UUID for every attach, replacement, and catch-up materialization—even when the event cursor is unchanged. Keep `(eventGeneration, lastEventSequence)` separately as the snapshot's base position.
2. **Freeze bytes before publishing the ID.** Eagerly serialize the messages into immutable chunks before `session_snapshot_begin` can be sent or socket backpressure can yield. Record the exact payload length and a digest over the defined frozen representation. Retries of that exact transfer read those same chunks; they never serialize live message objects again. The digest verifies the invariant; it does not replace freezing.
3. **Capture a coherent committed cut.** Under the per-session serialization boundary choose a persistent session epoch/high-watermark `R` and define the source as exactly the state committed through `R`. Snapshot construction currently awaits child discovery, so validate the client-visible revision before and after construction; if it changed, discard and retry. Freeze the view inside the short boundary, then encode/stream outside it. Events after `R` replay in order from `R + 1`; a gap triggers resync. A random transfer ID is not a state revision.
4. **Contain failure to the transfer.** A same-ID metadata/chunk mismatch should invalidate that transfer and request a fresh snapshot. It must not close/recover the root worker. Authentication, frame decoding, or routing corruption can still fail closed at the worker-channel level.
5. **Fence stale worker connections.** Include a worker-incarnation or connection-generation token in snapshot frame handling so buffered frames from a replaced connection cannot mutate the current transfer.
6. **Bound the attempt.** Cap bytes, chunk count, concurrent transfers, spill storage, and transfer lifetime per session/client and globally. A slow client loses its attachment/transfer, not the worker; every ACK/abort/disconnect path releases pins, files, timers, and credits exactly once.

Bind each attempt to `{sessionId, sessionEpoch, R, transferId, totalLength, payloadDigest}`. Cache/validation state must never be keyed by only `R` or only the transfer ID.

This makes the relevant state machine simple:

```text
TransferId -> Building -> Frozen -> Streaming -> Verified -> Installed | Aborted
```

A `TransferId` can never return to `Building`, switch payloads, or share chunk progress with another attempt. Duplicate chunks for one transfer are idempotent only when their index and bytes match. A conflict aborts that attempt; a new attempt always has a new ID.

### Guarantees

1. Same transfer ID implies the same frozen chunks.
2. Two attempts never share `duplicateChunkIndex` or other mutable validation state.
3. A stale frame cannot affect a newer worker incarnation/connection.
4. A completely verified snapshot at position `R` installs atomically and cannot replace a newer applied `(sessionEpoch, R)`; then client events replay from `R + 1`, and a gap requests resync.
5. A published `R` refers to a committed/reconstructable cut, not process-local state that disappears on restart.
6. A bad/slow client or transfer cannot restart the root worker or disturb sibling sessions.

These should be assertions in the protocol types/state machine, not conventions spread across handlers.

### If snapshot reuse is needed later

Only then introduce a second identity:

```text
SessionPosition = (sessionEpoch, eventSequence)   // logical state location
SnapshotRef     = (formatVersion, size, digest)   // immutable artifact
TransferId      = random ID                       // one delivery attempt
```

The worker would freeze and hash chunks, publish an immutable `SnapshotRef`, then create independent `TransferId`s for each receiver/retry. A manifest commits to ordered chunk hashes, schema/codec version, base position, count, and total length. This is stronger and enables safe reuse, but it is not necessary for the first fix because current clients already treat `snapshotId` as an opaque string.

### Backpressure and multiple roots

Track bulk snapshot backpressure per `(client, session)`, with bounded bytes and one coalesced pending catch-up to the newest position. Reserve control-plane capacity for ACK/status/abort traffic. A slow session must not set one global flag that delays every session attached through the same client.

Supervisor startup/adoption and update preparation should also report failures per root instead of awaiting one unhealthy worker in global all-or-nothing loops. That is a related failure-containment improvement, not required to fix the snapshot identity bug.

### What not to do

- **Do not merely suppress the comparison.** Equal IDs with unequal bytes indicate a broken identity contract.
- **Do not adopt a blanket “never close workers” rule.** True authentication/framing/routing corruption should still quarantine the affected worker connection.
- **Do not move all session authority into the shared supervisor.** One worker per root is already the right isolation boundary; centralizing durable session mutation would enlarge the blast radius.
- **Do not begin with a full event-sourcing/MVCC rewrite.** A durable WAL, content-addressed segment store, and transactional outbox may improve crash recovery later, but are unnecessary to remove this race.
- **Do not deep-copy an unbounded transcript for every client indefinitely.** Start with bounded eager chunks and the existing spill mechanism; if cost becomes material, evolve to immutable append-only transcript segments so snapshots reuse unchanged prefixes.

### Tests that make regression difficult

1. Create two snapshots at the same event cursor; assert distinct transfer IDs.
2. Mutate a source message after transfer creation; emitted chunks must remain byte-identical to the frozen transfer.
3. Interleave frames `A:0, B:0, A:1, B:1`; validation state must remain independent.
4. Deliver frames from an old worker connection after recovery; they must be inert.
5. Force a same-transfer conflict; the transfer fails but the worker, root, and sibling sessions remain live.
6. Change state during async snapshot construction; revision validation retries instead of publishing a mixed view.
7. Backpressure one large session while another root handles attach/status promptly.
8. Inject `ENOSPC` during snapshot spill; return a typed snapshot/storage failure without killing the worker or claiming a recoverable checkpoint.

A randomized property test should duplicate, reorder, delay, corrupt, and replay frames while checking the guarantees above. Compatibility tests should cover old/new supervisor-worker combinations; the unique opaque `snapshotId` is wire-compatible today, while new digest/incarnation fields should be capability-gated.

### Scope estimate

- Unique opaque transfer IDs, eager frozen chunks, scoped mismatch handling, and focused regressions: roughly **3–5 agent-hours**.
- Revision validation plus worker-incarnation fencing and a real-process stress test: roughly **2–4 additional agent-hours**.
- Content-addressed reusable snapshot artifacts and durable retention/GC: a separate, larger follow-up rather than a prerequisite.

## Multiple sessions

Multiple top-level sessions do share one supervisor event loop and a serialized all-worker peer-sync queue. More sessions, large snapshots, and child-update bursts can increase CPU, backpressure, and catch-up frequency, so they are likely amplifiers.

I found no evidence that session A's transcript bytes are directly compared with session B's: current cache maps are per worker, paths include the active-session-prefixed snapshot ID, and the failures group cleanly by one active session and one client. This is not the older shared spill-cache deletion bug.

## Related work

- [#397 / ENG-4602](https://github.com/PrimeIntellect-ai/prime-agent/pull/397): merged snapshot idempotency/non-fatal containment work. The observed initial mismatch still closes a healthy worker, so this appears to be incomplete coverage or a regression.
- [#407](https://github.com/PrimeIntellect-ai/prime-agent/pull/407): merged removal of the shared cross-worker spill cache.
- [#442](https://github.com/PrimeIntellect-ai/prime-agent/pull/442) and [#580](https://github.com/PrimeIntellect-ai/prime-agent/pull/580): merged replacement/catch-up serialization fixes.
- [#966](https://github.com/PrimeIntellect-ai/prime-agent/pull/966): open PR for the known case where recovery drops uncertain operations but leaves stale bytes under the same counter-derived ID. The first local mismatch above occurred before any recorded recovery, so recovery-time invalidation alone does not cover the initiating case.
- [#1200](https://github.com/PrimeIntellect-ai/prime-agent/issues/1200) and [#1222](https://github.com/PrimeIntellect-ai/prime-agent/issues/1222): shared-control-plane starvation and slow-hello misclassification. Those can amplify the symptoms but do not report this exact mismatch-to-worker-close transition.

Stable source references for `v0.7.1`:

- deterministic snapshot ID sites: [`daemon-mode.ts`](https://github.com/PrimeIntellect-ai/prime-agent/blob/v0.7.1/packages/coding-agent/src/modes/daemon/daemon-mode.ts#L3597), including replacement/catch-up around lines 6190 and 6388;
- shallow/lazy chunk materialization: [`snapshot-transcript-cache.ts`](https://github.com/PrimeIntellect-ai/prime-agent/blob/v0.7.1/packages/coding-agent/src/modes/daemon/snapshot-transcript-cache.ts#L27-L70);
- duplicate mismatch and fatal worker-channel handling: [`daemon-supervisor.ts`](https://github.com/PrimeIntellect-ai/prime-agent/blob/v0.7.1/packages/coding-agent/src/modes/daemon/daemon-supervisor.ts#L3883-L3920).

Neocher:
## Problem

Fixes #1229's kill-a-healthy-worker symptom.

Worker snapshot ids reuse `(activeSessionId, eventGeneration, lastEventSequence)` as transfer identity (daemon-mode.ts). When that position is reused with different bytes - e.g. a long-thinking turn with file reads produces multiple snapshots between event cursor advances - the supervisor's duplicate validation (daemon-supervisor.ts) throws `Duplicate snapshot ... did not match cached bytes` and kills a healthy worker. Users observe `Daemon worker client closed` on `-p --mode text` with `--thinking high/max` plus tool calls.

## Fix

Decouple snapshot identity from the event position. Added a monotonic per-session `snapshotSerial` to `ActiveSessionState`, incremented at each snapshot construction, appended to the id in all three construction sites (normal snapshot, replacement snapshot, and the attach-snapshot path). The id remains an opaque string on the wire, so there is no protocol or capability change.

## Validation

- `npm run check` passes (biome 909 files, tsgo, installer render, browser smoke)
- `test/agent-connection-daemon.test.ts`: 65/65 pass
- Reproduction: the previously-failing combo (`--mode text --offline --thinking high` + file-read tool call) now completes end-to-end when run from source

## Notes

This is the minimal stopgap the issue describes; a broader redesign of the snapshot identity/transfer model can build on it later. Behavior change: duplicate-snapshot byte mismatches can no longer occur for distinct content, so the supervisor's duplicate-validation path only sees true retransmissions.

> [!NOTE]
> ### Make daemon snapshot IDs unique with a per-session incrementing serial
> Snapshot IDs in the coding agent daemon were not unique across repeated emissions because they only encoded `(activeSessionId, eventGeneration, lastEventSequence)`. This adds a `snapshotSerial` counter to `ActiveSessionState` (initialized to 0) and appends an incrementing suffix to snapshot IDs in all three emission paths: initial/attach streaming, replacement snapshots, and catch-up snapshots.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 12f45e8.</sup>
> 
>

### DIFF
--- a/packages/coding-agent/src/modes/daemon/active-session-state.ts
+++ b/packages/coding-agent/src/modes/daemon/active-session-state.ts
@@ -40,6 +40,11 @@ export interface ActiveSessionState {
 	extensionUiRequests: Map<string, ActiveSessionExtensionUiRequest>;
 	eventGeneration: string;
 	lastEventSequence: DaemonEventSequence;
+	/** Monotonic per-session snapshot serial: decouples snapshot identity from
+	 *  the (eventGeneration, lastEventSequence) position, which can be reused
+	 *  with different content (fixes "Duplicate snapshot ... did not match
+	 *  cached bytes" killing healthy workers; upstream issue #1229). */
+	snapshotSerial: number;
 	unsubscribe?: () => void;
 	/** Latest background status summary, surfaced in the agents view. */
 	summaryState?: AgentStatus;
--- a/packages/coding-agent/src/modes/daemon/daemon-mode.ts
+++ b/packages/coding-agent/src/modes/daemon/daemon-mode.ts
@@ -1215,6 +1215,7 @@ export class AgentDaemon {
 			extensionUiRequests: new Map(),
 			eventGeneration: createActiveSessionId(),
 			lastEventSequence: 0,
+			snapshotSerial: 0,
 			clientEnv,
 		};
 		this.sessions.set(state.activeSessionId, state);
@@ -3595,7 +3596,7 @@ export class AgentDaemon {
 					});
 				}
 				if (streamsSnapshot) {
-					const snapshotId = `${state.activeSessionId}-${state.eventGeneration}-${state.lastEventSequence}`;
+					const snapshotId = `${state.activeSessionId}-${state.eventGeneration}-${state.lastEventSequence}-${state.snapshotSerial++}`;
 					let transcript: SnapshotTranscriptChunkSource;
 					try {
 						transcript = createSnapshotTranscriptChunks({
@@ -6199,7 +6200,7 @@ export class AgentDaemon {
 		state: ActiveSessionState,
 		message: Extract<DaemonOutbound, { type: "session_replaced" }>,
 	): void {
-		const snapshotId = `${state.activeSessionId}-${state.eventGeneration}-${state.lastEventSequence}`;
+		const snapshotId = `${state.activeSessionId}-${state.eventGeneration}-${state.lastEventSequence}-${state.snapshotSerial++}`;
 		// Mark before the registry read so later events queue behind this snapshot.
 		const snapshotSignal = markClientSnapshotStreaming(client, state.activeSessionId);
 		void this.prepareReplacementSnapshot(client, state, message, snapshotId, snapshotSignal).catch((error) => {
@@ -6397,7 +6398,7 @@ export class AgentDaemon {
 							),
 						});
 					}
-					const snapshotId = `${activeSessionId}-${state.eventGeneration}-${state.lastEventSequence}`;
+					const snapshotId = `${activeSessionId}-${state.eventGeneration}-${state.lastEventSequence}-${state.snapshotSerial++}`;
 					const snapshotSignal = markClientSnapshotStreaming(client, activeSessionId);
 					let transcript: SnapshotTranscriptChunkSource;
 					try {
--- a/packages/coding-agent/test/daemon-extension-binding.test.ts
+++ b/packages/coding-agent/test/daemon-extension-binding.test.ts
@@ -124,6 +124,7 @@ describe("daemon extension binding", () => {
 			extensionUiRequests: new Map(),
 			eventGeneration: "generation-slim",
 			lastEventSequence: 0,
+			snapshotSerial: 0,
 		};
 		await bindActiveSessionState(state, {
 			broadcast: (_state, message) => {
@@ -198,6 +199,7 @@ describe("daemon extension binding", () => {
 			extensionUiRequests: new Map(),
 			eventGeneration: "generation-test",
 			lastEventSequence: 0,
+			snapshotSerial: 0,
 			summaryState: { summary: "old recap", taskState: "completed", basedOnMessageCount: 2 },
 		};
 		await bindActiveSessionState(state, {
--- a/packages/coding-agent/test/suite/regressions/4257-update-restart-resume.test.ts
+++ b/packages/coding-agent/test/suite/regressions/4257-update-restart-resume.test.ts
@@ -77,6 +77,7 @@ function createState(
 		extensionUiRequests: new Map(),
 		eventGeneration: `generation-${activeSessionId}`,
 		lastEventSequence: 0,
+		snapshotSerial: 0,
 		...(options.clientEnv ? { clientEnv: options.clientEnv } : {}),
 	};
 }
--- a/packages/coding-agent/test/suite/regressions/4657-update-heartbeat-recovery.test.ts
+++ b/packages/coding-agent/test/suite/regressions/4657-update-heartbeat-recovery.test.ts
@@ -69,6 +69,7 @@ describe("ENG-4657 update heartbeat recovery", () => {
 			extensionUiRequests: new Map(),
 			eventGeneration: "restored-generation",
 			lastEventSequence: 0,
+			snapshotSerial: 0,
 		};
 		internals.sessions.set(activeSessionId, state);
 		internals.cronScheduler.start();


jonaowen (changes requested):
The serial allocation logic appears sound and wire-compatible, but this regression is not behaviorally tested yet.

1. Add a focused #1229 regression that forces two snapshot materializations at the same `(activeSessionId, eventGeneration, lastEventSequence)` with different transcript bytes. Assert distinct snapshot IDs and that the supervisor installs/caches both without closing the worker channel. Cover the attach, replacement, and catch-up allocation sites directly or through a shared allocator test.
2. `test/daemon-mode.test.ts` `makeState` and `test/suite/regressions/4602-snapshot-transfer-idempotency.test.ts` still omit `snapshotSerial`. Paths that postfix-increment those fixtures produce `undefined` and then `NaN`, while passing because IDs are not asserted. Initialize them to zero and assert the emitted IDs.

The protocol claim is otherwise credible: snapshot IDs are opaque strings, and a recovered worker gets a fresh event generation while same-state replacement retains the counter.

kevinjosethomas:

Issue #702, opened 2026-08-06
openai-codex models invisible to rlm subagents: model discovery sends prime-agent VERSION as codex client_version, backend returns empty catalog


chrisb4096-alt:
## Summary
`rlm()` child spawns and `find_models()` never offer any `openai-codex/*` model even with a valid, unexpired OAuth credential in `auth.json` — spawns fail with `Requested subagent model "openai-codex/gpt-5.6-sol" is unavailable, unauthenticated, or expired`.

## Root cause (verified on 0.7.0, macOS arm64)
`ModelRegistry.getExecutableModels()` does dynamic Codex model discovery against `https://chatgpt.com/backend-api/codex/models?client_version=<VERSION>` — and passes **prime-agent's own package version** (`0.7.0`) as `client_version`. The backend gates the catalog by *codex CLI* version, so `0.7.0` reads as an ancient client:

- `client_version=0.7.0` → HTTP 200, `{"models": []}` (empty)
- `client_version=0.146.1` (current codex CLI) → HTTP 200, 9 models (gpt-5.6-sol/terra/luna, gpt-5.5, gpt-5.4, ...)
- omitting the param → HTTP 400 (`Field required`)

Same bearer token, same `chatgpt-account-id`, same `originator` in all probes — only `client_version` changes the outcome. The empty set then filters every `openai-codex` model out of `getAvailable()`, `findRlmModels`, and `_resolveRlmSubagentModel` (agent-session.js), producing the misleading "unauthenticated, or expired" error despite `getProviderAuthStatus` being fresh.

## Suggested fix
Send a supported codex client version (or pin a known-good floor) in `openAICodexModelsUrl()` instead of the prime-agent package VERSION; optionally surface discovery-empty as a distinct error from auth failure.

## Local verification
Patching the URL builder to pin `0.146.1` immediately restores `openai-codex/*` in `find_models` and child spawns on a fresh session worker (no auth changes).

bilelrais:
Independent reproduction on **0.7.1, Linux x86_64** — same root cause, same single variable. Two additions that bear on the fix.

**1. The backend response is graded, not binary.** The catalog grows with the reported version rather than flipping empty/full:

| `client_version` | models returned |
|---|---|
| `0.7.0`, `0.7.1`, `0.7.2`, `0.8.0` | **0** |
| `0.100.0` | **4** |
| `0.143.0` | **6** |
[...]

PR #1375, opened 2026-08-13
fix(coding-agent): report a supported Codex client version for model discovery

bilelrais:
## Problem

`rlm` subagent spawns and `find_models()` never offer any `openai-codex/*` model, failing with `Requested subagent model "openai-codex/gpt-5.6-sol" is unavailable, unauthenticated, or expired` despite a valid, unexpired OAuth credential.

`ModelRegistry.getExecutableModels()` discovers Codex models against `https://chatgpt.com/backend-api/codex/models?client_version=<VERSION>` and passed Prime Agent's own package version. The backend gates its catalog on the Codex CLI version line, so `0.7.x` reads as an ancient client.

Measured against the backend with one valid token, varying only `client_version` (`originator` varied across `pi`, `codex`, `codex_cli_rs`, `codex_vscode` with no effect):

| `client_version` | models returned |
|---|---|
| `0.7.0`, `0.7.1`, `0.7.2`, `0.8.0` | 0 |
| `0.100.0` | 4 |
| `0.143.0` | 6 |
| `0.144.0`, `0.146.0`, `0.147.0` | 9 |

The response is graded rather than binary, so a floor that merely clears zero returns a silently partial catalog, which is harder to diagnose than the empty case because it looks like success.

The empty set then filters every `openai-codex` model out of `getExecutableModels()`. `/model` resolves through the unfiltered `getAvailable()`, so a user can be actively running `gpt-5.6-sol` as their main model while a subagent spawn reports it unavailable.

## Change

Replaced `VERSION` with a named `OPENAI_CODEX_CLIENT_VERSION` constant pinned to the current Codex CLI release (`0.147.0`), documented with the reason it exists and when to raise it. `VERSION` had no other use in the file, so its import was dropped.

## Testing

`packages/coding-agent/test/suite/regressions/702-codex-client-version.test.ts` builds a registry from a temporary `auth.json` holding a Codex OAuth credential, stubs `fetch`, and asserts the discovery URL reports a version other than Prime Agent's own and at or above the Codex CLI line, and that Codex models survive into `getExecutableModels()`. Verified to fail on the unfixed source (`expected '0.7.2' not to be '0.7.2'`) and pass with the change. No real provider APIs or tokens.

`npm run check` clean.

## Open

The pin needs raising when the Codex CLI gates newer models behind a later version; the comment says so, but nothing enforces it, and the failure mode is a silently smaller catalog.

Discovery failure is also indistinguishable from an auth failure at the call site: when the 5s `AbortSignal.timeout` fires or the request throws, Codex models are dropped and the caller reports "unavailable, unauthenticated, or expired". #702 suggests surfacing discovery-empty as a distinct error. That is a separate change to the error path and is not included here.

> [!NOTE]
> ### Fix Codex model discovery by reporting a supported `client_version` instead of package version
> The Codex backend gates its model catalog by the reported client version. The coding agent was sending its own package version, causing `openai-codex` models to be invisible to `rlm` subagents and `find_models`.
>
> - Introduces a fixed constant `OPENAI_CODEX_CLIENT_VERSION = "0.147.0"` in [model-registry.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/1375/files#diff-2fd8c4a2672b4ce27e73d494479df612c8b30ebac3dda5a160e2d861c612f453) that matches a known Codex CLI release.
> - Adds a regression test in [702-codex-client-version.test.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/1375/files#diff-9fc8217cf335dfa9678b228b20f710d3436ed51a659feca8a0e78305b6a62e5c) that asserts the discovery URL contains a semver-compatible `client_version` above a minimum threshold.
> - Behavioral Change: the `client_version` sent during Codex model discovery is now fixed at `0.147.0` and will not change automatically when the package version is bumped.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized c5680c0.</sup>
> 
>

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [Unreleased]
 
+- Fixed `openai-codex` models being invisible to `rlm` subagents and `find_models` because model discovery reported Prime Agent's own version as the Codex client version ([#702](https://github.com/PrimeIntellect-ai/prime-agent/issues/702)).
 - Fixed URLs not opening on click in fullscreen mode on terminals such as Ghostty; clicking a link in the transcript, dock, or overlays now opens it in the browser.
 
 ## [0.7.2] - 2026-08-11
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## [Unreleased]
 
-- Fixed `openai-codex` models being invisible to `rlm` subagents and `find_models` because model discovery reported Prime Agent's own version as the Codex client version ([#702](https://github.com/PrimeIntellect-ai/prime-agent/issues/702)).
+- Fixed `openai-codex` models being invisible to `rlm` subagents and `find_models` because model discovery reported Prime Agent's own version as the Codex client version ([#1375](https://github.com/PrimeIntellect-ai/prime-agent/pull/1375) by [@bilelrais](https://github.com/bilelrais)).
 - Fixed URLs not opening on click in fullscreen mode on terminals such as Ghostty; clicking a link in the transcript, dock, or overlays now opens it in the browser.
 
 ## [0.7.2] - 2026-08-11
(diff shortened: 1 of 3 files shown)


mrmoxon (approved):
Verified the Codex-only subagent failure path, current Codex CLI compatibility version, focused regression, and full repository checks. No blocking findings.

Issue #1668, opened 2026-08-23
IPython stays permanently unavailable after an unexpected kernel exit


Atroci:
## Description

On Prime Agent v0.8.0, an unexpected IPython kernel exit leaves `IpythonKernelProvisioner` holding the resolved stopped manager. Every later Python tool call receives that manager and fails. The session remains responsive, but Python stays unavailable until the process restarts.

## Steps to reproduce

1. Start Prime Agent v0.8.0 and execute one Python cell.
2. Stop the provisioned IPython kernel unexpectedly.
3. Execute another Python cell in the same session.
4. Execute another Python cell again.

## Expected behavior

The next explicit Python call provisions a fresh kernel. It restores the latest snapshot through the existing startup path.

The failed cell must not replay automatically. This avoids duplicate external effects.

## Actual behavior

Every later Python call reuses the stopped cached manager and fails with:

`Kernel has been shut down`

Python remains unavailable while the surrounding agent session stays responsive.

## Prime Agent version

v0.8.0

## Operating system

Linux

## Additional context

Root cause is in `IpythonKernelProvisioner.ensure()`. Its cached `managerPromise` remains resolved after the started manager stops.

A tested fix clears `managerPromise`, `startedManager`, `_lastRestore`, and `kernelManagerRef.current` before normal provisioning continues.

The fix reuses existing startup, snapshot restoration, and runtime bootstrap behavior. Concurrent recovery calls remain coalesced onto one startup.

Validation:

- `test/ipython-provisioner.test.ts`: 17 tests passed.
- `npm run check`: passed.
- Live recovery smoke returned `RECOVERED` after a forced kernel exit.

Implementation commit: Atroci/prime-agent@9c09cc8bf59396eabba87d4ce3c9f4ba91ccd8c1

Closed contribution PR: #1658

Bug discussion: #1659

MAB2908:
Reproduces on v0.8.1 (Linux x86_64, Node v22.23.2 standalone, kernel venv Python 3.11.16 / ipykernel 7.3.0 / pyzmq 27.2.0).

Same signature as described here: after the kernel exits unexpectedly (in my case it was killed by a `prime-agent shutdown --force` of the daemon), every later `ipython` call in the session returns only `Kernel has been shut down`. 9/9 attempts, no recovery, `/restart` in-session does not help. Only `prime-agent shutdown --force` + client restart revives the session.

Two extra observations beyond #1668:

[...]

PR #1696, opened 2026-08-24
fix(coding-agent): protect forkserver from status probes

sirouk:
## Summary

- exclude internal forkserver control sockets from daemon discovery
- prevent a second connection from replacing the established forkserver channel
- anchor the exclusion by real path, and promote the control channel only on the child's `ready` handshake
- cover both parser paths, the symlinked-tmpdir fail-open, the startup race, and the live forkserver protocol

## Why

`prime-agent status --json` scans Unix listeners owned by `prime-agent`, probes the internal forkserver socket, and disconnects. The disconnect marks the forkserver dead and terminates every IPython kernel forked from it.

This prevents the initiating teardown. It is distinct from #1659 and closed #1693, which recover after a kernel has already died.

@coleleavitt independently reproduced the fault and closed two gaps in the first version: a `resolve()`-only tmpdir compare that failed open when the socket is reported under an aliased path such as `/private/tmp`, and a startup race where a probe landing between `listen()` and the child's connect still reached `markDead()`. Both now have regressions that fail on the pre-hardening tree.

#1687 removes `fork-server.ts` in the REPL cutover. Until that stack lands both halves apply; after it lands the connection guard belongs on the REPL host, and the `daemon-ps` exclusion still stands.

Discussion: #1695

## Validation

macOS 27 arm64 (node 22.22) and Debian 12 aarch64 in docker (node 22.23):

- `test/daemon-ps.test.ts test/kernel-fork-server-protocol.test.ts`: 38 passed, 1 platform-gated skip on both
- `npm run check`: exit 0 on both, with the Linux run from a fresh clone of the branch
- each guard checked by reverting it: the discovery filter, the connection guard, the realpath anchor, and the handshake gate each fail their tests when removed
- a live forkserver socket driven through the real `ss` and `lsof` discovery paths is excluded, and both the forkserver and its forked kernel survive the probe
- full `npm run test:ci` on macOS shows no failure attributable to this change; the residual failures reproduce identically on unmodified `main`

> [!NOTE]
> ### Exclude forkserver control sockets from daemon discovery and guard control connection
> - `prime-agent status` was discovering internal kernel forkserver control sockets and connecting to them, which terminated active IPython forkservers. Discovery now skips these sockets.
> - New helper `isKernelForkServerSocketPath` in [daemon-ps.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/1696/files#diff-d26d176c2fa21f2284ca0dc614dfb273ec8abfce710655afccc3aa35628c848a) detects sockets whose basename is `control.sock`, whose parent dir starts with `prime-agent-forkserver-`, and whose grandparent resolves to the system tmpdir.
> - `parseSsListeners` (Linux) and `parseLsofListeners` (macOS) both filter out matching sockets before returning daemon discovery results.
> - `ForkServer` in [fork-server.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/1696/files#diff-65164112dcd341a9e27f5bbd59aa6b335e99b3c538915cde2701e2afa02a5e36) now destroys any inbound connection after the initial control connection is established, preventing probes from replacing the control channel.
> - Risk: any out-of-tree code that relied on discovering forkserver control sockets via `parseSsListeners` or `parseLsofListeners` will no longer see them in the results.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized ee8ede1.</sup>
> 
>

### DIFF
--- a/packages/coding-agent/.changes/fix-status-forkserver-probe.md
+++ b/packages/coding-agent/.changes/fix-status-forkserver-probe.md
@@ -1,0 +1,1 @@
+- Fixed `prime-agent status` terminating active IPython forkservers by excluding internal control sockets from daemon discovery and preserving the established forkserver connection.
--- a/packages/coding-agent/src/cli/daemon-ps.ts
+++ b/packages/coding-agent/src/cli/daemon-ps.ts
@@ -1,5 +1,6 @@
 import { spawnSync } from "node:child_process";
 import { existsSync, lstatSync, readdirSync, readFileSync, rmSync, unlinkSync } from "node:fs";
+import { tmpdir } from "node:os";
 import { basename, dirname, join, resolve } from "node:path";
 import chalk from "chalk";
 import { APP_NAME, getAgentDir, VERSION } from "../config.js";
@@ -99,7 +100,11 @@ export function parseSsListeners(stdout: string, appName: string): DiscoveredDae
 		if (!owner || !processNameMatches(owner[1]!, appName)) {
 			continue;
 		}
-		daemons.push({ pid: Number.parseInt(owner[2]!, 10), socketPath: normalizeSocketPath(socketPath) });
+		const normalized = normalizeSocketPath(socketPath);
+		if (isKernelForkServerSocketPath(normalized)) {
+			continue;
+		}
+		daemons.push({ pid: Number.parseInt(owner[2]!, 10), socketPath: normalized });
 	}
 	return daemons;
 }
@@ -116,6 +121,9 @@ export function parseLsofListeners(stdout: string): DiscoveredDaemonProcess[] {
 			pid = Number.parseInt(value, 10);
 		} else if (field === "n" && pid !== undefined && value.startsWith("/")) {
 			const socketPath = normalizeSocketPath(value);
+			if (isKernelForkServerSocketPath(socketPath)) {
+				continue;
+			}
 			const key = `${pid}:${socketPath}`;
 			if (!seen.has(key)) {
 				seen.add(key);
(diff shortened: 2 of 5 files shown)


sirouk:
macOS is fixed on the same branch: [`7982817`](https://github.com/sirouk/prime-agent/commit/79828172c31e0c249e4fe8f6faa613404e1d3b39) and [`bb54dea`](https://github.com/sirouk/prime-agent/commit/bb54deaba3a15c14ef556dcd3b5351d8015a2b51). `ee8ede1` is untouched.

Cause: the fixtures hardcoded `/tmp`, but `isKernelForkServerSocketPath` anchors to `os.tmpdir()`, which is a per-user path on macOS. The predicate correctly rejected the `/tmp` fixture, so three parser tests failed. The fixtures now derive the path from `os.tmpdir()`, matching the `defaultDaemonSocketDir()` convention the worker-socket test above them already uses, and one negative case pins the temp-dir anchor. No product code changed. `bb54dea` adds the external-contribution attribution to the changelog fragment.

Validation on macOS 27 arm64 (node 22.22) and Debian 12 aarch64 in docker (node 22.23):

- `test/daemon-ps.test.ts test/kernel-fork-server-protocol.test.ts`: 36 passed, 1 platform-gated skip on both.
- `npm run check`: exit 0 on both, with the Linux run done from a fresh clone of the branch.
- Reverting either half of the fix fails the tests on macOS: dropping the discovery filter fails 3, dropping the connection guard fails the probe test.
- Full `npm run test:ci` on macOS shows no failure attributable to this change. The residual failures reproduce identically on unmodified `main`.

Also confirmed on macOS that `lsof -nP -F pn -U` reports the bound `$TMPDIR` path rather than the `/private` realpath, so the discovery filter matches there at runtime.

Adjacent work I checked against, so nothing here duplicates it:

- #1693 and #1659 / #1668 are the recovery half and share no files with this PR. This one prevents the death; that one restores a kernel that already died.
- #1477 is a separate `fork-server.ts` lifecycle bug, `beforeExit` disposal orphaning live kernels. It reads well next to the connection guard.
- #1687 removes `fork-server.ts`, `fork-server-script.ts`, and `test/kernel-fork-server-protocol.test.ts` in the REPL cutover. If that stack lands, the connection guard belongs on the REPL host instead, while the `daemon-ps` exclusion is still needed until it does. I can rework this on top of it.
- #1523 touches the same two files. Merging it into this branch conflicts in one place, the `node:os` import in `test/daemon-ps.test.ts`, and `daemon-ps.ts` merges cleanly. Its new probe and repair path also gains from the exclusion.
- The protocol test extended here came from #1559, and the exclusion mirrors `isWorkerSocketPath` from #423.

This page still shows `ee8ede1` because a closed PR does not track branch pushes. Current diff: https://github.com/PrimeIntellect-ai/prime-agent/compare/main...sirouk:prime-agent:fix/status-forkserver-probe

I left it closed since reopening re-runs the contribution gate.

PR #1700, opened 2026-08-24
fix(coding-agent): avoid retrying delivered remote agent messages

snimu:
## What was wrong

Sending a message to another agent could deliver it twice. The code retried whenever it didn't get a confirmation back — but "no confirmation" often just means the reply got lost, not that the message wasn't delivered. So on a timeout or dropped connection, it would send the same message again.

## The fix

Split the work into two phases:
- Connecting to the daemon can be retried freely (nothing has been sent yet).
- Once connected, the message is sent exactly once. If anything goes wrong after that, the caller gets an error instead of a silent duplicate.

No new IDs, dedupe logic, or protocol changes — just stop retrying the part that isn't safe to retry.

## How it's verified

A new test has the server read the message and then kill the connection: the old code reconnected and sent a duplicate; the new code surfaces the error and the server receives exactly one copy. Type checks, lint, and the full test suite pass (the 4 daemon-mode failures also fail on main and are unrelated).

Note: intentionally no Linear ticket for this cleanup stack, so that check stays red.

---

> [!NOTE]
> **Medium Risk**
> Changes inter-agent messaging delivery semantics: callers may see errors where retries previously succeeded, but duplicate delivery risk is reduced.
> 
> **Overview**
> **Fixes duplicate remote agent messages** when supervisor RPCs time out or responses are lost after the message was already accepted.
> 
> `sendRemoteAgentSessionMessage` now **retries only connection and hello** inside the existing deadline loop. After a client is connected, it issues **one** `send_message` request on that socket; if the connection drops or the response never arrives, the call **fails with an error** instead of reconnecting and sending again.
> 
> A daemon-mode test simulates the supervisor receiving the request then destroying the socket and asserts **exactly one** `send_message` and a connection-closed error.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 798b2e1477381f9c47ca53ecffdabff05b6ad418. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

Linear: [ENG-5646](https://linear.app/primeintellect/issue/ENG-5646/fixcoding-agent-avoid-retrying-delivered-remote-agent-messages)

> [!NOTE]
> ### Fix `AgentDaemon.sendRemoteAgentSessionMessage` to send remote agent messages exactly once
> - Refactors [daemon-mode.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/1700/files#diff-f657fb72071b9bcb67235932ddc24164a0376dc918944a8f58b857c2ac43e450) so connection/hello is retried within the deadline, but `send_message` is performed exactly once per call
> - Errors that occur after the message is sent (response loss or timeout) are now surfaced instead of triggering a resend, preventing duplicate delivery
> - Adds a test in [daemon-mode.test.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/1700/files#diff-e6ffcbd62d8fe753528e06c93d849d9b1799ac2dbadf243304375c6a043243b5) that verifies no retry happens when the supervisor drops the socket after receiving the message
> - Behavioral Change: a lost or timed-out response to `send_message` now surfaces an error rather than resending; previously the method could resend and deliver the message twice
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 798b2e1.</sup>
> 
>

### DIFF
--- a/packages/coding-agent/.changes/remote-message-single-send.md
+++ b/packages/coding-agent/.changes/remote-message-single-send.md
@@ -1,0 +1,1 @@
+- Fixed remote agent messages being delivered twice when the daemon request timed out or the response was lost: the message is now sent exactly once per call, and post-send failures surface as errors instead of triggering a resend.
--- a/packages/coding-agent/src/modes/daemon/daemon-mode.ts
+++ b/packages/coding-agent/src/modes/daemon/daemon-mode.ts
@@ -5740,42 +5740,45 @@ export class AgentDaemon {
 			throw new Error(`Unknown active session: ${targetSelector}`);
 		}
 		const deadline = Date.now() + 30_000;
+		let client: DaemonClient | undefined;
 		let lastError: unknown;
 		while (Date.now() < deadline && !this.shuttingDown) {
-			const client = new DaemonClient(supervisorSocketPath);
-			let receivedResponse = false;
+			const candidate = new DaemonClient(supervisorSocketPath);
 			try {
-				await client.connect(1000);
-				await client.waitForHello(1000);
-				const response = await client.request(
-					{
-						type: "send_message",
-						targetActiveSessionId: targetSelector,
-						message,
-						fromActiveSessionId: fromState.activeSessionId,
-						agentOrigin: true,
-					},
-					30_000,
-				);
-				receivedResponse = true;
-				if (!response.success) {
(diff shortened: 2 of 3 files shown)


alexzhang13 (approved):
lgtm

PR #1782, opened 2026-08-26
fix(runtime): honor mcp.types.Tool's inputSchema alias

### LINKED ISSUE #1073: McpIntegration.list_tools() always returns an empty inputSchema (camelCase getattr vs snake_case SDK field)

kaluli123123:
Fixes #1073.

Posted for review after discussion in #1757 (per CONTRIBUTING.md's contribution process).

## Summary

`McpIntegration.list_tools()` (`prime-agent-runtime/src/rlm/mcp_base.py`) always returns `"inputSchema": {}` for every tool of every MCP server, even though the server sends a full JSON Schema. This was already reported and diagnosed in #1073, which is one of the "related reports" under the still-open tracker #1383 (*Harden MCP OAuth discovery and Codex transports* — "preserve MCP tool schemas ..."). The merged stack PR #1164 for that tracker didn't touch this file, so the bug is still present on `main` (verified on `06860844e`).

## Root cause

`mcp.types.Tool` is a pydantic model whose schema field is the snake_case `input_schema`, with a camelCase **alias** `inputSchema`:

```python
from mcp.types import Tool
Tool.model_fields['input_schema'].alias  # -> 'inputSchema'
```

Aliases are honored on validation and on `model_dump(by_alias=True)`, never as a plain attribute. So the existing code

```python
```

always misses and silently substitutes `{}`. The same value feeds the per-tool docstring in `__getattr__`, so the documented discovery path ("call `help(<tool>)` for the argument schema") returns nothing for every MCP integration, and callers have to reverse-engineer arguments from the server's validation errors one field at a time — this is exactly what the original report describes with Context7.

The existing regression test in `test_mcp_base.py` didn't catch this because its fake session built tools from a bare duck-typed stub that set the camelCase attribute directly (`t.inputSchema = schema`), which happens to satisfy the buggy `getattr` and therefore encodes the bug rather than catching it.

## Fix

Read the schema via `model_dump(by_alias=True)` first (falls back to plain `getattr` for non-pydantic stand-ins used in tests):

```python
def _tool_input_schema(t: Any) -> dict[str, Any]:
    if hasattr(t, "model_dump"):
        schema = dumped.get("inputSchema")
        if schema:
            return schema
    return getattr(t, "inputSchema", None) or {}
```

Also rebuilt the test double to construct a real `mcp.types.Tool` instead of the duck-typed stub, and added a regression test asserting the schema round-trips through `list_tools()` and the bound-tool docstring. Confirmed the new test fails against the pre-fix code and passes after the fix.

## Validation

```
cd prime-agent-runtime
uv run python -m unittest discover -s test   # 99 passed (was 98; +1 new test)
```

Also ran the full `npm run check` at the repo root (biome, `tsgo --noEmit`, installer render check, browser smoke check) — all green; this change doesn't touch `packages/*`, so no `.changes/` fragment is required by the changelog-fragment CI check.

### DIFF
--- a/prime-agent-runtime/src/rlm/mcp_base.py
+++ b/prime-agent-runtime/src/rlm/mcp_base.py
@@ -33,6 +33,25 @@ __all__ = ["McpIntegration", "McpToolError", "NotEnabled"]
 _EXPIRY_SKEW_SECONDS = 30
 
 
+def _tool_input_schema(t: Any) -> dict[str, Any]:
+    """Read a tool's JSON Schema, honoring the SDK's ``inputSchema`` alias.
+
+    ``mcp.types.Tool`` is a pydantic model whose field is the snake_case
+    ``input_schema`` with a camelCase alias ``inputSchema``. Aliases are only
+    honored on validation and ``model_dump(by_alias=True)``, never as a plain
+    attribute, so a bare ``getattr(t, "inputSchema", None)`` always misses and
+    silently returns an empty schema. Dump by alias first; fall back to
+    ``getattr`` for duck-typed stubs (e.g. in tests) that set the attribute
+    directly rather than being a real pydantic model.
+    """
+    if hasattr(t, "model_dump"):
+        dumped = t.model_dump(by_alias=True, exclude_none=True)
+        schema = dumped.get("inputSchema")
+        if schema:
+            return schema
+    return getattr(t, "inputSchema", None) or {}
+
+
 class NotEnabled(RuntimeError):
     """Raised when an integration has no usable credentials.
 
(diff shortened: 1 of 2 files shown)


kaluli123123:
Understood re: the vouching policy. Filed here for the record and cross-linked from the corresponding Discussion (#1757) so a maintainer can pick up the diff directly if useful -- reproduction, fix, and regression test are documented there and in this PR's description. Happy to re-open or resubmit from a vouched account if that's ever preferred.

PR #895, opened 2026-08-08
fix(coding-agent): resume suspended input pump on programmatic admission

twaldin:
## Problem

Any abort — `requestAbort()` from Esc/TUI interrupt, a daemon `abort`, a goal abort, or `abortForUpdateRestart()` — sets `_sessionInputPumpSuspended = true`. The user-typed prompt path (`_prompt`) clears that flag when the session is not streaming, but the programmatic admission path (`queueAgentMessagePrompt` → `_queuePreparedPrompt` → `_admitSessionInput`) never does. After one abort, every programmatically queued item — agent messages from children, heartbeat prompts, injected wake prompts — sits in the queue forever while the session is idle. Only a human-typed prompt or an explicit `resume_queue` revives delivery.

Reproduced clean-room on a pristine 0.7.0 install with a stub provider, driven over the daemon socket:

| scenario | result |
|---|---|
| long tool call / long streaming turn with items queued | delivered at boundary — fine |
| idle session, items queued | delivered immediately — fine |
| **abort during streaming with items queued** | **starved forever** (`queuedCount` persisted >90s idle) |
| then one user-typed prompt | flushes the entire starved backlog in ~2s |
| starved queue + heartbeat cron due | heartbeat silently skipped every fire |

The starved queue also blackholes heartbeats: `shouldDeferHeartbeatCronJob` defers while `unfinishedActionCount > 0`, and a skipped fire is lost until the next full interval because `nextRunAt` is recomputed from the skip time (adjacent to #820).

Related: #820 (heartbeat fires dropped when coincident with busy), #821 (send retry duplication), #823 (queued agent mail delivery). None of them names the suspended-pump starvation itself.

## Fix

- `_admitSessionInput`: when a programmatic turn admission arrives while the session is not streaming and the pump is suspended, clear the suspension, notify checkpoint waiters, and schedule the pump — the same resume semantics `_prompt()` already gives user-typed input.
- `recordDispatchResult`: a skipped heartbeat fire reschedules to the next scheduler tick (1 minute) instead of the next full interval, so a deferred fire retries instead of being lost.

## Testing

- New regression test `resume-suspended-input-pump.test.ts`: queued agent mail after an idle abort delivers; a backlog queued before an abort drains once a later programmatic admission arrives. Both fail without the fix.
- Updated `cron-jobs.test.ts` skip-reschedule expectation to the new retry-next-tick semantics.
[...]

PR #1783, opened 2026-08-26
fix(coding-agent): resume suspended input pump on programmatic admission

### LINKED ISSUE #1000: Programmatic prompts starve at idle sessions: input pump never resumes (agent messages queue for hours until a human types)
## Summary

Programmatic prompts (agent messages from subagents, scheduled follow-ups) queued at an **idle** session are never delivered until a human types a prompt. In our production sessions, batches of subagent results sat queued for **8+ hours** and all flushed in the same second a human message arrived.

## Root cause (two code paths in `_admitSessionInput`)

1. After an abort sets `_sessionInputPumpSuspended`, a typed prompt clears it (`_prompt()` path), but a programmatic admission does not: the flag is only cleared for `wake: "immediate"` turn actions, so queued steer-delivery actions wait forever at idle.
2. Worse, and independent of any abort: when an incoming follow-up **coalesces** into an already-queued owner (same queue key — the normal shape for any scheduler that re-fires, or repeated subagent messages), `_admitSessionInput` returns `{accepted: false, disposition: "queued"}` **before any code that schedules the session input pump**. Every re-fire is swallowed; nothing ever restarts delivery.

Reproduction is deterministic on a pristine install (stub provider): idle session + queued programmatic agent message → no turn starts within any timeout; a coalescing re-fire also never resumes the pump. A vitest regression that fails on unpatched code is included in PR #895.

## Why most users never see it

The trigger requires an idle session receiving only programmatic input with coalescing keys. Interactive users always type (which resumes the pump via `_prompt()`), so the starvation is invisible to them. Any headless/orchestrated deployment that relies on `agent_message` or scheduled follow-ups hits it immediately. Issue #823 (queued agent mail trickling one message per turn) looks adjacent to the same pump behavior.

## Fix

PR #895 resumes the suspended pump on programmatic admission (matching `_prompt()`), including on the coalesced early-return path, with a failing-first regression test. Happy to adjust to your preferred shape.

## Impact

Any multi-agent/headless deployment: subagent results, scheduled work, and cron-style follow-ups silently starve at idle sessions. For us this froze overnight autonomous operation until a human keystroke.

kaluli123123:
Fixes #1000.

Posted for review after discussion in #1761 (per CONTRIBUTING.md's contribution process).

## Summary

Programmatic prompts (agent messages from subagents, scheduled follow-ups/heartbeats) queued at an **idle** session are never delivered until a human types a prompt. In one report, batches of subagent results sat queued for 8+ hours and all flushed in the same second a human message arrived. This is #1000, one of the "related reports" under the still-open tracker #1382 (*Complete queued, interrupted, and archived session lifecycle recovery*).

Confirmed still present on `main` (`06860844e`) with a new regression test suite that reproduces all three shapes described in the original report.

## Root cause (`_admitSessionInput` in `packages/coding-agent/src/core/agent-session.ts`)

1. `requestAbort()` sets `_sessionInputPumpSuspended = true`. A typed prompt clears it via `_resumeSessionInputAdmission()` (the `prompt()` path), but a programmatic admission does not: the flag is only cleared when `action.payload.kind === "turn" && action.wake === "immediate"`, so queued follow-ups/steer actions with the default wake behavior wait forever at idle.
2. Independent of any abort: when an incoming follow-up **coalesces** into an already-queued owner (the normal shape for a re-firing scheduler, or a repeated agent message with the same `queueKey`), `_admitSessionInput` returns `{ accepted: false, disposition: "queued" }` **before any code that schedules the pump**. Every re-fire is swallowed; nothing ever restarts delivery.

## History

An earlier PR attempt, #895, proposed a fix and was closed in favor of a maintainer-owned stacked PR, #1162 (part of the merged #1158–#1165 stack for tracker #1382):

> This root cause is now covered by maintainer-owned stacked PR #1162 ... We did not inspect or reuse this PR's diff, branch, commits, implementation code, or tests.

#1162 merged, but this specific code path was not part of it — the bug is still live.

Notably, #895's proposed patch toggled `_sessionInputPumpSuspended` directly without checking `_sessionInputSuspendedForUpdateRestart`. That flag is a *different*, deliberate suspension reason (`abortForUpdateRestart()`) meant to hold queued inputs so they survive into the restart manifest instead of starting a turn during teardown (see the existing guarded call site around line 6197: `if (!this._sessionInputSuspendedForUpdateRestart) this._resumeSessionInputAdmission();`). Blindly clearing the flag would have silently reintroduced a second bug — a turn starting mid-restart-teardown — while fixing the first one.

## Fix

Resume a suspended pump on both paths (the coalesced early return, and the general admission path) via the existing `_resumeSessionInputAdmission()` helper, gated on: the action is a `turn`, the session isn't streaming, the pump is actually suspended, and **the suspension is not a restart hold**.

```ts
if (
	!options.restore &&
	options.wake !== false &&
	action.payload.kind === "turn" &&
	!this.isStreaming &&
	this._sessionInputPumpSuspended &&
	!this._sessionInputSuspendedForUpdateRestart
) {
	this._resumeSessionInputAdmission();
	this._scheduleSessionInputPump();
}
```

(mirrored in the coalesced-owner early-return branch above it).

## Tests

Added `packages/coding-agent/test/suite/regressions/resume-suspended-input-pump.test.ts` with 4 cases:
- delivers a queued agent message at idle after an abort suspended the pump
- delivers backlog queued before an abort once a later programmatic admission arrives
- resumes the pump when a coalesced follow-up re-fires at idle
- does **not** resume a pump suspended for an update restart — verified this case fails if the `_sessionInputSuspendedForUpdateRestart` guard is removed, so it actually exercises that hazard

All 4 fail against pre-fix `main` and pass after the fix. Also ran:
- `test/suite/agent-session-queue.test.ts`, `agent-session-action-contracts.test.ts`, `agent-session-action-races.test.ts`, `agent-session-autonomous.test.ts`, `agent-session-goal.test.ts` (195 passed)
- `test/suite/regressions/4257-update-restart-resume.test.ts`, the existing large update-restart suite (35 passed)
- A broader run of the full `packages/coding-agent` suite surfaced 3 unrelated pre-existing failures (`resource-loader.test.ts`, `package-command-paths.test.ts`, `4600-supervisor-singleton.test.ts`); confirmed by re-running them against unmodified `main` that they fail identically there (sandbox-specific: symlink/extension loading and real daemon-socket timing), unrelated to this change.
- Root `npm run check` (biome, `tsgo --noEmit`, installer render, browser smoke) — all green. Added a `packages/coding-agent/.changes/` fragment per the changelog-fragment CI check.

### DIFF
--- a/packages/coding-agent/.changes/eng-resume-suspended-input-pump-programmatic.md
+++ b/packages/coding-agent/.changes/eng-resume-suspended-input-pump-programmatic.md
@@ -1,0 +1,1 @@
+- Fixed queued agent messages and scheduled follow-ups starving for hours at an idle session after an abort suspended the input pump; only a typed prompt used to resume delivery.
--- a/packages/coding-agent/src/core/agent-session.ts
+++ b/packages/coding-agent/src/core/agent-session.ts
@@ -5541,6 +5541,22 @@ export class AgentSession {
 		}
 		const coalescedOwner = options.restore ? undefined : this._coalescedFollowUpOwner(action);
 		if (coalescedOwner) {
+			// A coalesced re-fire (the normal shape for a re-firing scheduler, or a
+			// repeated agent message) is still evidence a producer wants delivery.
+			// Resume a pump suspended by an abort so the owning queued action can
+			// drain at idle; otherwise this early return swallows the re-fire and
+			// nothing ever restarts the pump (see #1000). A restart-hold suspension
+			// is left alone: queued inputs must survive into the restart manifest.
+			if (
+				!options.restore &&
+				options.wake !== false &&
+				!this.isStreaming &&
+				this._sessionInputPumpSuspended &&
+				!this._sessionInputSuspendedForUpdateRestart
+			) {
+				this._resumeSessionInputAdmission();
+				this._scheduleSessionInputPump();
+			}
 			if (action.agentMessageId !== coalescedOwner.agentMessageId) {
 				this._rejectAgentMessage(
 					action.agentMessageId,
@@ -5575,6 +5591,23 @@ export class AgentSession {
 			if (action.payload.kind === "turn" && action.wake === "immediate") {
 				this._resumeSessionInputAdmission();
 			}
+			this._scheduleSessionInputPump();
+		}
+		if (
+			!options.restore &&
+			options.wake !== false &&
+			action.payload.kind === "turn" &&
+			!this.isStreaming &&
+			this._sessionInputPumpSuspended &&
+			!this._sessionInputSuspendedForUpdateRestart
+		) {
+			// Programmatic admissions (agent messages, scheduled follow-ups) must resume
+			// a pump suspended by an abort, matching what a typed prompt already does via
+			// _resumeSessionInputAdmission(); otherwise queued work starves at idle until
+			// a human types (see #1000). A restart-hold suspension is left alone: queued
+			// inputs must survive into the restart manifest instead of starting a turn
+			// during teardown.
+			this._resumeSessionInputAdmission();
 			this._scheduleSessionInputPump();
 		}
 		return { accepted: true, disposition, ticket: controller.ticket };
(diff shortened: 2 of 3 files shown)


kaluli123123:
Understood re: the vouching policy. Filed here for the record and cross-linked from the corresponding Discussion (#1761) so a maintainer can pick up the diff directly if useful -- reproduction, fix, and regression test are documented there and in this PR's description. Happy to re-open or resubmit from a vouched account if that's ever preferred.

Issue #709, opened 2026-08-06
Validate session entries and make append rollback re-entrancy-safe


morluto:
## Problem

Session JSONL data is deserialized with a TypeScript cast and then consumed by the session index, context builder, metadata scanner, and export/TUI paths without a runtime shape check. A malformed or hand-edited row can therefore reach code that assumes fields such as `message.role`, tree IDs, usage, or compaction targets exist.

The append rollback path has a separate re-entrancy hazard. `_appendEntryWithRollback` currently identifies the failed append by removing the last entry. A persist listener can append its own entry during notification, making that last entry belong to the listener rather than the caller whose append failed.

## Proposed outcome

Make session persistence integrity explicit at two boundaries:

```text
JSONL rows --> parse --> runtime shape guard --> migrate/index/context
                          invalid rows dropped
                          legacy rows normalized

caller append --> persist + listeners --> failure --> rollback caller entry by identity
                                             listener-owned entries remain
```

The guard should preserve supported legacy session formats while rejecting malformed current entries and orphaned tree rows. Rollback should remove only the entry owned by the failed append and preserve re-entrant listener state.

## Evidence

Regression coverage should include:

- missing message payloads and incomplete known entry types;
- unknown/garbage rows;
- current-version missing IDs and orphaned children;
- v1 entries and compaction indexes after filtering;
- legacy messages with omitted/null content;
- assistant thinking/tool-call metadata scanning;
- a re-entrant persist listener that appends before the outer persist fails.

Related scheduler and saved-session work in [#540](https://github.com/PrimeIntellect-ai/prime-agent/pull/540), [#576](https://github.com/PrimeIntellect-ai/prime-agent/pull/576), and [#584](https://github.com/PrimeIntellect-ai/prime-agent/pull/584) does not cover these deserialization or rollback ownership boundaries.

kevinjosethomas:
Hi, thanks for taking the time to report this! Since open sourcing Prime Agent, we’ve received far more Issues than we can reliably triage and maintain. Rather than leave a large backlog that we cannot meaningfully work through, we’re closing the current Issue queue and moving to a discussion-first process.

[...]

PR #1785, opened 2026-08-26
fix(coding-agent): repair an incomplete JSONL tail before resuming session appends

### LINKED ISSUE #928: Repair an incomplete JSONL tail before resuming session appends
## Problem and impact

After a crash interrupts the final JSONL write, the next valid append is concatenated to the fragment. Both records become one malformed line, so the new entry is also lost and the session tree can break.

## Reproduction

Create a valid session ending in a truncated JSON object without newline, resume it, append a valid entry, and reload. The combined line is discarded.

## Evidence

[`packages/coding-agent/src/core/session-manager.ts:613-630`](https://github.com/PrimeIntellect-ai/prime-agent/blob/a18809e0/packages/coding-agent/src/core/session-manager.ts#L613-L630) silently skips an invalid final line; `:1231-1260` resumes without repairing it; `:1467-1473` appends directly to the existing bytes.

## Acceptance criteria

- Resume distinguishes a complete final record from an incomplete tail.
- Invalid tail bytes are quarantined or truncated before append.
- A subsequent entry round-trips.
- Earlier valid records are preserved.
- Recovery is observable.

## Solution design

Validate the terminal byte range at open, retain recovery metadata, and rewrite/truncate under the session lease before marking the file flushed.

## Verification plan

Tests for mid-UTF-8, mid-JSON, missing newline on valid JSON, malformed middle row, empty file, and crash between repair and append.

## Related

Related to #709, but not covered by its row-validation and rollback scope.

## Audit metadata

- Audit ID: `P1-07`
- Priority: `P1`
- Effort: `M`
- Implementation risk: `MED`
- Confidence: `HIGH`
- Suggested repository labels: `bug`
- Package scope: `pkg:coding-agent`
- Audited revision: `fa900d5a` (local merge tree)
- Evidence revision: [`a18809e0`](https://github.com/PrimeIntellect-ai/prime-agent/commit/a18809e0)
- Audit date: 2026-08-08

The issue author could not apply labels. Apply the suggested repository labels above and the package-scope labels once they are provisioned.

The audited tree differs from the linked upstream evidence revision only by an unrelated local design document.

kaluli123123:
Fixes #928.

Posted for review after discussion in #1764 (per CONTRIBUTING.md's contribution process).

## Summary

A crash mid-write can leave a session's `.jsonl` file's final record truncated. This is #928, one of the "related reports" under the still-open tracker #1380 (*Make persisted session and configuration state crash-safe*). Reproduced fresh on current `main` (`06860844e`) with a new failing-first regression test — not just read from the source.

## Root cause

`parseEntriesFromBuffer` (and friends) already silently skip any line that fails `JSON.parse`, including a truncated trailing fragment — so in-memory state after resume *looks* fine. But the raw bytes of the fragment are never removed from disk, and the hot append path in `_persist()` calls `appendFileSync()` directly onto the existing file:

```ts
} else {
	mkdirSync(dirname(this.sessionFile), { recursive: true });
	appendFileSync(this.sessionFile, `${JSON.stringify(entry)}
`);
	...
}
```

The next real entry lands right after the fragment, producing one combined malformed line that also fails to parse — so the new entry is silently lost too, and this repeats on every future crash/resume cycle at that same file. Confirmed with a direct repro: write a valid header + message + a truncated trailing record (no closing brace, as a crash mid-`fwrite` would leave it), open the session, append a new message, reload — the new message is gone.

## Fix

All in `packages/coding-agent/src/core/session-manager.ts`:

- The buffer/stream parsers now also track whether the file's **final** non-blank line existed but failed to parse, reporting it separately from ordinary mid-file skips. Only the trailing case is a safe, unambiguous repair target — an earlier malformed line is followed by content that already round-tripped, so it isn't attributable to an in-flight write and stays silently skipped exactly as before (out of scope, same as the row-validation done for #709, per the original report's own note).
- New `loadEntriesFromFileWithTail` / `loadEntriesFromFileAsyncWithTail` expose `{ entries, incompleteTail }`; the existing `loadEntriesFromFile` / `loadEntriesFromFileAsync` keep their original `FileEntry[]` signature as thin wrappers, so the one other call site (`context-tree.ts`) is unaffected.
- `SessionManager.setSessionFile` (used by both the sync `SessionManager.open` and, via a new `preloadedIncompleteTail` parameter, the async `SessionManager.openAsync`) now calls a new `_repairIncompleteTail()` after loading: it quarantines the fragment's raw bytes to a `<file>.incomplete-tail-<ts>-<id>.jsonl` sidecar for forensic recovery, logs a warning, and immediately rewrites the session file through the existing atomic temp-file+rename path (`_rewriteFile`) — so the corrupt bytes are gone from disk before any subsequent append can concatenate onto them, not just before the next write happens to occur.

## Acceptance criteria from #928 — all covered by new tests

- Resume distinguishes a complete final record from an incomplete tail.
- Invalid tail bytes are quarantined before append (and truncated immediately, not deferred).
- A subsequent entry round-trips.
- Earlier valid records are preserved.
- Recovery is observable (a log line via `getLogger("coding-agent.session-manager")`, plus the quarantine sidecar).

## Tests

Added 6 tests to `packages/coding-agent/test/session-manager/file-operations.test.ts`, all verified to fail against pre-fix `main` and pass after:
- `loadEntriesFromFileWithTail`/`-AsyncWithTail` report a truncated final record without dropping earlier entries.
- No tail is reported when the last record is valid but simply unterminated (no trailing newline) — not a corruption case.
- No tail is reported for a malformed line that isn't the last one (unchanged legacy skip behavior, intentionally out of scope).
- `SessionManager.open` and `SessionManager.openAsync` both: quarantine the fragment, keep earlier records, physically truncate the on-disk file immediately, and a subsequently appended entry round-trips through a reload instead of being lost.

Also ran the full session-manager-adjacent suite for regressions: `test/session-manager/`, `session-manager-flush.test.ts`, `session-manager-git-state.test.ts`, `sdk-session-manager.test.ts`, `context-tree.test.ts` — **161 + 14 passed, 0 failed**. `npx tsgo -p tsconfig.json --noEmit` and the root `npm run check` (biome, tsgo, installer render, browser smoke) both pass. Added a `packages/coding-agent/.changes/` fragment per the changelog-fragment CI check.

### DIFF
--- a/packages/coding-agent/.changes/eng-session-jsonl-incomplete-tail-repair.md
+++ b/packages/coding-agent/.changes/eng-session-jsonl-incomplete-tail-repair.md
@@ -1,0 +1,1 @@
+- Fixed a crash-truncated session record silently swallowing the next message too by repairing an incomplete trailing JSONL record on resume instead of leaving it on disk for the next append to concatenate onto.
--- a/packages/coding-agent/src/core/session-manager.ts
+++ b/packages/coding-agent/src/core/session-manager.ts
@@ -1,5 +1,13 @@
 import type { AgentMessage } from "@earendil-works/pi-agent-core";
-import type { AssistantMessage, ImageContent, Message, ServiceTier, TextContent, Usage } from "@earendil-works/pi-ai";
+import {
+	type AssistantMessage,
+	getLogger,
+	type ImageContent,
+	type Message,
+	type ServiceTier,
+	type TextContent,
+	type Usage,
+} from "@earendil-works/pi-ai";
 import { randomUUID } from "crypto";
 import {
 	appendFileSync,
@@ -29,6 +37,8 @@ import {
 	createCustomMessage,
 } from "./messages.js";
 import { cloneUsage } from "./usage.js";
+
+const log = getLogger("coding-agent.session-manager");
 
 export const CURRENT_SESSION_VERSION = 3;
 const SESSION_LIST_SEARCH_TEXT_MAX_CHARS = 64 * 1024;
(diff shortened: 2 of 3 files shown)


kaluli123123:
Understood re: the vouching policy. Filed here for the record and cross-linked from the corresponding Discussion (#1764) so a maintainer can pick up the diff directly if useful -- reproduction, fix, and regression test are documented there and in this PR's description. Happy to re-open or resubmit from a vouched account if that's ever preferred.

PR #1787, opened 2026-08-26
fix(runtime): isolate %%bash/%%script cell subprocesses into their own process group and escalate interrupts against it

### LINKED ISSUE #849: Esc/interrupt leaves `%%bash` subprocesses running forever: forceAbort resolves the tool call after 1s but never reaches the cell's grandchildren

kaluli123123:
Fixes #849.

Posted for review after discussion in #1786 (per CONTRIBUTING.md's contribution process).

## Summary

Interrupting (Esc) an `ipython` cell running a `%%bash` command returns control to the model, but the shell subprocess (and its own children, e.g. a backgrounded `grep`) keep running indefinitely at full CPU. The only recovery is killing the leaf process by hand or choosing "Kill kernel and restart", which loses the whole Python namespace. This is #849, one of the "related reports" under the still-open tracker #1382 (*Complete queued, interrupted, and archived session lifecycle recovery*).

## Root cause

`IPython.core.magics.script.ScriptMagics.shebang` (a third-party, pip-installed dependency, not part of this repo) spawns the cell's shell via `asyncio.create_subprocess_exec` with **no process-group isolation**, so the shell (and anything it execs/forks) shares the kernel process's own OS process group. Its own `KeyboardInterrupt -> SIGINT/terminate/kill` escalation only ever signals the direct child, never that child's own children — and in practice never even runs, since it blocks IPython's dedicated worker thread while the SIGINT-raised `KeyboardInterrupt` is delivered to the *main* thread. `kernel/index.ts`'s `onAbort()`/`interrupt()` only sends a Jupyter `interrupt_request`, which ipykernel turns into `SIGINT` to the kernel's own pid — it never reaches a subprocess either.

Confirmed directly against the real installed kernel-venv (IPython 9.16.1):
- `asyncio.create_subprocess_exec(..., start_new_session=True)` isolates the spawned shell into its own process group.
- A backgrounded child it forks (e.g. `sleep 30 &`) inherits that same group even after being reparented to init.
- `os.killpg` on that group reaches the whole tree while leaving the parent Python process completely untouched.
- A non-interactive bash sets `SIGINT` to be **ignored** for any `&`-backgrounded child (standard POSIX/bash behavior, confirmed empirically) — so an escalation that stops at SIGINT is not sufficient; it must actually reach SIGTERM/SIGKILL.

## Fix

Entirely inside `prime-agent-runtime` — **no TypeScript changes**, since the host's existing `onAbort()`/`interrupt()` already delivers the triggering SIGINT to the kernel's own pid today:

- `prime-agent-runtime/src/rlm/_script_process_group.py` (new): scopes a process-group-isolating `asyncio.create_subprocess_exec` proxy to `IPython.core.magics.script`'s own module namespace only (no other code in the process is affected). Tracks every subprocess spawned that way in a small, self-pruning registry — pruned by **process-group** liveness (`os.killpg(pgid, 0)`), not the original spawned pid's own liveness, since a non-interactive bash exits on EOF well before a backgrounded child it started finishes; checking the leader's pid alone would wrongly conclude the whole group is already gone and skip escalating, leaking exactly the child this exists to stop. Installs a `SIGINT` handler (chaining to whatever was previously installed, e.g. ipykernel's own, so the kernel's own interrupt semantics are unchanged) that signals each tracked group with SIGINT immediately, then escalates to SIGTERM and SIGKILL on a background thread if the group is still alive. An explicitly backgrounded script (`%%bash --bg`, tracked in IPython's own `ScriptMagics.bg_processes`) is excluded, since it's expected to outlive its cell.
- `prime-agent-runtime/src/rlm/__init__.py`: installs the patch unconditionally at import time (`rlm` is already imported for every kernel's bootstrap readiness check), idempotently.

A design that instead put the **kernel itself** in its own process group and group-signaled it on the force-abort timeout was considered and rejected: escalating to SIGKILL on that group would also kill the kernel process on every interrupt-timeout, destroying the user's Python namespace automatically instead of leaving that as the existing, explicit manual "Kill kernel and restart" fallback — a much larger behavior change than this bug calls for.

## Tests

Added `prime-agent-runtime/test/test_script_process_group.py` (4 new tests, all verified to fail against the pre-fix code — either via `ImportError` with the module absent, or the underlying unpatched behavior — and pass after the fix):
- the patched `create_subprocess_exec` isolates a new process group
- SIGINT escalates to kill the whole group (a backgrounded grandchild included) without touching the test process (standing in for the kernel) itself
- an explicitly backgrounded (`%%bash --bg`) subprocess survives an unrelated later interrupt
- a completed subprocess is pruned from the tracked registry

## Validation

```
cd prime-agent-runtime && uv run python -m unittest discover -s test   # exact CI command
# 102 passed (98 existing + 4 new), 0 failed
```

Re-verified RED: with `_script_process_group.py` absent and `__init__.py` reverted, the same command fails with `ImportError`. Root `npm run check` (biome, `tsgo --noEmit`, installer render, browser smoke) passes; this change touches no `packages/*` files, so no changelog fragment is required by the changelog-fragment CI check.

### DIFF
--- a/prime-agent-runtime/src/rlm/__init__.py
+++ b/prime-agent-runtime/src/rlm/__init__.py
@@ -21,6 +21,14 @@ try:
 except Exception:  # pragma: no cover - only available in kernels
     get_ipython = None  # type: ignore[assignment]
 
+from . import _script_process_group as _script_process_group
+
+# Give every %%bash/%%script cell's subprocess tree its own OS process group
+# and an escalating SIGINT/SIGTERM/SIGKILL interrupt path, without ever
+# touching the kernel process itself. Installed unconditionally so it is live
+# before any user cell (including the very first one) can run. See #849.
+_script_process_group.install()
+
 HOST_COMM_TARGET = "host.request"
 
 
--- a/prime-agent-runtime/src/rlm/_script_process_group.py
+++ b/prime-agent-runtime/src/rlm/_script_process_group.py
@@ -1,0 +1,216 @@
+"""Process-group isolation and interrupt escalation for %%bash/%%script cells.
+
+IPython's ``ScriptMagics.shebang`` (``IPython/core/magics/script.py``, a
+third-party, pip-installed dependency -- not part of this repo) spawns a
+``%%bash``/``%%script`` cell's shell via ``asyncio.create_subprocess_exec``
+with no process-group isolation, so the shell (and anything it execs or
+forks, e.g. a backgrounded ``grep``) shares this kernel process's own OS
+process group. Interrupting the cell (Esc in Prime Agent) only ever reaches
+the kernel's own pid -- via ipykernel's SIGINT delivery for the host's
+``interrupt_request`` -- never the shell's own children, so a stuck or
+backgrounded child leaks and keeps running at full CPU indefinitely (see
+prime-agent#849).
+
+This module, imported unconditionally from :mod:`rlm` at kernel bootstrap:
+
+1. Scopes a process-group-isolating ``asyncio.create_subprocess_exec`` proxy
+   to ``IPython.core.magics.script``'s own module namespace only (no other
+   code in the process is affected).
+2. Tracks every subprocess spawned that way in a small, self-pruning
+   registry.
+3. Installs a ``SIGINT`` handler (chaining to whatever was previously
+   installed, e.g. ipykernel's own) that escalates SIGINT -> SIGTERM ->
+   SIGKILL against each tracked subprocess's *process group* -- never the
+   kernel process itself, which is never made a member of any of these
+   groups and is therefore untouched.
+
+An explicitly backgrounded script (``%%bash --bg``, tracked in IPython's own
+``ScriptMagics.bg_processes``) is excluded: it is expected to outlive its
+cell, so an unrelated later interrupt must not sweep it.
+"""
+
+from __future__ import annotations
+
+import os
+import signal
+import threading
+from typing import Any, Callable
+
+try:
+    from IPython.core.magics import script as _ipython_script_magics
+except Exception:  # pragma: no cover - only available inside a real kernel
+    _ipython_script_magics = None  # type: ignore[assignment]
+
+# Escalation timeline. Kept as module attributes (not local constants) so
+# tests can shrink them instead of sleeping through production-sized delays.
+SIGTERM_DELAY_SECONDS = 1.0
+SIGKILL_DELAY_SECONDS = 0.5
+
+_INSTALLED_MARKER = "_prime_agent_script_process_group_patch"
+
+# pid -> pgid, for every process the proxy below has spawned that has not yet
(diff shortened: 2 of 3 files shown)


kaluli123123:
Understood re: the vouching policy. Filed here for the record and cross-linked from the corresponding Discussion (#1786) so a maintainer can pick up the diff directly if useful -- reproduction, fix, and regression test are documented there and in this PR's description. Happy to re-open or resubmit from a vouched account if that's ever preferred.

PR #1789, opened 2026-08-26
fix(coding-agent): coalesce chatty RLM child usage attribution writes

### LINKED ISSUE #1054: Child usage attribution flood freezes session worker: 550+ child_usage_attributed entries in 20 min with active RLM subagents; attach/heartbeats time out, session unreachable (v0.7.1)
## Summary

With active RLM subagents, **every** child assistant `message_end` appends a `child_usage_attributed` entry to the parent session's transcript. With 4 fast-working subagents this becomes a firehose: **553 attribution entries in 20 minutes** (peak 73/min), each one a full ~600-byte transcript entry that mutates the parent's assistant-message usage and advances the transcript leaf. The single worker process hosting the parent + children got saturated (**97–99% CPU, 3.1 GB RSS for ~30 min**), so the supervisor's `attach` / `heartbeats_list` commands timed out for 25+ minutes and the session became completely unreachable from the UI ("session stuck"). Only `SIGKILL` of the worker unblocked it; the supervisor then auto-recovered the session intact (transcripts preserved).

## Environment

| | |
|---|---|
| prime-agent | 0.7.1 (npm global install) |
| OS | CachyOS Linux (Arch-based), kernel 7.1.4-1-cachyos |
| Node | v26.4.0 |
| Model | `opencode-go` / `deepseek-v4-flash` (root session; subagents inherit) |
| Hardware | 16 cores, 14 GB RAM |
| Topology | 1 root session + 4 RLM subagents (`rlmDepth` 1, `rlmMaxDepth` 2), all in one daemon worker process |

## Steps to reproduce

1. Start a root session (we used `/goal ...` with a large orchestration task).
2. Spawn 4 RLM subagents (`await rlm('...', name='...')`) that perform rapid, tool-call-heavy work (editing files, running builds/tests).
3. Parent ends its turn (per the recommended non-blocking pattern).
4. Wait ~10–20 minutes while the children work.

Observed result: the parent worker saturates a core, the transcript floods with `child_usage_attributed` entries, and the session can no longer be attached from the Agents View / CLI (`Timed out waiting for daemon worker response to attach`).

## Expected behavior

- Usage attribution should be cheap and invisible: coalesced/batched, never starving the event loop.
- The session should remain responsive (attach, prompt, heartbeat) while subagents work, no matter how chatty they are.

## Actual behavior / evidence

**Attribution flood in the parent transcript** (`session 019fe503-0f3e-71ff-8244-2c86f7b71dec`):

```
total child_usage_attributed events : 553   (all origin: "spawn_task")
window                             : 2026-08-09T05:43:27Z → 06:03:31Z  (20.1 min)
avg rate                           : 27.5/min
peak minute                        : 73 events/min (05:44Z)
distinct targetIds                 : 2  (af9da882 ×408, abe4513d ×145)
parentId == previous event id      : 506/553 (92%)
transcript bytes from these lines  : ~340 KB of 1.7 MB (20%)
parent's own user/assistant msgs in window: 0  (last assistant msg 05:43:50Z, right after spawning)
```

The events correlate **1:1 with real child activity** — during the same window the 4 children produced 551 assistant messages (145+147+139+120), so this is not a single event re-triggering itself; it is per-message persistence with no batching/coalescing.

**Worker unresponsive** (`ps`):

```
242761  97.1% CPU   21.0% MEM (3.1 GB)   29:31 elapsed   prime-agent   (worker hosting the session tree)
```

**Supervisor control commands time out** (daemon log):

```
05:57:12  Supervisor command attach failed: Error: Timed out waiting for daemon worker response to attach
05:57:50  Supervisor command attach failed: Error: Timed out waiting for daemon worker response to attach
05:59:33  Supervisor command attach failed: Error: Timed out waiting for daemon worker response to attach
05:59:53  Could not list heartbeats from a worker: Timed out waiting for daemon worker response to heartbeats_list
```

**Recovery worked well once the worker was killed** (this part is positive):

```
06:03:33  Recovered worker 8f831993adb5 without replaying uncertain operations:
          rlm_child_update, tool_execution_start, message_start, turn_start, turn_start
06:03:35  Session worker 8f831993adb5 ... listening ... (new pid)
```

After recovery the session resumed healthy (`taskState: needs_input`, normal ~25 s heartbeats, CPU back to ~4%). All session and subagent transcripts were intact on disk.

## Code path (v0.7.1, `dist/bundle/chunk-VNU2AJHD.js`)

Child-event handler in `AgentSession` — fires on **every** child assistant `message_end`:

```js
} else if (event.type === "message_end" && event.message.role === "assistant") {
  const assistant = event.message;
  if (assistant.stopReason !== "error" && assistant.stopReason !== "aborted") {
    attributeChildUsage(parentAssistantForUsage?.usage ?? emptyUsage(), assistant.usage);
    if (parentAssistantForUsage) {
      const parentEntry = this._findAssistantEntryForMessage(parentAssistantForUsage); // linear scan
      if (parentEntry) {
        // origin computed from preceding prompt ...
        this.sessionManager.appendChildUsageAttribution(parentEntry.id, assistant.usage,
                                                        parentAssistantForUsage.usage, origin);
      }
    }
  }
  ...
}
```

`appendChildUsageAttribution` — writes a full entry, **mutates an existing assistant entry**, and parents the new entry to the current leaf (not to the target):

```js
appendChildUsageAttribution(targetId, childUsage, aggregateUsage, origin) {
  const target = this.byId.get(targetId);
  if (target?.type !== "message" || target.message.role !== "assistant") {
    throw new Error(`Assistant message entry ${targetId} not found`);
  }
  target.message.usage = cloneUsage2(aggregateUsage);          // mutates existing entry
  const entry = {
    type: "child_usage_attributed",
    id: generateId(this.byId),
    parentId: this.leafId,                                     // chains onto previous event
    timestamp: new Date().toISOString(),
    targetId,
    childUsage: cloneUsage2(childUsage),
    aggregateUsage: cloneUsage2(aggregateUsage),
    ...origin ? { origin } : {}
  };
  this._appendEntry(entry);                                    // advances leaf to this entry
  return entry.id;
}
```

Consequences visible in our transcript:
- `parentId: this.leafId` ⇒ 92% of events are chained onto the previous attribution event (the chain never attaches to a message), and while children are active the transcript **leaf is perpetually a bookkeeping entry**, never an assistant message.
- Every event also mutates the parent's spawn-message usage (`target.message.usage`), which invalidates/updates anything derived from that entry.
- There is no rate limiting, no coalescing window, and no "only persist if the aggregate changed" check anywhere in this path.

## Root-cause hypotheses (for maintainers)

1. **Firehose + shared event loop**: per-child-message persistence with no coalescing, running in the same single-threaded worker that must also answer supervisor control commands. Under sustained child activity the loop is starved; with a fixed 20 s timeout and no escalation, `attach`/`heartbeats_list` fail forever and the session is unrecoverable from the UI.
2. **Bookkeeping entries advance the transcript leaf** (`parentId: this.leafId`), which keeps the parent "working" state ambiguous and may interact badly with context/leaf-based logic while children are active.
3. **In-place mutation of existing entries** from the child path may trigger session-update subscribers / context rebuilds per event (contributing to the CPU saturation).
4. **Memory**: RSS grew to 3.1 GB in ~30 min for a ~1.7 MB transcript — worth investigating whether usage clones/stream buffers are retained.

## Suggested fixes

- **Coalesce attribution**: keep the per-target aggregate in memory and persist a `child_usage_attributed` entry at most once per N seconds per target, on child session end, or when the parent turn ends — not once per child message.
- **Don't advance the leaf for bookkeeping**: set `parentId` to the target assistant entry (or `null`) instead of `this.leafId`, so the leaf always settles on a real message.
- **Flood watchdog**: if attribution entries exceed a threshold per minute with no accompanying user/assistant messages, stop per-event persistence, log once, and/or auto-detach.
- **Escalate control-channel timeouts**: on repeated `attach`/`heartbeats_list` timeouts, automatically kill+recover the worker (the existing recovery path proved fast and safe: ~2 s, transcripts intact, no replay of uncertain ops).
- Investigate the RSS growth under this load.

## Impact

- Session unreachable from the UI for 25+ minutes; repeated `attach` timeouts in the client.
- ~1.7 MB transcript, 20% of it bookkeeping noise.
- No data loss in our case: subagent work on disk and all transcripts survived the worker kill + auto-recovery.

---

*Sample event (abridged):*

```json
{"type":"child_usage_attributed","id":"...","parentId":"<previous attribution event id>",
 "timestamp":"2026-08-09T05:44:xxZ","targetId":"af9da882",
 "childUsage":{"input":...,"output":...},"aggregateUsage":{"input":...,"totalTokens":138820,...},
 "origin":"spawn_task"}
```

kaluli123123:
Fixes #1054.

Posted for review after discussion in #1788 (per CONTRIBUTING.md's contribution process). Credits @Lauritz-Timm's prior #1060/#1061 upfront -- see the Discussion for the comparison.

## Summary

With active RLM subagents, every child assistant `message_end` appends a `child_usage_attributed` entry to the parent session's transcript, synchronously and unconditionally. With a few fast-working subagents this becomes a firehose: one report measured 553 attribution entries in 20 minutes (peak 73/min), each a full ~600-byte transcript entry that also mutates the parent's assistant-message usage object and advances the transcript leaf/index on every single call. The worker process hosting the parent + children saturated (97-99% CPU, 3.1 GB RSS observed), so the supervisor's `attach`/`heartbeats_list` commands timed out for 25+ minutes and the session became completely unreachable from the UI. This is #1054, one of the "related reports" under the still-open tracker #1382.

**Prior art, credited upfront:** @Lauritz-Timm already diagnosed this and proposed a fix in #1060/#1061 (both closed, not merged, via the maintainer's standard backlog-sweep note citing #1162 as "covering" this — but the underlying bug is still live on current `main`, confirmed with a new regression test below). Their approach and mine converge on the same root cause and a similar shape (accumulate pending usage, flush at a natural boundary, force-flush on cleanup), which isn't surprising given there's really one place in the code this can be fixed. The concrete difference: #1060/#1061 flush on `agent_end` and whenever the attribution `origin` changes; this PR instead debounces on a fixed time window (`RLM_CHILD_USAGE_ATTRIBUTION_COALESCE_MS`, 1s) regardless of origin, plus the same forced flush on run settlement. Either is a reasonable design; I'm not claiming this supersedes their work, just documenting an independent pass with a different coalescing trigger, in case it's a useful alternative or point of comparison for whoever picks this up.

## Root cause

The RLM child event subscriber in `spawnRlmChild` (`agent-session.ts`) calls `attributeChildUsage()` and `sessionManager.appendChildUsageAttribution()` directly and synchronously inside the `"message_end"` handler, once per child assistant message, with no batching of any kind.

## Fix

Track a per-run pending usage delta (summed via the existing `addAssistantUsage`, not just "keep the latest" — each message's usage is an independent, additive cost, so skipping intermediate messages would undercount) and debounce the actual attribute+persist call to at most once per `RLM_CHILD_USAGE_ATTRIBUTION_COALESCE_MS` (1000ms) of inactivity for that child. The run's existing `finally` block also force-flushes any still-pending delta before the child run settles, so a run that finishes inside the coalescing window never silently drops its last batch of usage. This lives entirely at the call site; `SessionManager.appendChildUsageAttribution`'s own contract (synchronous JSONL append) is unchanged, since other tests (`context-tree.test.ts`) rely on it persisting immediately when called.

## Tests

`packages/coding-agent/test/agent-session-recursion.test.ts`:
- Updated the existing "attributes every tool-loop turn..." test (previously asserted exactly 2 separate attribution entries for a 2-turn tool loop): now asserts exactly 1 coalesced entry with the correct combined usage.
- Added "persists far fewer entries than turns for a chatty child...": 20 rapid tool-loop turns produce far fewer than 20 attribution entries (< 10), while the sum of every persisted entry's `childUsage` still equals the exact total across all 20 turns — directly modeling the reported ~550-messages-to-~550-entries shape, now collapsed (21 entries for 20 turns on the pre-fix code, matching the bug's ratio almost exactly).

Both verified to fail against the pre-fix code and pass after.

## Validation

```
cd packages/coding-agent && npx tsx ../../node_modules/vitest/dist/cli.js --run \
  test/agent-session-recursion.test.ts test/context-tree.test.ts \
  test/suite/agent-session-queue.test.ts test/suite/agent-session-autonomous.test.ts \
  test/suite/agent-session-goal.test.ts
# 301 passed. 3 pre-existing, unrelated failures (rlmMaxDepth env-var pollution)
# reproduce identically on unmodified main.
```

`npx tsgo -p tsconfig.json --noEmit` and the root `npm run check` (biome, tsgo, installer render, browser smoke) both pass. Added a `packages/coding-agent/.changes/` fragment per the changelog-fragment CI check.

### DIFF
--- a/packages/coding-agent/.changes/eng-coalesce-child-usage-attribution.md
+++ b/packages/coding-agent/.changes/eng-coalesce-child-usage-attribution.md
@@ -1,0 +1,1 @@
+- Fixed chatty RLM subagents saturating the worker (high CPU, stuck attach/heartbeats) by coalescing rapid-fire child usage attribution writes into a bounded number of persisted entries instead of one per child message.
--- a/packages/coding-agent/src/core/agent-session.ts
+++ b/packages/coding-agent/src/core/agent-session.ts
@@ -247,7 +247,13 @@ import {
 	transitionSessionAction,
 	type WakePolicy,
 } from "./session-action-store.js";
-import type { BranchSummaryEntry, CompactionEntry, SessionContext, SessionMessageEntry } from "./session-manager.js";
+import type {
+	BranchSummaryEntry,
+	ChildUsageAttributionEntry,
+	CompactionEntry,
+	SessionContext,
+	SessionMessageEntry,
+} from "./session-manager.js";
 import {
 	CURRENT_SESSION_VERSION,
 	getLatestCompactionEntry,
@@ -917,6 +923,10 @@ interface RlmSubagentModelSelection {
 
 const KERNEL_STATE_LISTING_TIMEOUT_MS = 5000;
 const RLM_MAX_DEPTH_STATE_CUSTOM_TYPE = "rlm_max_depth_state";
+// A chatty child can emit tens of assistant messages per minute; persisting a full
+// child_usage_attributed JSONL entry per message saturates the worker (#1054).
+// Coalesce rapid-fire attributions to the same parent entry into one flush per window.
+const RLM_CHILD_USAGE_ATTRIBUTION_COALESCE_MS = 1000;
 
 function noopRlmChildAbort(): void {}
 function noopRlmChildEventUnsubscribe(): void {}
@@ -10246,6 +10256,39 @@ export class AgentSession {
 		const label = rlmChildLabel(prompt);
 		let answerPreview: string | undefined;
 		let durationMs: number | undefined;
+		// Coalesce this child's rapid-fire usage attributions into at most one
+		// persisted child_usage_attributed entry per RLM_CHILD_USAGE_ATTRIBUTION_COALESCE_MS
+		// window instead of one JSONL append per assistant message (#1054).
+		let pendingChildUsageDelta: Usage | undefined;
+		let pendingChildUsageOrigin: ChildUsageAttributionEntry["origin"] | undefined;
+		let childUsageAttributionFlushTimer: ReturnType<typeof setTimeout> | undefined;
+		const flushChildUsageAttribution = () => {
+			if (childUsageAttributionFlushTimer) {
+				clearTimeout(childUsageAttributionFlushTimer);
+				childUsageAttributionFlushTimer = undefined;
+			}
+			const delta = pendingChildUsageDelta;
+			const origin = pendingChildUsageOrigin;
+			pendingChildUsageDelta = undefined;
+			pendingChildUsageOrigin = undefined;
+			if (!delta || !parentAssistantForUsage) return;
+			const parentEntry = this._findAssistantEntryForMessage(parentAssistantForUsage);
+			if (!parentEntry) return;
+			attributeChildUsage(parentAssistantForUsage.usage, delta);
+			this.sessionManager.appendChildUsageAttribution(parentEntry.id, delta, parentAssistantForUsage.usage, origin);
+		};
+		const scheduleChildUsageAttribution = (usage: Usage, origin: ChildUsageAttributionEntry["origin"]) => {
+			if (!pendingChildUsageDelta) pendingChildUsageDelta = emptyUsage();
+			addAssistantUsage(pendingChildUsageDelta, usage);
+			pendingChildUsageOrigin = origin;
+			if (!childUsageAttributionFlushTimer) {
+				childUsageAttributionFlushTimer = setTimeout(
+					flushChildUsageAttribution,
+					RLM_CHILD_USAGE_ATTRIBUTION_COALESCE_MS,
+				);
+				childUsageAttributionFlushTimer.unref?.();
+			}
+		};
 		let toolUseCount = 0;
 		let runningToolCount = 0;
 		let activity: RlmChildAgentActivity | undefined;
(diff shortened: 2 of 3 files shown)


kaluli123123:
Understood re: the vouching policy. Filed here for the record and cross-linked from the corresponding Discussion (#1788) so a maintainer can pick up the diff directly if useful -- reproduction, fix, and regression test are documented there and in this PR's description, along with an upfront credit to @Lauritz-Timm's prior #1060/#1061. Happy to re-open or resubmit from a vouched account if that's ever preferred.

Issue #1072, opened 2026-08-09
Daemon leaves sessions stuck (idle-eviction failures, stuck-in-closing, stuck-in-executing) — v0.7.1


leolebkuchen:
## Summary

The Prime Agent daemon periodically leaves sessions stuck in an unrecoverable state — either they never exit "working" after a tool call has visibly already finished, or reattaching errors out with "Active session is closing". This has happened repeatedly over one working day (v0.7.1), roughly hourly, and previously worked fine on the same setup.

## Environment

- Prime Agent version: `0.7.1` (confirmed already latest via `prime-agent update`)
- OS: macOS (Darwin)
- Model/provider: local `toshllm` (llama.cpp OpenAI-compatible endpoint), but the daemon-level symptoms below look provider-independent

## Symptom 1: Idle eviction sweep failures leaking worker processes

`~/.prime/agent/logs/daemon.sock.*.log` shows this recurring roughly hourly across a full day:

```
supervisor: Idle eviction sweep failed: Error: Timed out draining daemon mutations for idle eviction
```

Each occurrence leaves an orphaned worker + ipykernel process pair running with no corresponding session in `prime-agent list` — confirmed via `ps aux | grep ipykernel` showing kernels with no matching tracked session, oldest observed at 4.5h idle.

## Symptom 2: Reattach fails with "session is closing"

```
supervisor: Session worker <id> stderr: daemon command "attach" failed: Error: Active session <session-id> is closing
```

The underlying worker process was found stuck in OS state `T` (suspended) on the same tty, not actually exiting. Killing the stuck PID directly resolved it; the session's `.jsonl` history under `~/.prime/agent/sessions/` was intact and resumed cleanly via `prime-agent -r <session-id>`.

## Symptom 3: Session stays "Executing" after its tool call already finished

Most recently: a session showed "Executing" in the UI for 10+ minutes running a trivial `lsof -t -i :3001` check. `ps aux` showed no `lsof` process running at all (i.e. it had already completed), and the llama-server backend showed `is_processing: false` on both slots — so neither the shell command nor the model was actually doing anything. `prime-agent stop <session-id>` (this time responding immediately, no force-kill needed) followed by `prime-agent -r <session-id>` recovered it with all prior messages intact.

## Common thread

[...]

PR #1796, opened 2026-08-26
fix(coding-agent): recover a session close stuck on an OS-stopped kernel

kaluli123123:
Fixes part of #1072 (see Scope note below -- this is a partial fix, not a full fix for all three reported symptoms).

Posted for review after discussion in #1795 (per CONTRIBUTING.md's contribution process).

## Scope note

This addresses one of #1072's three reported symptoms — "Active session `<id>` is closing" persisting forever after reattach fails, with the underlying worker process observed stuck in OS state `T` (job-control-stopped, e.g. an external SIGSTOP). The other two reported symptoms (idle-eviction-sweep timeouts leaking worker processes, and a session showing "Executing" after its tool call already finished) share the same general shape — something the daemon awaits assumes will always eventually settle, and a genuinely frozen (not merely slow) peer process breaks that assumption — but are separate leak points in different code paths and are **not** covered by this change. This is one focused sub-fix, not a full fix for #1072, and I want to be upfront about that rather than overclaim.

## Root cause

`closeSession()` (`daemon-mode.ts`) sets `closingSessions.set(activeSessionId, close)` before awaiting `closePromise` (→ `closeSessionOnce`, which awaits `state.runtime.session.abort()` and `state.runtime.dispose()`), and only clears that bookkeeping in `closePromise`'s own `finally` block. If the session's kernel process becomes OS-level job-control-stopped, it can no longer respond to anything short of `SIGCONT`/`SIGKILL`, so whatever `closeSessionOnce` is awaiting on it never settles, the `finally` block never runs, and `closingSessions` never clears — every future attach or close for that session fails with "is closing" until the whole worker/daemon is killed and the OS-level state is gone.

## Why not just add a timeout

This project deliberately does **not** use blanket timeouts for RPCs/waits that can legitimately run a long time — see PR #1701, *"remove blanket RPC timeouts, reject on real boundaries."* A naive fixed-duration timeout here would just as wrongly treat a slow-but-healthy close as failed. The fix instead checks the kernel process's actual OS state directly, so it only ever acts on a *confirmed-frozen* peer, never a merely-busy one.

## Fix

- `utils/child-process.ts`: new `isStoppedProcess(pid)`, reusing the existing `/proc`-stat-then-`ps`-fallback pattern already used by `isZombieProcess`, checking for POSIX state `T` specifically.
- `core/kernel/index.ts` / `core/tools/ipython.ts` / `core/agent-session.ts`: a new `kernelPid`/`kernelProcessPid` getter chain (`KernelManager` → `IpythonKernelProvisioner`, already held privately by `AgentSession` → a public `AgentSession.kernelProcessPid`) exposes the running kernel's OS pid for this kind of liveness check, without adding any new public surface to the Jupyter-protocol-facing parts of the kernel.
- `modes/daemon/daemon-mode.ts`: `closeSession()` now starts a `watchForStoppedKernelDuringClose()` watchdog alongside the close. It polls `isStoppedProcess()` on the session's kernel pid every 2s while the close is still pending; on the first confirmed-stopped poll it sends one `SIGCONT` (a session that was merely, legitimately suspended resumes and the close proceeds normally); if it is still confirmed stopped after further polls despite that, it releases the `closingSessions` entry (matching the existing entry by identity, the same check `closeSession`'s own cleanup already uses, so a race between the two is harmless) so the session becomes reattachable again. The original close keeps running in the background and settles normally whenever/if the kernel resumes on its own.

## Tests

- `test/child-process.test.ts`: `isStoppedProcess` against a **real** SIGSTOP/SIGCONT cycle on a real spawned subprocess (not mocked), confirming it is true only while genuinely stopped and distinguishes "stopped" from "dead"/"zombied"/"merely running".
- `test/suite/regressions/1072-stopped-kernel-close-recovery.test.ts` (new): drives the real `closeSession()` path (not the watchdog method in isolation) with a deliberately-hung `runtime.dispose()` and a mocked `isStoppedProcess`/`process.kill`, using fake timers for determinism: confirms `SIGCONT` is sent once, `closingSessions` is released only after the configured number of confirmations, and (second test) a merely-slow (never reported stopped) close is left completely alone.

Both verified to fail against pre-fix `daemon-mode.ts` and pass after the fix.

## Validation

```
cd packages/coding-agent && npx tsx ../../node_modules/vitest/dist/cli.js --run \
  test/suite/regressions/1072-stopped-kernel-close-recovery.test.ts test/child-process.test.ts \
  test/daemon-mode.test.ts test/suite/regressions/4257-update-restart-resume.test.ts
# 244 passed, 0 failed.
```

`npx tsgo -p tsconfig.json --noEmit` and the root `npm run check` (biome, tsgo, installer render, browser smoke) both pass. Added a `packages/coding-agent/.changes/` fragment per the changelog-fragment CI check.


kaluli123123:
Understood re: the vouching policy. Filed here for the record and cross-linked from the corresponding Discussion (#1795) so a maintainer can pick up the diff directly if useful. Scoped explicitly to one of #1072's three symptoms -- see the Discussion/PR description for what is and isn't covered. Happy to re-open or resubmit from a vouched account if that's ever preferred.

PR #1928, opened 2026-08-31
feat: refresh Prime Inference models from live API

sethkarten:
Part of [RES-1239](https://linear.app/primeintellect/issue/RES-1239/refresh-prime-agent-models-from-the-live-prime-inference-endpoint). Supersedes #1633.

## What changed

- Refreshes Prime Inference models directly from the public `https://api.pinference.ai/api/v1/models` endpoint.
- Uses an authenticated request to discover team-authorized `internal/*`, `dev/*`, and deployment routes.
- Applies live additions, removals, display names, pricing, context limits, modalities, and reasoning support without a Prime Agent release.
- Keeps models for all other providers in the release-bundled catalog.
- Loads a validated disk cache or the bundled Prime Inference list immediately, then refreshes the public list in the background.
- Preserves local `models.json` definitions, provider/model overrides, extension providers, and bundled client compatibility metadata.
- Removes the proposed R2 publication workflow, credentials, aggregate schema, and catalog URL configuration.

## Safety and fallback behavior

- Both public and authenticated requests are pinned to the Prime Inference HTTPS origin.
- Live data cannot inject headers or change the provider, API type, or request URL.
- Unknown models require complete `specs`; entries without them are skipped rather than assigned guessed limits or capabilities.
- Existing models with missing `specs` retain bundled capabilities and limits while accepting live names and prices.
- New models use conservative client compatibility defaults and do not send unconfirmed reasoning controls.
- A response must retain at least 50% of the bundled Prime Inference model IDs. Severe truncation, invalid responses, timeouts, and oversized responses fall back to the last valid cache or bundled models.
- Public responses exclude `internal/*`, `dev/*`, and deployment IDs containing `:`. Authenticated-only routes use a separate mode-`0600` cache keyed by a credential/team fingerprint; they never enter the public cache.
- `PI_OFFLINE=1` skips network refreshes.

## Current backend follow-up

The endpoint currently returns 120 public routes. It provides complete `specs` for 113. Five new Qwen routes and the currently authorized private routes cannot be added dynamically yet because `specs` is `null`. To remove the remaining client fallbacks and route-name heuristics, the endpoint should provide:

- complete `specs` for every public and private route;
- an explicit public/private visibility field;
- supported reasoning controls or a thinking-level map;
- an optional `featured` field if realtime promotion is desired.

## Validation

- `npm run check`
- AI package: 343 passed, 720 skipped
- Focused Prime Inference and registry tests: 123 passed
- Live generation: 1,290 models across 32 providers; 120 public Prime Inference routes
- Runtime live smoke test: 115 usable public Prime Inference models, five additions, zero unintended removals, zero private-route leaks, and zero transport/header changes

---

> [!NOTE]
> **Medium Risk**
> Runtime model pickers and context limits now follow live API data with merge/fallback rules; mistakes could hide models, leak private routes, or mis-state limits, though guards and tests target those cases.
> 
> **Overview**
> Prime Inference model lists now update from the live **`/models`** endpoint instead of relying only on release bundles and ad hoc generator logic.
> 
> **`@earendil-works/pi-ai`** adds a shared **`parsePrimeInferenceModelCatalog`** parser (pricing, cache costs, specs, private-route detection) and moves generated catalog output through **`renderModelsFile`** so remote strings are JSON-escaped. The **`generate-models`** script hits the public endpoint without Prime CLI credentials, applies truncation/coverage guards, and still merges OpenRouter gaps where needed.
> 
> **Prime Agent** loads a validated **`prime-inference-models-cache.json`** (or bundled models) immediately, background-refreshes the public catalog, and merges live entries via **`buildPrimeInferenceModels`** / **`mergePrimeInferenceModels`** while keeping bundled **`compat`** and thinking maps. Authorized **`internal/*`**, **`dev/*`**, and deployment routes are built from authenticated catalog responses and persisted as full model records, not ID sets. **`PI_OFFLINE=1`** skips network refresh; docs describe the new behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 24d3413c304cefbb063e86a7e41c85207916d90e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Refresh Prime Inference models from live API with validated cache fallback
> - Adds a shared catalog parser and builder for Prime Inference models, validating metadata like pricing, cache costs, context/output limits, and image support
> - The generator now fetches from the public `/models` endpoint without Prime CLI credentials, with a severe-truncation guard preserving the existing snapshot if live results are too sparse
> - The registry merges authorized private model objects with full metadata, excludes public IDs from private results, and reloads models after catalog or authorization changes
> - Public refreshes are backgrounded with cache fallback, request deduplication, atomic 0600 cache writes, and `PI_OFFLINE=1` support
> - Risk: `loadModels` in [model-registry.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/1928/files#diff-2fd8c4a2672b4ce27e73d494479df612c8b30ebac3dda5a160e2d861c612f453) replaces bundled Prime Inference entries with live entries when supplied; existing tests for fixed maxTokens and context assertions are relaxed to allow clamped live metadata
>
> 
> #### Changes since #1928 opened
>
> - Changed `ModelRegistry` private Prime authorization flow to cache and restore full model metadata instead of model IDs, enabling immediate availability of authorized private deployments at startup without network access [24d3413]
> - Tightened validation in `parsePrimeInferenceModelCatalog` utility to require both input and output modalities to be arrays of strings, treating modalities as undefined when validation fails and only attaching full specs when all required fields are valid [24d3413]
> - Changed minimum coverage criteria in `buildPrimeInferenceModels` utility to measure coverage of bundled models rather than total model count [24d3413]
> - Added test coverage for cached authorized deployment metadata restoration without network access, retention of priced entries without complete specs, and coverage measurement against bundled models [24d3413]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized e7e810c.</sup>
> 
>


packages/coding-agent/src/core/remote-model-catalog.ts
```diff
@@ -0,0 +1,262 @@
...
+	const response = await fetchFn(url, {
+		headers: { accept: "application/json" },
+		signal: AbortSignal.timeout(MODEL_CATALOG_FETCH_TIMEOUT_MS),
+	});
+	if (!response.ok) throw new Error(`Model catalog request failed with status ${response.status}`);
+	const contentLength = Number(response.headers.get("content-length"));
+	if (Number.isFinite(contentLength) && contentLength > MAX_MODEL_CATALOG_BYTES) {
+		throw new Error("Model catalog response is too large");
+	}
+	const text = await response.text();
```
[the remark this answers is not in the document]
sethkarten: Fixed in ee1538d4bc. The response body is now read incrementally and the stream is cancelled as soon as it exceeds 8 MiB. A chunked-response regression test verifies early cancellation.

packages/coding-agent/src/core/remote-model-catalog.ts
```diff
@@ -0,0 +1,262 @@
...
+}
+
+export function readCachedRemoteModelCatalog(cachePath: string): Model<Api>[] | undefined {
+	return readCache(cachePath, getRemoteModelCatalogUrl())?.catalog.models;
+}
+
+function offlineModeEnabled(): boolean {
+	const value = process.env.PI_OFFLINE?.toLowerCase();
+	return value === "1" || value === "true" || value === "yes";
+}
```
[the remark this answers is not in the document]
sethkarten: Fixed in aeb795a733. Model refresh, package management, tool management, and CLI flag handling now use the shared `isTruthyEnvFlag` helper.

packages/ai/scripts/model-catalog-format.ts
```diff
@@ -0,0 +1,83 @@
...
+	if (typeof value.generatedAt !== "string" || !Number.isFinite(Date.parse(value.generatedAt))) {
+		throw new Error("Model catalog has an invalid generatedAt timestamp");
+	}
+	if (!Array.isArray(value.models) || value.models.length === 0 || value.models.length > MAX_MODEL_CATALOG_MODELS) {
+		throw new Error("Model catalog has an invalid model count");
+	}
+
+	const seen = new Set<string>();
+	for (const model of value.models) {
+		if (!isCatalogModel(model)) throw new Error("Model catalog contains an invalid model");
```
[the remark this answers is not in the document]
sethkarten: Fixed in ee1538d4bc. Publication and runtime now call the same strict `parseModelCatalog` implementation from `@earendil-works/pi-ai`, with tests for the shared bounds.

packages/coding-agent/src/core/remote-model-catalog.ts
```diff
@@ -0,0 +1,174 @@
...
+		}
+	} finally {
+		reader.releaseLock();
+	}
+	const text = Buffer.concat(chunks, bytesRead).toString("utf8");
+	return parseRemoteModelCatalog(JSON.parse(text) as unknown);
+}
+
+export async function refreshRemoteModelCatalog(
+	cachePath: string,
```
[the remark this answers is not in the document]
sethkarten: Fixed in b299f64031. A non-empty hosted catalog that produces zero usable transport matches now falls back to cloned bundled models; mixed catalogs remain authoritative when at least one entry is accepted. Added a regression test.

packages/ai/src/model-catalog.ts
```diff
@@ -0,0 +1,109 @@
...
+	if (typeof value.generatedAt !== "string" || !Number.isFinite(Date.parse(value.generatedAt))) {
+		throw new Error("Invalid model catalog timestamp");
+	}
+	if (!Array.isArray(value.models) || value.models.length === 0 || value.models.length > MAX_MODEL_CATALOG_MODELS) {
+		throw new Error("Invalid model catalog model count");
+	}
+	const seen = new Set<string>();
+	for (const model of value.models) {
+		if (!isCatalogModel(model)) throw new Error("Invalid model catalog entry");
+		const key = `${model.provider}/${model.id}`;
```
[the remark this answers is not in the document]
sethkarten: Fixed in b299f64031. Duplicate detection and runtime exact matching now encode `[provider, id]` as a JSON pair, with a regression test covering slash-containing values.

packages/ai/src/model-catalog.ts
```diff
@@ -0,0 +1,109 @@
...
+		Number.isSafeInteger(value.contextWindow) &&
+		value.contextWindow > 0 &&
+		value.contextWindow <= 100_000_000 &&
+		typeof value.maxTokens === "number" &&
+		Number.isSafeInteger(value.maxTokens) &&
+		value.maxTokens > 0 &&
+		value.maxTokens <= 100_000_000 &&
+		(value.featured === undefined || typeof value.featured === "boolean") &&
+		(value.headers === undefined || isBoundedJsonObject(value.headers)) &&
+		(value.compat === undefined || isBoundedJsonObject(value.compat))
```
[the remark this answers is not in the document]
sethkarten: Fixed in b299f64031. Catalog compatibility metadata is now validated by API and exact field type, including nested OpenRouter and Vercel routing objects. Invalid cross-API and malformed nested values are covered by tests, and the live 1,245-model catalog passes the stricter validator.

packages/ai/scripts/generate-models.ts
```diff
@@ -809,6 +818,7 @@ async function fetchOpenRouterModels(): Promise<Model<any>[]> {
 		return models;
 	} catch (error) {
 		console.error("Failed to fetch OpenRouter models:", error);
+		if (STRICT_MODEL_CATALOG_REFRESH) throw error;
```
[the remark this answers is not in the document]
sethkarten: Fixed in e4e286108f. Strict generation now throws when the OpenRouter source produces zero tool-capable models, matching the existing models.dev and Vercel AI Gateway guards. Strict live generation and validation still produce 1,245 models.

packages/coding-agent/src/core/remote-model-catalog.ts
```diff
@@ -0,0 +1,178 @@
...
+): Model<Api>[] {
+	if (!remoteModels) return bundledModels.map((model) => structuredClone(model));
+	const exact = new Map(bundledModels.map((model) => [modelKey(model.provider, model.id), model]));
+	const transports = new Map<string, Model<Api>>();
+	for (const model of bundledModels) transports.set(`${model.provider}\0${model.api}\0${model.baseUrl}`, model);
+
+	const merged: Model<Api>[] = [];
+	for (const remote of remoteModels) {
+		const key = modelKey(remote.provider, remote.id);
+		const template = exact.get(key) ?? transports.get(`${remote.provider}\0${remote.api}\0${remote.baseUrl}`);
```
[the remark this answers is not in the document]
sethkarten: Fixed in 91ea1c7cc2. Exact provider/model matches now also require identical `api` and `baseUrl` before hosted metadata, including `compat`, is applied. Tests cover both API and URL mismatches while preserving header pinning.

packages/ai/src/model-compat-schema.ts
```diff
@@ -0,0 +1,103 @@
...
+	return {
+		openAICompletions,
+		openAIResponses,
+		anthropicMessages,
+		provider: Type.Union([openAICompletions, openAIResponses, anthropicMessages]),
+	};
+}
+
+const strictSchemas = createCompatSchemas(true);
+export const ProviderCompatSchema = createCompatSchemas(false).provider;
```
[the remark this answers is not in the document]
sethkarten: Fixed in a1abe483f1. The permissive local compatibility schemas are now intersected, so every known optional field is type-checked while unknown provider extension fields remain allowed. Added a focused regression test for valid, malformed, and extension fields.

packages/coding-agent/src/core/remote-model-catalog.ts
```diff
@@ -0,0 +1,178 @@
...
+			chunks.push(value);
+		}
+	} finally {
+		reader.releaseLock();
+	}
+	const text = Buffer.concat(chunks, bytesRead).toString("utf8");
+	return parseModelCatalog(JSON.parse(text) as unknown);
+}
+
+export async function refreshRemoteModelCatalog(
```
snimu: With a cold or expired cache, model resolution awaits this refresh, so startup can hang up to the full 5s timeout on a slow or flaky network. Consider serving the bundled/cached catalog immediately and refreshing in the background — a one-day-stale catalog is harmless, a 5s startup stall is felt every cold start.

[written by prime-agent, checked by snimu]

packages/ai/src/model-catalog.ts
```diff
@@ -0,0 +1,94 @@
...
+	}
+	if (!Array.isArray(value.models) || value.models.length === 0 || value.models.length > MAX_MODEL_CATALOG_MODELS) {
+		throw new Error("Invalid model catalog model count");
+	}
+	if (!Value.Check(ModelCatalogSchema, value)) throw new Error("Invalid model catalog entry");
+
+	const models = value.models as Model<Api>[];
+	const seen = new Set<string>();
+	for (const model of models) {
+		if (!isModelCompat(model.api, model.compat)) throw new Error("Invalid model catalog entry");
```
snimu: A single catalog entry with an unrecognized `api` or compat shape fails validation for the entire catalog, so the day a new API ships, older clients lose all remote updates instead of just skipping the new models. Skipping unknown entries (with a debug note) would keep old releases updatable exactly when the catalog matters most.

[written by prime-agent, checked by snimu]

packages/coding-agent/src/core/remote-model-catalog.ts
```diff
@@ -0,0 +1,178 @@
...
+		headers: template.headers ? { ...template.headers } : undefined,
+		compat: remote.compat ? structuredClone(remote.compat) : undefined,
+	};
+}
+
+function modelKey(provider: string, id: string): string {
+	return JSON.stringify([provider, id]);
+}
+
+export function mergeRemoteModelCatalog(
```
snimu: A syntactically valid but truncated catalog — say 3 models after a bad publish — is accepted wholesale and replaces the bundled view of several hundred. A cheap floor (reject if the remote catalog has materially fewer models than the bundled one) would make a botched publish degrade to "stale" instead of "almost everything missing".

[written by prime-agent, checked by snimu]

packages/ai/src/model-compat-schema.ts
```diff
@@ -0,0 +1,108 @@
...
+
+const strictSchemas = createCompatSchemas(true);
+const localSchemas = createCompatSchemas(false);
+export const ProviderCompatSchema = Type.Intersect([
+	localSchemas.openAICompletions,
+	localSchemas.openAIResponses,
+	localSchemas.anthropicMessages,
+]) as unknown as typeof localSchemas.provider;
+
+export function isModelCompat(api: string, value: unknown): boolean {
```
snimu: Six APIs shipped in the bundled catalog have no strict compat schema here, so remote entries claiming those APIs fall through to `return false` and get rejected with weaker/no validation compared to the three covered ones. Either add the schemas or have the generator assert every bundled API has one, so a gap can't reappear silently.

[written by prime-agent, checked by snimu]

.github/workflows/refresh-model-catalog.yml
```diff
@@ -0,0 +1,61 @@
...
+          PRIME_AGENT_MODEL_CATALOG_STRICT: "1"
+        run: npm run generate-models
+
+      - name: Validate catalog
+        working-directory: packages/ai
+        run: npm run validate-model-catalog -- /tmp/model-catalog.json
+
+      - name: Publish catalog to R2
+        env:
+          AWS_ACCESS_KEY_ID: ${{ secrets.R2_ACCESS_KEY_ID }}
```
snimu: The R2 credential used here should be confirmed bucket-scoped and write-only, since a broader token would make this workflow the catalog's weakest link. Nothing in-repo can verify this — flagging for a one-time check in the Cloudflare dashboard.

[written by prime-agent, checked by snimu]

packages/coding-agent/src/core/remote-model-catalog.ts
```diff
@@ -0,0 +1,178 @@
+import { Buffer } from "node:buffer";
+import { existsSync, readFileSync, renameSync, unlinkSync, writeFileSync } from "node:fs";
+import { type Api, type Model, type ModelCatalogV1, parseModelCatalog } from "@earendil-works/pi-ai";
+import { isTruthyEnvFlag } from "../utils/env.js";
+
+const DEFAULT_PRIME_AGENT_DOWNLOAD_BASE_URL = "https://pub-728493de92a943e2a9b2d17b4719f318.r2.dev";
```
snimu: This constant and the env-override/trailing-slash normalization below duplicate `version-check.ts:3,89` verbatim; if the download host ever changes, the two copies drift silently. Worth one shared constant/helper for both callers.

[written by prime-agent, checked by snimu]

packages/coding-agent/src/core/remote-model-catalog.ts
```diff
@@ -0,0 +1,178 @@
...
+				? exactTemplate
+				: transports.get(`${remote.provider}\0${remote.api}\0${remote.baseUrl}`);
+		if (template) merged.push(cloneTransport(template, remote));
+	}
+	return merged.length > 0 ? merged : bundledModels.map((model) => structuredClone(model));
+}
+
+export function getRemoteModelCatalogUrl(): string {
+	const explicit = process.env.PRIME_AGENT_MODEL_CATALOG_URL?.trim();
+	if (explicit) return explicit;
```
snimu: The URL from `PRIME_AGENT_MODEL_CATALOG_URL` (and the base override) is used without an `https:` check, so a plain-HTTP override silently downgrades the transport for unsigned JSON that feeds model resolution. Rejecting non-HTTPS here is one line.

[written by prime-agent, checked by snimu]

packages/coding-agent/src/core/remote-model-catalog.ts
```diff
@@ -0,0 +1,178 @@
...
+}
+
+export function mergeRemoteModelCatalog(
+	bundledModels: readonly Model<Api>[],
+	remoteModels: readonly Model<Api>[] | undefined,
+): Model<Api>[] {
+	if (!remoteModels) return bundledModels.map((model) => structuredClone(model));
+	const exact = new Map(bundledModels.map((model) => [modelKey(model.provider, model.id), model]));
+	const transports = new Map<string, Model<Api>>();
+	for (const model of bundledModels) transports.set(`${model.provider}\0${model.api}\0${model.baseUrl}`, model);
```
snimu: When two bundled models share a `(provider, api, baseUrl)` transport key, the later one silently wins as the template for remote entries; which fields that decides (headers in particular) is invisible at the call site. One sentence documenting last-wins — or picking deterministically — would spare the next reader an archaeology session.

[written by prime-agent, checked by snimu]

packages/coding-agent/src/core/remote-model-catalog.ts
```diff
@@ -0,0 +1,178 @@
...
+		provider: remote.provider,
+		baseUrl: template.baseUrl,
+		reasoning: remote.reasoning,
+		...(remote.thinkingLevelMap ? { thinkingLevelMap: { ...remote.thinkingLevelMap } } : {}),
+		input: [...remote.input],
+		cost: { ...remote.cost },
+		contextWindow: remote.contextWindow,
+		maxTokens: remote.maxTokens,
+		...(remote.featured !== undefined ? { featured: remote.featured } : {}),
+		headers: template.headers ? { ...template.headers } : undefined,
```
snimu: A remote entry carrying its own `headers` field is silently discarded in favor of the template's. That's the right call, but doing it silently keeps the "remote data can't touch transport" invariant implicit — rejecting such entries loudly (or a short note here) would make it visible.

[written by prime-agent, checked by snimu]

packages/coding-agent/src/core/remote-model-catalog.ts
sethkarten: Fixed in 95dfe25177. `refreshAvailableModels()` now serves cached/bundled models immediately and applies the hosted refresh in the background. The regression test holds the fetch open and verifies the method returns before it completes.

packages/ai/src/model-catalog.ts
sethkarten: Fixed in 95dfe25177. Publication still uses strict whole-catalog validation, while runtime/cache parsing skips incompatible future entries and retains compatible entries. A catalog with no compatible entries still fails closed.

packages/coding-agent/src/core/remote-model-catalog.ts
sethkarten: Fixed in 95dfe25177. Clients now require at least 50% usable coverage relative to the bundled catalog. A truncated refresh is rejected before cache write, so a healthy stale catalog is preserved; merge also enforces the same floor.

packages/ai/src/model-compat-schema.ts
sethkarten: Addressed in 95dfe25177. The six APIs intentionally have `compat: never`, so undefined compat remains valid and any compat object is rejected. The validator now uses an exhaustive `Record<KnownApi, TSchema | null>`, so adding a bundled API without an explicit schema/no-schema decision fails at compile time.

packages/coding-agent/src/core/remote-model-catalog.ts
sethkarten: Fixed in 95dfe25177. The default download origin and `PRIME_AGENT_DOWNLOAD_BASE_URL` normalization now live in one shared `prime-agent-download.ts` helper used by both catalog refresh and version checks.

packages/coding-agent/src/core/remote-model-catalog.ts
sethkarten: Fixed in 95dfe25177. Resolved catalog URLs are parsed and must use `https:`. Invalid or insecure overrides safely fall back without fetching or crashing startup, and the provider docs now state the HTTPS requirement.

packages/coding-agent/src/core/remote-model-catalog.ts
sethkarten: Fixed in 95dfe25177. Shared transport templates now select the lexicographically lowest bundled model ID, making the result independent of input order. Regression coverage uses two templates with different headers.

packages/coding-agent/src/core/remote-model-catalog.ts
sethkarten: Fixed in 95dfe25177. Publication removes headers, strict catalog validation rejects entries that carry them, and merge-level transport pinning still copies only bundled headers as defense in depth.

.github/workflows/refresh-model-catalog.yml
sethkarten: Agreed. This cannot be verified from the repository. I am leaving this thread open and escalating a one-time Cloudflare dashboard check that the workflow credential is restricted to the intended bucket with the narrowest available object-write permission.

packages/coding-agent/src/core/remote-model-catalog.ts
```diff
@@ -0,0 +1,202 @@
...
+	let url: string;
+	try {
+		url = getRemoteModelCatalogUrl();
+	} catch {
+		return undefined;
+	}
+	const now = options.now ?? Date.now();
+	const minimumModels = options.minimumModels ?? 1;
+	const cached = readCache(cachePath, url, minimumModels);
+	if (cached && now - cached.fetchedAt < MODEL_CATALOG_CACHE_TTL_MS) return cached.catalog.models;
```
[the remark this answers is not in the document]
sethkarten: Fixed in 97d6e39a23. A cached catalog is fresh only when `fetchedAt <= now` and its age is below the TTL. Added regression coverage showing a future-dated cache triggers another fetch.

.github/workflows/refresh-model-catalog.yml
sethkarten: Superseded by the Prime Inference pivot. This PR now deletes the R2 publication workflow and reads public models directly from `https://api.pinference.ai/api/v1/models`, so no catalog bucket or Cloudflare credentials are involved.

packages/coding-agent/src/core/prime-inference-models.ts
```diff
@@ -24,7 +31,7 @@ const PRIVATE_PRIME_INFERENCE_MODELS: readonly Model<"openai-completions">[] = [
 ];
 
 export function isPrivatePrimeInferenceModel(model: Pick<Model<string>, "provider" | "id">): boolean {
-	return model.provider === "prime-inference" && model.id.startsWith("internal/");
+	return model.provider === "prime-inference" && (model.id.startsWith("internal/") || model.id.startsWith("dev/"));
```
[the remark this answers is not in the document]
sethkarten: Fixed in 973ae48d1. A shared, case-insensitive private-route predicate now classifies `internal/*`, `dev/*`, and IDs containing `:`. Model resolution and registry authorization therefore gate deployment routes consistently. Added regression coverage for prefixed and colon deployment IDs.

packages/coding-agent/src/core/prime-inference-model-catalog.ts
```diff
@@ -0,0 +1,185 @@
...
+
+export function buildPrimeInferenceModels(
+	bundledModels: readonly Model<"openai-completions">[],
+	entries: readonly PrimeInferenceCatalogEntry[],
+	options: { includePrivate?: boolean; minimumModels?: number } = {},
+): Model<"openai-completions">[] | undefined {
+	const bundled = new Map(bundledModels.map((model) => [model.id.toLowerCase(), model]));
+	const models = entries.flatMap((entry): Model<"openai-completions">[] => {
+		const normalizedId = entry.id.toLowerCase();
+		if (!options.includePrivate && (normalizedId.startsWith("internal/") || normalizedId.startsWith("dev/"))) {
```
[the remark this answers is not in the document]
sethkarten: Fixed in 973ae48d1. Anonymous runtime merges and bundled generation now exclude IDs containing `:` as private deployment routes. Authenticated discovery still admits them only after the team-scoped `/models` response authorizes them. Added regression coverage.

packages/ai/.changes/prime-inference-model-catalog.md
```diff
@@ -0,0 +1,5 @@
+---
+"@earendil-works/pi-ai": patch
+---
+
+Parse Prime Inference model names, pricing, limits, modalities, and reasoning support from the live `/models` response.
```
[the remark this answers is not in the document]
sethkarten: Fixed in 973ae48d1. Both package fragments now use the repository’s required `- ...` bullet format with no Changesets front matter.

packages/ai/src/prime-inference-model-catalog.ts
```diff
@@ -0,0 +1,75 @@
...
+		const inputModalities = Array.isArray(modalities.input) ? modalities.input : [];
+		const contextWindow = positiveInteger(specs.context_window);
+		const maxTokens = positiveInteger(specs.max_output_tokens);
+		const reasoning = typeof specs.supports_reasoning === "boolean" ? specs.supports_reasoning : undefined;
+		const hasSpecs = contextWindow !== undefined && maxTokens !== undefined && reasoning !== undefined;
+
+		seen.add(item.id);
+		models.push({
+			id: item.id,
+			...(typeof item.display_name === "string" && item.display_name.trim()
```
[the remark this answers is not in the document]
sethkarten: Fixed in 973ae48d1. Generated source now serializes every string field with `JSON.stringify` in a dedicated renderer, including IDs and display names. Added a regression test with quotes, a newline, and injected source text.

sethkarten:
@snimu The PR has been fully redesigned around the live Prime Inference `/api/v1/models` endpoint. The R2 catalog workflow is removed, public refreshes are anonymous and nonblocking, team-private routes stay authorization-gated, and incomplete unknown routes are skipped. All 21 checks are green and all bot threads are resolved. Could you take the final approval review?

packages/ai/src/prime-inference-model-catalog.ts
```diff
@@ -0,0 +1,77 @@
...
+			output,
+			...(cacheRead !== undefined ? { cacheRead } : {}),
+			...(cacheWrite !== undefined ? { cacheWrite } : {}),
+			...(hasSpecs
+				? {
+						contextWindow,
+						maxTokens: Math.min(maxTokens, contextWindow),
+						vision: inputModalities.includes("image"),
+						reasoning,
+					}
```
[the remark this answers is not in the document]
sethkarten: Fixed in 24d3413c3. Live specs are now considered complete only when both input and output modality arrays are present and valid. Otherwise known models retain bundled metadata and unknown models are skipped. Added a regression test for limits/reasoning with a missing input modality list.

packages/coding-agent/src/core/prime-inference-model-catalog.ts
```diff
@@ -0,0 +1,175 @@
...
+			input: (entry.vision ?? template?.input.includes("image")) ? ["text", "image"] : ["text"],
+			cost: { input: entry.input, output: entry.output, ...cacheCosts(entry, template) },
+			contextWindow,
+			maxTokens,
+			...(template?.featured ? { featured: true } : {}),
+			compat: structuredClone(template?.compat ?? DEFAULT_COMPAT),
+		});
+	}
+	const minimumModels = options.minimumModels ?? Math.ceil(bundledModels.length * MIN_CATALOG_COVERAGE);
+	return models.length >= minimumModels ? models : undefined;
```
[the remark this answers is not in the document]
sethkarten: Fixed in 24d3413c3. The coverage floor now counts live IDs that overlap the bundled Prime Inference catalog, rather than total live IDs. A disjoint catalog of entirely new IDs is rejected, while additions remain accepted when bundled coverage is healthy. Added a regression test for the disjoint-ID case.

packages/coding-agent/src/core/model-registry.ts
```diff
@@ -803,38 +848,51 @@ export class ModelRegistry {
 			this.authorizedPrivatePrimeInferenceModelIds = new Set(cached.modelIds);
 			this.authorizedPrivatePrimeInferenceTeamId = teamId;
 			const cacheIsFresh = Date.now() - cached.refreshedAt < PRIVATE_PRIME_AUTHORIZATION_CACHE_TTL_MS;
-			if (cacheIsFresh || isOfflineModeEnabled()) {
-				return;
-			}
+			const missingModelMetadata = [...cached.modelIds].some((id) => !this.find(PRIME_INFERENCE_PROVIDER_ID, id));
+			if (isOfflineModeEnabled() || (cacheIsFresh && !missingModelMetadata)) return;
```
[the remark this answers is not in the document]
sethkarten: Fixed in 24d3413c3. The credential/team-fingerprint-scoped private cache now stores validated private catalog metadata, not only IDs, and restores those models before returning. The cache remains separate from the anonymous public cache, uses mode 0600 atomic writes, and rebuilds models through the fixed Prime Inference transport. Added a restart/offline regression test for a deployment route.

packages/coding-agent/src/core/prime-inference-model-catalog.ts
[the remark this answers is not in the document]
sethkarten: Fixed in 24d3413c3. The coverage floor now counts live IDs that overlap the bundled Prime Inference catalog, rather than total live IDs. A disjoint catalog of entirely new IDs is rejected, while additions remain accepted when bundled coverage is healthy. Added a regression test for the disjoint-ID case.

sethkarten:
Included in #2068: https://github.com/PrimeIntellect-ai/prime-agent/pull/2068

PR #1952, opened 2026-09-01
fix(coding-agent): make subagent deletion user-requested in the system prompt

snimu:
The delegation section of the system prompt instructs the model to "Delete a direct child explicitly with `await rlm.delete_subagent(child)` when it is no longer needed." Models follow it: finished subagents get deleted routinely, destroying their sessions and any warm context that would make them cheap to reuse for follow-up work. Idle children cost little to keep.

The sentence now reads: "You can delete a direct child explicitly with `await rlm.delete_subagent(child)`. Only do this if and when a user asks for it." The capability stays documented; the default behavior flips from delete-when-done to keep-unless-asked.

Prompt-content tests still pass unchanged (they pin the call form, not the old sentence).

Linear: [ENG-5836](https://linear.app/primeintellect/issue/ENG-5836/system-prompt-tells-models-to-delete-subagents-when-done)

> [!NOTE]
> ### Restrict subagent deletion to user-requested only in `buildRlmPrompt`
> Updates the delegation guidance in the RLM system prompt so that explicit deletion of a direct child subagent occurs only when the user asks for it. Previously the prompt directed deletion when the child was no longer needed. Idle children now remain available for follow-ups.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 5523618.</sup>
> 
>

### DIFF
--- a/packages/coding-agent/.changes/prompt-subagent-deletion-user-only.md
+++ b/packages/coding-agent/.changes/prompt-subagent-deletion-user-only.md
@@ -1,0 +1,1 @@
+- The system prompt no longer tells the model to delete finished subagents by default; deletion is now described as user-requested only, so idle children stay available for follow-ups.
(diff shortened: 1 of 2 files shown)


hallerite (approved):

sethkarten (commented):
Not approved. This will lead to substantial subagent build up and kill RAM

PR #1991, opened 2026-09-02
feat(coding-agent): add public agent creation command

sethkarten:
## Summary

- add a public `prime-agent create [options] [name] -- <message>` command
- create the top-level agent and submit its first user prompt through the existing daemon protocol
- start the background service on cold invocations
- cover routing, prompt parsing, empty-message rejection, JSON output, and cold startup

## Context

PR #423 intentionally removed the public `prime-agent daemon ...` hierarchy in favor of task-oriented commands. Automation that created top-level sessions with `daemon create` and `daemon prompt` then had no supported replacement, while `list`, `send`, `attach`, and other agent operations did.

This adds the missing task-oriented operation without restoring the old daemon command group. It is a backward-compatible CLI addition and does not change the daemon protocol or response schema.

Linear: [RES-1256](https://linear.app/primeintellect/issue/RES-1256/add-a-public-cli-command-for-top-level-agent-creation)

Related: #423

## Validation

- `cd packages/coding-agent && npx tsx ../../node_modules/vitest/dist/cli.js --run test/daemon-command.test.ts test/daemon-launch.test.ts test/public-command.test.ts` (85 passed)
- `npm run check`

### DIFF
--- a/packages/coding-agent/.changes/fix-public-agent-create.md
+++ b/packages/coding-agent/.changes/fix-public-agent-create.md
@@ -1,0 +1,1 @@
+- Added a public `prime-agent create` command for starting top-level agents with an initial message.
--- a/packages/coding-agent/src/cli/command-registry.ts
+++ b/packages/coding-agent/src/cli/command-registry.ts
@@ -27,6 +27,21 @@ export const COMMAND_SPECS: readonly CommandSpec[] = [
 		options: ["-a, --all  Include saved agents", "--json      Print JSON"],
 	},
 	{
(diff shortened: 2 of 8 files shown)


sethkarten:
Closing in favor of #1996.

This PR added a second CLI entry point, but the intended capability is explicit top-level session creation from the RLM host. #1996 keeps `rlm(...)` child semantics unchanged and adds the separate `rlm.create_session(...)` operation for daemon-backed depth-0 agents.

PR #2051, opened 2026-09-04
fix(coding-agent): memoize the passive RLM topology derivation

snimu:
## Purpose

`listPassiveRlmSubagents` walks the whole persisted RLM tree — a full spawn-ledger replay plus a per-child session scan, display-file read, and legacy-registry read — and is invoked uncoalesced from nine daemon sites (session list, snapshot stabilization, cron recovery, agent-observe, agent-message, family catalog, passivation, selector resolution). With N persisted children and M requests needing topology, cost was N×M walks; on large trees this is the remaining request multiplier behind the duplicate-FD/CPU incidents in discussion #1671 (the per-child scan cost is fixed by #2043).

## Mechanism

One memoized, strictly serialized derivation per argument shape in `daemon-mode.ts`:

- A memo hit requires the spawn-ledger stat, the resident roster, the live saved-root set, every listed child file's stat, and every embedded root-state identity to be unchanged — ledger appends (spawn/rename/delete from any process), residency changes (add/passivate/rehydrate, including same-id state-object swaps), and child-file appends all invalidate.
- Walks of the same shape run one at a time and a later caller never joins an earlier walk: its own pass revalidates against state at or after its call time (same freshness contract as #2043's scan chaining).
- A walk whose inputs changed while it ran never memoizes — in particular the first walk, which seeds the ledger from legacy registries.
- Empty topologies (no live saved roots, no persisted resident session) stay IO-free, exactly as before.

All nine call sites become readers of the cached walk; no call-site changes.

## Measurement (200 passive children under one resident root, real `AgentDaemon`)

| scenario | main | this PR |
| --- | --- | --- |
| 50 sequential listings | 965ms | 17ms |
| 20 concurrent listings | 342ms | 7ms |

(Both sides measured without #2043, so the delta isolates the walk memo; the two compose multiplicatively on actively-written trees where the per-child `(size,mtime)` cache never hits.)

## Validation

- One pin verified fail-unfixed on main: unchanged inputs return the identical cached derivation, and a child-file append invalidates and re-derives fresh infos (fails on main with fresh arrays per call).
- Full `daemon-mode` suite (187) passes under a sanitized env, including the passive-listing, hydration-race, and snapshot-stabilization tests; root `npm run check` passes.

## LOC

Total src: **+137/−9 (net +128)**; tests: +149/−0 (net +149).
Src +139/−3: additive memo layer around the existing walk (renamed `walkPassiveRlmSubagents`, body unchanged except the hoisted lifecycle filter); no behavioral rewrite of the walk itself. Tests +44.

Squashes the request-multiplier half of discussion #1671; with #2043 and #2050 this closes the P4 session-file IO program (also reduces the #1503 OOM profile).

Linear: RES-1272 https://linear.app/primeintellect/issue/RES-1272

---

> [!NOTE]
> **Medium Risk**
> Incorrect memo invalidation could serve a stale passive topology across session list, messaging, and passivation paths; the change relies on comprehensive file-stat and roster fingerprinting rather than trivial logic.
> 
> **Overview**
> **Cuts duplicate work when the daemon repeatedly derives passive (non-resident) RLM subagent trees** — the expensive spawn-ledger walk plus per-child session reads that many request paths already shared.
> 
> `listPassiveRlmSubagents` now memoizes results per argument shape (include-resident flag + live saved roots), with invalidation driven by spawn-ledger stat, resident session roster, saved-root set, per-file stats captured during the walk, and whether cached root parent state objects are still the live in-memory sessions. Same-shape walks are serialized so a later caller does not piggyback on an in-flight walk; walks that see inputs change mid-flight or mark **degraded** (e.g. unreadable but present child file) are not cached. Empty topologies with no live roots and no persisted resident sessions stay IO-free.
> 
> The traversal body moves to `walkPassiveRlmSubagents`, which records input stats and a degraded flag; call sites keep calling `listPassiveRlmSubagents`. New `daemon-mode` tests cover cache reuse, invalidation on transcript/display/ledger changes, absent-then-restored children, resident-child appends not busting the memo, and rejection cleanup without unhandled rejections.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0ee114ce77942d9298136ca193ddc2c0b5d27d58. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Memoize passive RLM topology derivation in `AgentDaemon`
> - Caches results of `listPassiveRlmSubagents` by argument shape, retaining at most four memo entries; repeated requests with unchanged inputs return the cached result without filesystem I/O.
> - Adds validation helpers that fingerprint the spawn-ledger stat, resident-session roster, and live root paths; a cached entry is rejected when any visited file's stat identity changes, a resident root is removed, or the walk degraded.
> - Extracts the filesystem walk into `walkPassiveRlmSubagents`, which captures per-input stat identities needed for cache validation and excludes resident-child transcript appends from invalidation.
> - Same-key concurrent walks run sequentially without sharing promises; failed walks remain request-scoped and are retried by later calls.
> - Behavioral Change: `listPassiveRlmSubagents` in [daemon-mode.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/2051/files#diff-f657fb72071b9bcb67235932ddc24164a0376dc918944a8f58b857c2ac43e450) now returns empty without I/O when no relevant roots or persisted residents exist, and cached results are dropped if `passiveRlmRootsStillResident` detects the referenced root state objects are no longer in the daemon roster.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 0ee114c.</sup>
> 
>

### DIFF
--- a/packages/coding-agent/.changes/res-1272-memoized-passive-topology.md
+++ b/packages/coding-agent/.changes/res-1272-memoized-passive-topology.md
@@ -1,0 +1,1 @@
+- Fixed daemon request latency on large agent trees: the passive-subagent topology is derived once and memoized, with every consumer (session list, snapshots, cron recovery, agent messaging, passivation) reading the cached walk until the spawn ledger, residency, or a child session file changes.
--- a/packages/coding-agent/src/modes/daemon/daemon-mode.ts
+++ b/packages/coding-agent/src/modes/daemon/daemon-mode.ts
@@ -1436,6 +1436,12 @@ export class AgentDaemon {
 				const sessionKey = resolve(entry.sessionFile);
 				if (entry.status === "deleted" || visited.has(sessionKey)) continue;
 				visited.add(sessionKey);
+				// A resident child walks its own subtree as an outer root below. Avoid
+				// duplicate rows and attributing its descendants to an ancestor - and
+				// keep its actively-streamed transcript out of the input set, where
+				// every append would needlessly bust the memo (the roster axis
+				// already carries resident identity).
+				if (!includeResident && this.findSessionBySessionFile(entry.sessionFile)) continue;
 				const childStat = await recordInputStat(entry.sessionFile);
 				const info = await readSessionInfo(entry.sessionFile);
 				if (!info) {
@@ -1444,9 +1450,6 @@ export class AgentDaemon {
 					if (childStat !== "absent") degraded = true;
 					continue;
 				}
-				// A resident child walks its own subtree as an outer root below. Avoid
-				// both duplicate rows and attributing its descendants to an ancestor.
-				if (!includeResident && this.findSessionBySessionFile(entry.sessionFile)) continue;
 				const chain = [...parentChain, entry];
 				passive.push({ ...root, entry, info, chain });
 				await visit(root, { sessionId: info.id, sessionFile: entry.sessionFile }, chain, visited);
--- a/packages/coding-agent/src/modes/daemon/daemon-mode.ts
+++ b/packages/coding-agent/src/modes/daemon/daemon-mode.ts
@@ -1269,15 +1269,18 @@ export class AgentDaemon {
 	// agent-observe, agent-message, family catalog, passivation) reads the same
 	// cached walk instead of re-walking the persisted tree per request. A hit
 	// requires the spawn ledger stat, the resident roster, the live saved
-	// roots, every listed child file's stat, and every root-state identity to
-	// be unchanged; ledger appends, residency changes, and child appends all
-	// invalidate. Walks of the same shape run strictly one at a time, and a
-	// caller never joins an earlier walk: its own pass re-validates against
-	// state at or after its call time.
+	// roots, every root-state identity, and the stat of every file input the
+	// walk visited (child transcripts, display records, legacy registries,
+	// including absent ones) to be unchanged; ledger appends, residency
+	// changes, child appends, and display rewrites all invalidate. Input stats
+	// are captured before each read, so a write landing mid-walk can only
+	// cause one extra re-walk, never a stale memo. Walks of the same shape run
+	// strictly one at a time, and a caller never joins an earlier walk: its
+	// own pass re-validates against state at or after its call time.
 	private readonly passiveRlmSubagentWalks = new Map<string, Promise<PassiveRlmSubagent[]>>();
 	private readonly passiveRlmSubagentMemo = new Map<
 		string,
-		{ fingerprint: string; childStats: string[]; result: PassiveRlmSubagent[] }
+		{ fingerprint: string; inputStats: Map<string, string>; result: PassiveRlmSubagent[] }
 	>();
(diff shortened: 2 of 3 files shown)


snimu:
Addressed all four findings in c6b121724, itemized:

- **H (unhandled rejection from discarded `finally()`) — fixed.** Cleanup runs via `walk.then(cleanup, cleanup)`; the rejection still reaches the caller through `walk` itself and no orphan promise remains. Pin: a malformed ledger line makes the listing reject request-scoped with no unhandled rejection, and the next call after repair succeeds (fails on the previous head with vitest reporting an Unhandled Rejection).
- **M1 (child stats captured after the reads) — fixed.** Every file input's stat is recorded immediately before its read inside the walk (`recordInputStat`), so the memo's invalidation identity can only be older than the derived content: a write landing mid-walk makes the hit-time re-stat differ and costs one extra re-walk, never a stale memo — same capture-before/compare-after contract as #2043's scan chaining. No deterministic pin: constructing the mid-walk interleaving needs a test-only seam between a child read and the stat capture, which I declined to add; the mechanism is the ordering itself and is exercised by the other input-stat pins.
- **M2 (only listed children tracked) — fixed.** The identity map records every visited edge's child path including absent ones ("absent" is a stat identity), and a walk that failed to read a *present* child file is marked degraded and never memoizes, so transient read errors retry on the next call while permanently-corrupt children simply keep today's re-walk behavior. Pin: a grandchild file renamed away and back re-lists (fails on the previous head — the restored file was invisible indefinitely).
- **M3 (display records not in the identity) — fixed.** Display files and legacy registry files are recorded as walk inputs per edge, so a cross-process `running` -> `completed` display rewrite after the final transcript append re-derives instead of serving stale status (which snapshots map to `error`). Pin: rewriting only `rlm-subagent.json` flips the listed status (fails on the previous head).

`daemon-mode` 190/190 under a sanitized env; `npm run check` green. Memo-hit cost grows to one stat per walk input (children + display + registry paths) — for the 200-child bench fixture the listings stay at 17ms vs 965ms on main.

snimu:
Correction to the bench line above: with the per-input stat validation the 200-child fixture measures 64ms per 50 sequential listings and 23ms per 20 concurrent (was 17ms/7ms before this round; main: 965ms/342ms). The added cost is the promised one-stat-per-input hit validation.

snimu:
Fixed the resident-transcript finding in ea80edefb: the walk recorded a resident child's session-file stat before skipping it as resident, so every streamed append from an active child busted the memo — constant re-walks with no staleness benefit, since a resident child's row is never in the passive result and its identity comes from the roster axis. The resident skip now runs before the stat capture and the read (which also saves the read itself). Pin: the memo survives appends to a resident child's transcript while its passive grandchild stays listed (fails on the previous head). `daemon-mode` 191/191 sanitized.

sethkarten:

PR #2130, opened 2026-09-08
preserve reasoning in refinement and side questions

kevinjosethomas:
Fixes [ENG-6010](https://linear.app/primeintellect/issue/ENG-6010), reported in [discussion #1482](https://github.com/PrimeIntellect-ai/prime-agent/discussions/1482) and [discussion #2081](https://github.com/PrimeIntellect-ai/prime-agent/discussions/2081). Refinement discarded session reasoning and side questions forced it off. Auxiliary calls now map enabled reasoning to low and clamp it to the model’s supported levels, preserve supported off/minimal preferences, and leave the parent session unchanged.

Refinement reserves input and message-framing space, shortens older conversation text when needed, and gives reasoning-enabled calls the remaining output allowance. The conservative byte bound handles dense ASCII and Unicode; reserving up to half the context for output avoids starving reasoning and final JSON. Small Anthropic budgets enforce at least 1,024 thinking tokens below the total output ceiling, and impossible limits fail before dispatch. Truncated JSON remains an explicit error without a new inference retry.

Validation:

- 178 focused tests and `npm run check` passed locally and in the final-head Linux sandbox, covering reasoning preferences, complete/truncated JSON, long English and Unicode, dense ASCII, provider ceilings, and parent preservation.
- Separately compiled, packed, and installed baseline/fixed builds reproduced 12 failing budget cases; the correction passed all 18 expected outcomes. The final compiled head also passed those 18 probes and 10 GLM/Muse adapter/control operations.
- Three real Prime Inference `internal/glm-5.3-fast` calls passed on the final head: refinement returned parsed edits, automatic review returned parsed JSON, and the side question completed without changing the parent. Provider edge-case probes use loopback responses or stop before network dispatch; no live Z.ai, Muse, or Anthropic verification is claimed.
- Reused the shared Ubuntu apt-source repair and literal-prompt test compatibility fix. The latter still checks exact user text and a single provider call.

The conservative bound can shorten older context more than an exact tokenizer would. The retained live preview and full evidence are linked in the task handoff.

Linear: [ENG-6010](https://linear.app/primeintellect/issue/ENG-6010)

---

> [!NOTE]
> **Medium Risk**
> Changes how refinement, auto-review, and side questions call models (reasoning level and token/context math), which can alter JSON completion behavior on small context windows or reasoning-heavy models.
> 
> **Overview**
> Fixes **ENG-6010** by keeping reasoning on for auxiliary calls when the main session uses it, instead of forcing refinement, auto-refine review, and side questions off.
> 
> **Auxiliary thinking:** New `getAuxiliaryThinkingLevel` maps the session level to a model-safe auxiliary level (typically **low**, or **off**/**minimal** when supported). Side questions and refinement/review now pass that level into `completeSimple` / the side agent without changing the parent session.
> 
> **Refinement budgets:** A new `refinementRequest` helper reserves prompt space from the context window (byte-based input estimate, optional binary search to keep the **recent** conversation slice, Unicode-safe trimming), caps request `maxTokens` to remaining output room, and fails early when there is no output capacity. With reasoning **on**, the old 32k/4k output caps are dropped in favor of the model limit; with reasoning **off**, those caps still apply.
> 
> **Anthropic (`pi-ai`):** Budget-based thinking now enforces a **1024** minimum `budget_tokens` and rejects token ceilings that cannot satisfy provider rules.
> 
> **CI:** System package installs in GitHub Actions use Ubuntu’s official `ubuntu.sources` only, avoiding third-party apt repos on the runner.
> 
> Regression tests cover auxiliary reasoning matrices, refinement output/context behavior, and minimum Anthropic thinking budgets.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5f2630d1a64db3813c2996f2ea3792e9554b15b5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Preserve auxiliary reasoning in refinement and side-question calls
> - Adds `getAuxiliaryThinkingLevel` in [thinking-levels.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/2130/files#diff-7897093f92589b7a24ce01a6035d106ebaf16e383bb34d4f9039111386d778af) to map the parent session's thinking preference to a model-supported level (`off`, `minimal`, or `low`) for auxiliary calls
> - Refinement planning and auto-review in [refinement.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/2130/files#diff-17b614d876315ad8ac0165562a8f07a5c4ada3b28928eb02d90f6b93a96ea91d) now pass the auxiliary thinking level and a context-aware output budget to `completeSimple` instead of unconditionally suppressing reasoning; reasoning-off requests keep fixed caps (32,000 for planning, 4,096 for review)
> - Adds `refinementRequestModel` helper that reserves prompt bytes plus a fixed overhead from the context window and caps the model's `maxTokens` to the remaining output space, rejecting requests with no output capacity
> - Side-question runner in [side-question.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/2130/files#diff-854e565ab4c061597eb4ed5ef465f9735fd0b98ef7186bc8424f1df97bd18265) uses the parent's auxiliary thinking level instead of always disabling reasoning
> - Expands Linear-ticket CI validation in [linear-ticket.yml](https://github.com/PrimeIntellect-ai/prime-agent/pull/2130/files#diff-ce6fb4658459ed854d72950d809841bbd48b2cbae4ed1ab56d7e1a6b6be59b52) to accept both `ENG` and `RES` ticket references
> - Risk: reasoning-enabled refinement and review now share output budget between reasoning tokens and JSON output, so models with small context windows may produce shorter JSON than under the previous fixed caps; check `refinementRequestModel` in [refinement.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/2130/files#diff-17b614d876315ad8ac0165562a8f07a5c4ada3b28928eb02d90f6b93a96ea91d)
>
> 
> #### Changes since #2130 opened
>
> - Implemented character-based token estimation and context-aware truncation for refinement operations [8b48d18]
> - Enforced minimum thinking token budget of 1024 for Anthropic provider requests [8b48d18]
> - Added tests validating minimum thinking budgets and refinement context handling [8b48d18]
> - Constrained apt package manager operations in GitHub Actions CI workflow to use only `/etc/apt/sources.list.d/ubuntu.sources` with explicit `-o Dir::Etc::sourcelist` and `-o Dir::Etc::sourceparts` options in both `build-check` and `test` jobs [e13c481]
> - Moved harness state delivery from system prompt to separate digest messages injected at cold session boundaries and attached to compaction summaries [d21ed07]
> - Replaced refinement system prompt mutation with separate refinement notice messages and added refinement source tracking [d21ed07]
> - Added LiteLLM context overflow detection pattern and fixed orphaned tool result filtering [d21ed07]
> - Added OpenCode provider header enrichment for session tracking [d21ed07]
> - Fixed prompt template argument substitution to preserve literal dollar sequences and prevent re-expansion [d21ed07]
> - Fixed active goal resumption after manual compaction [d21ed07]
> - Updated Agents View visual hierarchy with muted styling for details and inline prompt rendering [d21ed07]
> - Bumped package versions from 0.9.3 to 0.9.4 across monorepo [d21ed07]
> - Added changeset notes for visual hierarchy, overflow handling, prompt arguments, goal continuation, and refinement cache visibility [d21ed07]
> - Replaced token estimation heuristic with UTF-8 byte-length upper bound for refinement input token budgeting [5f2630d]
> - Changed output token reservation in refinement requests to conditionally use model maximum tokens when reasoning is enabled [5f2630d]
> - Expanded and restructured refinement output budget regression tests to validate new token budgeting behavior [5f2630d]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 8a71650.</sup>
> 
>


packages/coding-agent/src/core/refinement/refinement.ts
```diff
@@ -177,24 +178,28 @@ Return JSON only:
...
+function refinementRequestModel(model: Model<Api>, systemPrompt: string, userPrompt: string): Model<Api> {
+	// Reserve one token per UTF-8 byte plus message framing, without assuming an English chars/token ratio.
+	const inputReserve =
+		Buffer.byteLength(systemPrompt, "utf8") +
+		Buffer.byteLength(userPrompt, "utf8") +
+		REFINEMENT_CONTEXT_OVERHEAD_TOKENS;
+	const maxTokens = Math.min(model.maxTokens, model.contextWindow - inputReserve);
+	if (maxTokens <= 0) {
+		throw new Error(
+			"Refinement prompt leaves no room for output in the model's context window; retry with a smaller request.",
```
snimu: `/refine` on a longer session can now be rejected before the provider is ever called, even when the request fits the model. On a 65,536-context model such as the catalog's `zai-org/GLM-4.5V` (16,384 output tokens), the existing 80,000-character trajectory slice can be ordinary English text taking roughly 20,000 tokens, so that prompt plus the full output budget fit — but the new admission check rejects the call because the conversation alone exceeds 65,536 bytes.

The check treats each UTF-8 byte as a token and, when that upper bound exceeds the context window, rejects the request instead of reducing the trajectory. This affects reasoning-off sessions too, because `refinementRequestModel` is called unconditionally. It regresses previously valid `/refine` calls on longer sessions, and multibyte text hits the false rejection even sooner.

A conservative output reservation must not become an unconditional false-negative admission check. It would be better to truncate to fit the chosen conservative bound, or use a closer token estimate/count before rejecting.

packages/coding-agent/src/core/refinement/refinement.ts
```diff
@@ -177,24 +178,28 @@ Return JSON only:
...
+		);
+	}
+	// Bound the request's model ceiling too: some adapters add thinking tokens before clamping to it.
+	return { ...model, maxTokens };
```
snimu: A refinement request that enables reasoning can still reach Anthropic as an invalid request instead of failing fast or being bounded locally. With Claude Sonnet 4/4.5 (not adaptive thinking), a prompt whose byte reservation leaves 1,500 tokens passes this check — the request-local `model.maxTokens` can now be any positive number — but enabling reasoning imposes an additional minimum on the budget-based Anthropic path that the check does not account for.

The `low` reasoning request then reaches `adjustMaxTokensForThinking` in `packages/ai/src/providers/simple-options.ts`, which sees the 1,500-token model ceiling and lowers the 2,048-token thinking budget to `1500 - 1024 = 476`; `streamSimpleAnthropic` forwards that, and the payload sends `thinking: { type: "enabled", budget_tokens: 476 }` (`packages/ai/src/providers/anthropic.ts:1041`). Anthropic requires at least 1,024 thinking tokens, so this is an invalid request, not the intended bounded truncation/fail-fast behavior. Remaining caps at or below 1,024 are also invalid, because the adapter falls back to a 1,024 thinking budget that is not smaller than `max_tokens`.

Previously refinement disabled thinking and did not manufacture this small request-model ceiling. Admission needs to honor the enabled adapter's minimum total budget (or shrink the prompt further), not only check `maxTokens > 0`.

snimu (approved):

PR #2132, opened 2026-09-08
preserve saved-session renames across live worker turns

kevinjosethomas:
- path-only saved-session renames now reach the owning worker, preserving the name across streamed turns while retaining offline renames and client ownership checks.
- local validation passed: 27 focused tests and `npm run check`; the new regression failed on baseline and passes with the fix.
- tracks [eng-6013](https://linear.app/primeintellect/issue/ENG-6013/a-saved-session-rename-reverts-in-the-session-list-when-the-live), from [discussion #2096](https://github.com/PrimeIntellect-ai/prime-agent/discussions/2096); daemon command, event, and response schemas are unchanged.

prime sandbox `vqccjkl80dpl7xmaw2tn47nd`: `node:24-bookworm` container, linux x64, node `24.18.0`, npm `11.16.0`, 2 cpu / 8 gb / 20 gb, 60-minute limit. separately compiled, bundled, packed, and installed the pinned upstream baseline and exact draft head, then exercised the installed cli, attached resident worker, and held-open loopback sse stream.

| packaged revision | file / initial list | worker / list during and after streaming | active-id controls |
| --- | --- | --- | --- |
| baseline `4ec05a0969825a32261ffed43ab5740071a73ae6` | new name | old name: reproduced | passed |
| draft head `96bad2b3be2d5191d68220a39640a65fd8fe6041` | new name | new name: passed | passed |

all 27 focused tests also passed in the sandbox. fixture processes exited, evidence was collected locally, and the sandbox was deleted. dummy credentials and a local provider only; no paid inference or graphical ui test. npm skipped unapproved postinstall scripts, so optional installer setup was outside this test. validation archives retained version `0.9.3`; no release was published.

Linear: [ENG-6013](https://linear.app/primeintellect/issue/ENG-6013/a-saved-session-rename-reverts-in-the-session-list-when-the-live)

> [!NOTE]
> ### Forward saved-session renames to live workers instead of offline catalog
> - Path-only `rename_saved_session` commands for live sessions now route to the owning worker in `DaemonSupervisor.handleCommand`, requiring client ownership; offline sessions still use the catalog path.
> - `DaemonSupervisor.forwardToWorker` translates `set_session_name` into the worker rename command, waits for earlier roster frames, and amends the current worker row with the acknowledged name. The requested name stays reserved until the forwarded operation completes or fails.
> - Adds extensive regression coverage in `6013-saved-session-rename.test.ts` for live-worker renames, offline catalog fallback, recovering/stopping worker rejection, client ownership, delayed roster ordering, competing create reservation, worker disconnect, and session-switch identity.
> - Risk: name reservation logic in `DaemonSupervisor.forwardToWorker` holds names across async roster writes; reviewers should check that rejected renames and row disappearance paths release the reservation so names are not permanently locked.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 9c70ba1.</sup>
> 
>

---

> [!NOTE]
> **Medium Risk**
> Supervisor rename routing, roster ordering, and name reservation behavior changed for live sessions; path-only renames can now fail for non-owners or recovering/stopping workers instead of silently using the catalog.
> 
> **Overview**
> Fixes **ENG-6013**: saved-session renames no longer snap back in the session list when a live worker streams another turn, and duplicate names are blocked while a rename is in flight.
> 
> **Live path-only renames** — When `rename_saved_session` is sent with only a session path and a resident worker owns that file, the supervisor **forwards to the worker** (with client access checks) instead of updating the offline catalog alone. Offline sessions still go through the catalog and RLM ledger.
> 
> **Roster commit after ack** — On successful `rename`, `set_session_name`, or forwarded `rename_saved_session`, `forwardToWorker` waits on `rosterApplyChain`, resolves the correct roster row (including post–session-switch identity), writes the trimmed acknowledged name, and maps worker `set_session_name` to `rename` on the wire.
> 
> Adds a **regression suite** (`6013-saved-session-rename.test.ts`) plus small test harness fixes (agent-traces outbox timing, supervisor rename mock expectations). `daemon-mode.ts` changes are formatting only.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9c70ba11158a721aba3bcdec3aa269c58413d312. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/coding-agent/src/modes/daemon/daemon-supervisor.ts
```diff
@@ -2457,6 +2457,11 @@ export class DaemonSupervisor {
 				return await this.withSessionNameReservation(target, async () => {
 					await this.assertSupervisorSavedSessionNameAvailable(command.sessionPath, target.name);
 					if (!command.activeSessionId) {
+						const owner = this.findWorkerBySessionFile(command.sessionPath);
+						if (owner) {
+							this.assertWorkerAccessibleToClient(client, owner, command.sessionPath);
+							return await this.forwardToWorker(owner, command);
```
snimu, line 2463: Two different top-level sessions can end up with the same name. The new early return releases the session-name reservation before the supervisor necessarily knows the new name: for a top-level session, the worker applies the rename and returns success immediately, but publishes its roster update through `scheduleRosterFlush()` (`setImmediate`), and `forwardToWorker()` updates the supervisor roster synchronously for `rename`, not for `rename_saved_session`.

If another request to rename or create a different top-level session with the same name reaches the supervisor after the first success but before that roster delta, its availability check uses the old roster name — `familyCatalogEntries()` deliberately ignores disk records for roster-owned paths — so both sessions can acquire the same name.

The previous path-only branch wrote the roster entry before releasing the reservation. It would be good to preserve that ordering on the new worker route, for example by updating the matching roster entry after successful forwarding while still inside the reservation. The regression fixture publishes every session event synchronously, so it cannot expose this IPC/publication gap.

packages/coding-agent/src/modes/daemon/daemon-supervisor.ts
```diff
@@ -5017,13 +5022,46 @@ export class DaemonSupervisor {
...
-			this.writeRosterEntry(workerRosterEntryFromSummary(response.data), worker);
-			return { ...response, id: command.id, data: this.publicSummary(worker, response.data) };
+		if (
+			response.success &&
+			(command.type === "rename" || command.type === "set_session_name" || command.type === "rename_saved_session")
+		) {
+			// An acknowledged name survives disconnects. Drain earlier frames, then amend the
+			// same session's current row before releasing its name reservation.
+			await worker.rosterApplyChain;
+			const renamedSummary =
```
snimu, line 5042: If a session switch lands between the set_session_name request and its ack, the ack names no session, so the guard matches a stale pre-request snapshot and skips the roster commit. The worker's own flush is then the only record of the claimed name; if the worker dies before that flush, the name is invisible to duplicate checks — the live row masks the catalog file — leaving a narrow duplicate-claim window. Returning the renamed summary in the ack, as rename_saved_session already does, would let the commit match on identity and close the window.

xeophon (approved):

PR #2133, opened 2026-09-08
preserve accepted harness memories across concurrent saves

kevinjosethomas:
Concurrent kernel memory writes and host `/refine` saves could both report success, then the later save replaced the entire file from an older snapshot and silently discarded the earlier accepted memory. Fixes [ENG-6007](https://linear.app/primeintellect/issue/ENG-6007/concurrent-python-and-refinement-writes-discard-accepted-harness), independently reproduced from the symptoms reported in [discussion #2009](https://github.com/PrimeIntellect-ai/prime-agent/discussions/2009).

Both writers now acquire the same adjacent file lock, reread the latest state, and merge only changes since their load snapshot before replacing the file atomically. Unrelated memories and refinement events survive; conflicting edits to an entry, schema, or replaced history fail with a reload-and-retry error. Python also preserves the stored schema, generates distinct default refinement IDs, rejects new non-finite JSON values, and normalizes legacy invalid numbers while retaining valid data.

Validation on commit `32bf15dcb885310d6e7eec0bc04342e8eb254896`:

- 116 TypeScript tests, 41 Python harness controls, and `npm run check` pass locally and in a Linux Prime sandbox. The TypeScript set includes 30 concurrency regressions and existing refinement/literal-prompt controls.
- The original bug reproduced against baseline in both save orders. New schema, refinement-history, identifier, and invalid-number regressions also failed before their respective fixes and pass afterward.
- 29 installed-artifact cases pass using the packaged host writer and shipped Python runtime, with no model calls. The installed version is `0.9.4-eng6007.32bf15dcb8`; the loaded Python module matches the installed artifact and pinned source by SHA-256.
- [Current CI](https://github.com/PrimeIntellect-ai/prime-agent/actions/runs/34388793236) passes, including all test shards and build/check. CodeQL, changelog, and Linear ticket checks also pass.

CI setup now scopes apt dependency installation to Ubuntu sources after repeated unrelated Chrome repository hash failures. The shared literal-prompt test correction checks the final provider user message, allowing the persistent-memory context added on main while preserving the exact prompt and transcript assertions.

Limits: both writers must be updated. A killed writer can leave a lock that must be removed only after all writers stop. Controlled interleavings establish the tested persistence behavior, not production failure frequency. Automatic stale-lock recovery, power-loss durability, and separate audit-log transactions are outside this change. Live inference is not part of this validation.

Linear: [ENG-6007](https://linear.app/primeintellect/issue/ENG-6007/concurrent-python-and-refinement-writes-discard-accepted-harness)

---

> [!NOTE]
> **Medium Risk**
> Changes shared continual-harness persistence used by refinement and the Python RLM kernel; incorrect merge/validation logic could still lose or block memories, though conflicts now fail closed and coverage is extensive.
> 
> **Overview**
> Fixes **ENG-6007**, where concurrent host `/refine` saves and Python kernel writes could both succeed but the last full-file write dropped the other writer’s accepted memories.
> 
> **TypeScript and Python** now share an adjacent `harness_state.json.lock`, re-read disk under the lock, and **merge only deltas** from each writer’s load snapshot (`mergeHarnessStateChanges` / `_merge_harness_changes`). Unrelated entries and compatible refinement appends survive; **conflicts** on the same entry, schema, or rewritten refinement history fail with **reload-and-retry** and persist nothing from that save. Python also **serializes** mutations on a shared `HarnessState` instance with an `RLock`.
> 
> **Save behavior change:** corrupt, malformed schema-1, unsupported schema, or **lossy/non-finite JSON** files stay **read-only**—loads still degrade for session use, but saves **refuse to overwrite** instead of self-healing. Writes must match **schema 1**; default Python refinement IDs are **UUID-based** to avoid duplicates under concurrency. Protocol and ops notes are in `harness-persistence.md`, with a large **6007** regression suite covering host/Python interleavings.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d06ef048172dcf5078b38024dc9efbb41754a976. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Preserve harness memories across concurrent TypeScript and Python saves
> - Adds a shared cross-runtime lock directory that serializes harness saves between TypeScript and Python, failing after a 10-second timeout. Lock is removed unconditionally on success or failure.
> - Introduces three-way state merging in [harness-persistence.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/2133/files#diff-84d32875dcf1f2ffc6958fd91aadc77223c0c7b69991ccb2be3d09602c9aad8b) and [harness.py](https://github.com/PrimeIntellect-ai/prime-agent/pull/2133/files#diff-441d64dfc5d74f553c86b609beff62210693b468d021d8bded095dea768138cf): saves re-read the latest on-disk state, preserve unrelated entries and compatible appended refinement events, and reject conflicting changes to the same entry, history, or schema.
> - Makes malformed, non-finite, or unsupported-schema harness files read-only: reads still return an empty view, but saves now refuse to overwrite instead of self-healing the file.
> - Changes default refinement event IDs from sequence-derived to UUID-based to avoid duplicates under concurrent writes.
> - Documents the full protocol in [harness-persistence.md](https://github.com/PrimeIntellect-ai/prime-agent/pull/2133/files#diff-597894eb3dfdd848bb92d9379fec45b19ae95696aff07612a16553a7ea83b584).
> - Behavioral Change: saves to a malformed or unsupported-schema file now throw `invalidHarnessStateError` / `_invalid_harness_state_error` rather than silently rewriting it; saves now block on the shared lock and may fail with a timeout error after 10 seconds if another writer holds it.
>
> 
> #### Changes since #2133 opened
>
> - Added schema validation to reject non-numeric and non-finite values before saving harness state [5e4293a]
> - Updated and added tests to verify schema validation behavior [5e4293a]
> - Added comprehensive validation infrastructure for harness state writes [d1b8199]
> - Replaced non-finite number checks with unsafe JSON number detection [d1b8199]
> - Modified harness state loading to tolerate malformed data while marking it invalid for writes [d1b8199]
> - Modified harness state saving to validate before writing and preserve existing content on validation failure [d1b8199]
> - Added regression test verifying concurrent save protection for corrupted harness files [d1b8199]
> - Clarified malformed file handling in harness persistence documentation [d1b8199]
> - Updated `rlm.harness._contains_unsafe_json_number` utility to detect integer-valued floats exceeding the maximum safe JSON integer [23bb98d]
> - Modified concurrent harness memory persistence test to corrupt the memory map container instead of individual entries [23bb98d]
> - Added per-instance reentrant locking to `HarnessState` in Python to serialize concurrent method calls [0bb2d6b]
> - Added validation to reject harness state JSON files containing numeric literals that lose precision when parsed [0bb2d6b]
> - Added tests verifying concurrent mutation serialization and rejection of lossy decimal literals in harness state files [0bb2d6b]
> - Updated documentation and changelog noting the concurrent mutation fix and Python serialization behavior [0bb2d6b]
> - Modified `readHarnessStateResult` function to preserve harness state entries and refinements when unsafe JSON numbers are detected [1c114e0]
> - Expanded concurrent harness memory persistence test coverage to include parameterized tests for both writer types and multiple unsafe number scenarios [1c114e0]
> - Added validation to reject negative zero and invalid decimal representations during JSON parsing in Python harness state loading [96f2fef]
> - Modified `normalizeJsonNumberToken` utility to preserve the sign prefix when normalized significant digits collapse to empty after trimming leading zeros [96f2fef]
> - Extended test coverage to verify rejection of extreme exponent forms and signed zero tokens across Python and host writers [96f2fef]
> - Modified JSON safety validation utilities to reject negative zero [6ae7772]
> - Added test coverage for negative zero rejection in harness memory persistence [6ae7772]
> - Modified `saveHarnessState` within the `refinement` module to select merge baseline based on whether the snapshot's recorded `statePath` matches the current target path, using the prior snapshot state only when paths match and falling back to an empty harness state otherwise [d06ef04]
> - Extended the `harnessSnapshots` module-level WeakMap value type to include a `statePath` property, introduced `resolveHarnessStateTarget` utility function for computing canonical target paths by realpath-resolving parent directories, and updated `loadHarnessState` and `saveHarnessState` to record resolved state paths in snapshots [d06ef04]
> - Added regression test case verifying that unchanged entries are copied when a loaded harness state is saved to a different directory [d06ef04]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 4744ae2.</sup>
> 
>


packages/coding-agent/src/core/refinement/refinement.ts
```diff
@@ -275,7 +278,12 @@ export function loadHarnessState(
 	harnessStateDir: string = getGlobalHarnessStateDir(),
 	scope: HarnessScope = "global",
 ): HarnessState {
-	const statePath = getHarnessStatePath(harnessStateDir);
+	const state = readHarnessState(getHarnessStatePath(harnessStateDir), scope);
+	harnessSnapshots.set(state, { state: structuredClone(state), scope });
+	return state;
+}
+
+function readHarnessState(statePath: string, scope: HarnessScope): HarnessState {
```
snimu, line 286: The TS side still wipes a legacy NaN-poisoned harness file: readHarnessState JSON.parse-fails, degrades to empty state, and the next refinement save silently overwrites the whole store with that empty state. The recovery added here lives only in the Python writer, so whichever writer touches the file first decides whether the user keeps their harness — the exact legacy scenario this PR fixes on the other writer. Wiring the same recovery into the TS reader — with a cross-writer round-trip test (poisoned file → TS load/save → Python load, and the reverse) so the two recoveries can't drift — would make the outcome writer-independent.

packages/coding-agent/src/core/refinement/harness-persistence.ts
```diff
@@ -0,0 +1,69 @@
...
+			throw new Error("Harness refinement history changed before save. Reload and retry.");
+		}
+		merged.refinements.push(...structuredClone(proposed.refinements.slice(baseline.refinements.length)));
+	} else {
+		if (!isDeepStrictEqual(baseline.refinements, latest.refinements)) {
+			throw new Error("Harness refinement history changed before save. Reload and retry.");
+		}
+		merged.refinements = structuredClone(proposed.refinements);
+	}
+	if (proposed.schema !== baseline.schema) {
```
snimu, line 64: Both readers accept any finite schema number, so a future-schema file loads fine — and an old Python load->save then strips its unknown entry fields and kinds while keeping the new schema number, leaving the file silently corrupted yet still stamped with that newer number. A `schema > known` guard that preserves the file untouched or refuses the save would close this. The round-trips are also asymmetric — TS preserves unknown entry fields while Python strips them — worth aligning while the schema handling is open.

xeophon (approved):

PR #2135, opened 2026-09-08
stop unresponsive headless python kernels after cancellation

kevinjosethomas:
Cancelling a headless Python call could return while an unresponsive CPython process kept consuming CPU and blocked later calls. The manager now kills it after the interrupt grace period. It stays dead; a fresh kernel starts only if the continuing agent later requests Python. Cooperative interrupts preserve the process and state, and interactive sessions retain their wait/kill choice.

- Bound cancellation diagnostics to the caller's output limit and report truncation. Windows orphan cleanup performs identity checks and tree kills asynchronously so it cannot block daemon cancellation.
- Validated `e131805b883995dc61c4fcc8ad6ee3c14ba0abe1`: 78 focused tests and `npm run check` passed. Installed-release tests with CPython 3.11.16 covered output limits, termination without a later call, later recovery, and cooperative state preservation.
- A real Prime Inference sub-agent ran a CPU-bound Python cell at 95.8% CPU. Deleting it stopped the process in 5.1 seconds; 38 checks over 10.09 seconds found no replacement kernel or surviving child. Windows subprocess responses were simulated; no native Windows run or 50-minute soak was performed. The live preview used the same required GLM model with documented model configuration to omit a provider-rejected optional `enable_thinking` parameter.
- Included shared check corrections: complete Linear ticket identifiers, Ubuntu-only apt sources, and compatibility with the harness context preceding a literal user prompt. Current-head CI passed after rerunning one unrelated trace-test temporary-directory cleanup failure; Bugbot and Macroscope completed review with no new issues.

Linear: [ENG-6008](https://linear.app/primeintellect/issue/ENG-6008/headless-python-cancellation-leaves-a-busy-kernel-consuming-cpu)

---

> [!NOTE]
> **Medium Risk**
> Changes kernel kill/shutdown timing and process reaping on all platforms; incorrect `liveKernels` or orphan cleanup could leak processes or delay teardown (Windows reaper timeouts cap at ~10s per step).
> 
> **Overview**
> Fixes **ENG-6008**: headless IPython cancellation could return while a wedged CPython process kept burning CPU and blocking later tool calls.
> 
> **Headless abort behavior** adds `ExecuteOptions.killOnAbortTimeout`. After the existing interrupt grace period, the manager resolves the cell as **aborted**, appends a **bounded** “kernel was killed…” diagnostic (respecting `maxOutputChars` and the truncation notice), and **`SIGKILL`s** the kernel. The IPython tool enables this when `!ctx?.hasUI`; interactive sessions still get wait/kill prompts, and cooperative interrupts keep the process when the cell actually stops.
> 
> **Teardown / orphan cleanup** splits resource release from journal reaping: `cleanupResources` is async, dedupes per-kernel work via `orphanCleanupTasks`, and **Windows** identity checks + `taskkill /T` run through non-blocking `execFileHidden` (sync path only on `disposeSync`). `liveKernels` removal is deferred so stale exit cleanup does not drop a replacement kernel still running.
> 
> Regression and Windows-simulated tests cover output limits, cancellation without a follow-up call, and shutdown/kill joining pending orphan reaps.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2e2f0780ca60a351b63b0ea83b0278d87e49da75. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Kill unresponsive headless Python kernels after abort timeout and harden orphan cleanup
> - Headless (no-UI) IPython executions now pass `killOnAbortTimeout` in `ExecuteOptions`; when an interrupted cell does not stop within the abort grace period, `ReplKernelManager.execute` resolves the execution as aborted, appends a bounded kernel-killed diagnostic, and asynchronously kills the kernel instead of leaving it stuck.
> - `chooseBusyKernelAction` now selects kernel replacement (not cancellation) for headless callers that encounter a still-busy kernel after interrupt; interactive contexts still offer wait/kill choices.
> - Orphan-process teardown is reworked in [repl-manager.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/2135/files#diff-787a4ef333fe6eca0d80d5f5a8a4f93970d4f0716e6e43a446241a1dcf488097) and [orphan-process-journal.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/2135/files#diff-31e05b95e0bcdcd7d6020871664b89d5ede314e555ae4166c109276a28f66229): cleanup is asynchronous on Windows, guarded by process-start-time identity verification, and shared via `orphanCleanupTasks` so concurrent `kill`/`shutdown`/`disposeSync` paths join the same per-pid cleanup. A synchronous path (`cleanupResourcesSync`) covers process-exit disposal.
> - Windows orphan reaping now uses absolute `System32` PowerShell/taskkill paths with a 10-second timeout and only tree-kills after identity match; journal records are preserved on lookup mismatch or kill failure.
> - Risk: `performShutdown`, `kill`, `killChildToIdle`, and `disposeSync` now wait for or invoke orphan cleanup before removing the manager from `liveKernels` — any hang in the Windows reaper commands (bounded at 10s each) can delay teardown completion.
>
> 
> #### Changes since #2135 opened
>
> - Introduced generation-based conditional cleanup in `ReplKernelManager` to prevent removal from `liveKernels` when a replacement kernel has been started [2e2f078]
> - Added test coverage for generation-based kernel replacement tracking during stale exit cleanup [2e2f078]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 801c1cf.</sup>
> 
>


packages/coding-agent/src/core/kernel/repl-manager.ts
```diff
@@ -1313,7 +1325,7 @@ export class ReplKernelManager {
 			if (pid !== undefined && signaled) recordOrphanProcessState(pid, false);
 			// A killed/crashed kernel cannot run its own shutdown hook, so the host
 			// reaps the bash() process groups it journaled under this kernel pid.
-			if (pid !== undefined) reapKernelOrphanProcesses(pid);
+			if (pid !== undefined) void reapKernelOrphanProcesses(pid);
```
snimu, line 1328: The `void reapKernelOrphanProcesses(pid)` here makes the win32 orphan reap async, but teardown fires it where it can never finish. disposeSync runs under process.on('exit') with a sync-only contract (documented at repl-manager.ts:1647), and the SIGINT/SIGTERM path calls process.exit(130/143) right after graceful shutdown resolves — so taskkill never launches at exit, whereas on main the synchronous reap finished first; the journal has no startup replay, so a missed reap is never retried. Today the gap is masked because the owned-worker frontend reaps synchronously at worker exit (owned-session-worker.ts:290-308) — an unacknowledged dependency. Awaiting can't fix the exit path (nothing can be awaited under process.on('exit')): keeping the reap synchronous on the exit path specifically — blocking is free in a dying process, and the non-blocking motivation only applies to cancellation paths where the kernel lives on — would restore the guarantee; replaying the journal for dead kernels at next startup would additionally cover host crashes, which even the synchronous version missed.

xeophon (approved):

PR #2140, opened 2026-09-09
1. default to compiled agent installs on macos and linux

kevinjosethomas:
This is part of the PR stack migrating us to use Bun to compile Prime Agent for releases. This makes supported installations self-contained, removing the need to download and install Node.js and npm packages separately.

This PR changes the installer to use the compiled version by default, so new users on supported Macs and Linux machines get the self-contained Prime Agent. The installer checks the download and confirms it runs before making it active. Machines that can't run it automatically use the existing Node.js installation.

Tested fresh installs and failed or interrupted downloads across Debian 12, Ubuntu 18.04/20.04/22.04/24.04, and Alpine 3.24. CI also tested compiled packages on macOS 15 and Ubuntu 24.04, on x64 and ARM64. Actual macOS 13/14 execution remains untested.

[ENG-6027](https://linear.app/primeintellect/issue/ENG-6027) · Stack: **#2140** → [#2165](https://github.com/PrimeIntellect-ai/prime-agent/pull/2165) → [#2166](https://github.com/PrimeIntellect-ai/prime-agent/pull/2166). Ship all three together.

---

> [!NOTE]
> **High Risk**
> This changes the default install and release distribution path for all supported macOS/Linux users, including download security rules and managed filesystem state, so regressions affect production installs directly.
> 
> **Overview**
> Adds **standalone release archives** for four platforms and wires them through CI and the release pipeline. A new reusable workflow compiles with Bun, assembles tarballs, tests the extracted app without Node/npm/Bun on `PATH`, and uploads artifacts; **CI** and **Release Prime Agent** now depend on that job and pack/publish the verified archives alongside existing npm tarballs.
> 
> **`install.sh`** now defaults to a **managed native install** on macOS 13+ and glibc Linux (arm64/x64): HTTPS-only downloads (strict curl proto rules), checksum validation, safe archive extraction, executable probe, serialized `.install-lock`, and `bin/previous` rollback. It falls back to the Node/npm path in `auto` when the release is npm-only or the binary cannot run; `binary` mode does not fall back. Native installs can run **`prime-agent --prime-agent-bootstrap`** to prepare Python/tools.
> 
> Supporting changes include refactored **`build:binary`** scripts, release packing via **`--binary-dir`**, kernel runtime resolution for the flat executable layout, docs/README HTTPS install lines, and broad installer/artifact tests.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ccbc3c9177d19b33be45053422a4cb1eada16bcf. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Default `install.sh` to compiled agent installs on macOS and Linux
> - `install.sh` now defaults to compiled `prime-agent` binaries on macOS 13+ and Linux glibc 2.17+, falling back to the Node installer only in `auto` mode.
> - Added build scripts and CI workflows to compile, package, and test four standalone platform archives (darwin-arm64, darwin-x64, linux-arm64, linux-x64).
> - Installer release downloads now enforce HTTPS for base URLs and redirects, with a narrowly scoped loopback exception for local tests.
> - Added `--prime-agent-bootstrap` CLI command to prepare the Python kernel and tools; the kernel now checks the `prime-agent-runtime` sidecar first.
> - Risk: `install.sh` in `binary` mode now exits when a native platform is not available, instead of falling back to Node.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized ccbc3c9.</sup>
> 
>


snimu:
`darwin release binaries`: A user who follows the new browser-download + manual-extract flow gets the binary killed by Gatekeeper on Apple Silicon — the darwin artifacts are unsigned and not notarized, and the browser download leaves the quarantine xattr in place. The curl-installer path keeps working only because it strips/avoids quarantine. The failure therefore appears exactly where the docs now send users who will not use the installer. Please sign + notarize the darwin artifacts; until then, remove the manual-download instructions from the docs and document only the installer path.

snimu:
`checksum verification`: Anyone who controls the download origin — or a MITM on an accepted http:// base URL — ships arbitrary code that passes a perfectly valid checksum: verification is same-origin SHA-256 only, there is no signature or pinning, and the pre-activation probe executes the downloaded binary. The docs calling these downloads 'verified' overstates what the check proves. #2166's update path already does dual-source verification (manifest checksum vs SHA256SUMS); aligning the install path to that standard, plus rejecting http:// base URLs, would close most of the gap.

fettpl:
The first beta produced from the merged commit has a separate failure from Gatekeeper quarantine: the published `darwin-arm64` Mach-O has an invalid embedded ad-hoc signature and is killed even when executed from a local temp directory.

Verified on Apple Silicon, macOS 27.0 (26A428), using beta `0.9.4-beta.675.1.dd760d3`:

- archive SHA-256 is `f290be7e1e26eb319c8b9b96506669885f8c124238a10c329caeb0c5855d938c`, matching `beta.json`;
- extracted `prime-agent --version` exits 137;
- `codesign --verify --deep --strict --verbose=4 prime-agent` reports `invalid signature (code or signature have been modified)` for arm64;
- copying the executable and applying a fresh ad-hoc signature makes strict verification pass, after which `--version` prints `0.9.4-beta.675.1.dd760d3` and `--help` exits 0.

This means the macOS auto installer falls back to Node and forced binary/browser-archive use fails before Gatekeeper notarization is considered. @kevinjosethomas, please add strict `codesign` verification against the final archived artifact and publish a valid final signature before promoting this path to stable. The evidence establishes signature invalidity after publication; it does not yet isolate which build or archive step changed the signed bytes.

kevinjosethomas:
@fettpl thanks for reporting, nice to see people testing nightly! I'll get this fixed soon, sorry about the inconvenience!

PR #2165, opened 2026-09-09
2. migrate existing npm installs to the compiled agent

kevinjosethomas:
This is part of the PR stack migrating us to use Bun to compile Prime Agent for releases. This makes supported installations self-contained, removing the need to download and install Node.js and npm packages separately.

This PR moves existing users to the compiled version after they update and open Prime Agent again. Settings, credentials, and conversations stay intact. Unsupported machines continue using Node.js, and failed migrations leave the existing installation working. The old Node.js installation files are kept for now.

Tested this by simulating upgrades from ten previous releases (0.7.2 through 0.7.4, 0.8.0 through 0.8.1, and 0.9.0 through 0.9.4) to a simulated Bun-compiled release, including skipping straight to a later release. Also tested Alpine, which has no supported compiled package in this stack and automatically stayed on Node.js. An existing npm 12 restriction requires a documented recovery command.

[ENG-6046](https://linear.app/primeintellect/issue/ENG-6046) · Stack: [#2140](https://github.com/PrimeIntellect-ai/prime-agent/pull/2140) → **#2165** → [#2166](https://github.com/PrimeIntellect-ai/prime-agent/pull/2166).

---

> [!NOTE]
> **Medium Risk**
> Changes the default npm CLI entrypoint and can download artifacts and replace the user's public command symlink; failures are designed to fall back to Node, but botched handoff or races could briefly affect how `prime-agent` is invoked.
> 
> **Overview**
> Adds an **npm-to-native migration bridge** so users on conventional global npm installs can move to the compiled agent after updating, without losing settings or breaking launches when migration cannot complete.
> 
> Release packing (when `--binary-dir` is set) makes **`dist/bundle/cli.js` the bridge**, renames the real Node CLI to **`cli-node.js`**, and ships **`install.sh`** plus **`native-release.json`** in the npm tarball. On supported macOS/Linux global installs whose `bin` link still points at the package, the bridge can **download/activate** a managed release via the existing installer (foreground interactive only), **reuse** an equal-or-newer managed build after validation, and **hand off** the public command symlink to the native launcher while keeping Node as fallback. Daemon workers, `--help`/`--version`, piped/JSON runs, offline mode, unsupported hosts, competing npm installs, and failed/cancelled migrations are explicitly excluded or deferred.
> 
> Adds **`readNativeInstallation`** helpers, a long **“Migration from npm”** doc section (including npm 12 `EALLOWREMOTE` recovery), changelog notes, a large **`npm-native-bridge`** test suite, and small **agent-traces** test waits so uploads finish logging before assertions.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a6ecd39e129762c4abbca36ca165c32f12c7ed8b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Migrate global npm installs to the compiled native agent launcher
> - Adds `migrationTarget` in [npm-native-bridge.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/2165/files#diff-5d9f7699f3823261a2d3acee1e93b66d7bf1dfd1fa1effddbc56388989b53613): for eligible interactive foreground launches on Darwin/Linux conventional global npm layouts, it downloads or reuses a compatible managed native release, validates platform and executable version, and hands off the public command link to the native launcher.
> - Falls back to the bundled Node CLI (`cli-node.js`) when migration is skipped or fails; opt-outs, offline mode, unsupported hosts, daemon/worker/version/help launches, and competing command owners all stay on Node.
> - Suppression is based on 24-hour timestamp marker files so failed migrations and probe diagnostics don't retry every launch; explicit retries bypass the window and cancellation removes the marker.
> - Adds `readNativeInstallation` and `getNativeInstallation` in [native-installation.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/2165/files#diff-a2ab020031a9ef348417ba499366f4f1867fcf4750555b15c4989be7b7b4bea5) to validate managed installation layout, launcher target, checksum marker, release metadata, platform, and source URL before accepting it.
> - Updates [pack-prime-agent-release.mjs](https://github.com/PrimeIntellect-ai/prime-agent/pull/2165/files#diff-9dd43bd6a4a36a4d005682430f238282f0f2144bc51ed40c2159f18056ec751c) so binary packages install the bridge as `cli.js` and retain the Node CLI as `cli-node.js`, shipping the installer and release metadata alongside.
> - Risk: foreground migration now spawns a detached installer (450s timeout) and forwards its output to stderr; if the shipped installer or release feed is misconfigured, users on conventional global npm installs will see stderr diagnostics and remain on Node until the 24-hour suppression window expires. Command-link handoff is exclusive with restoration on races, but a competing unrelated regular file at the command path is never overwritten.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized a6ecd39.</sup>
> 
>


snimu:
`npm-native-bridge.ts`: First launch after an update looks hung for up to 7.5 minutes with zero output, because the bridge runs install.sh synchronously on any non-internal invocation — including --version, --help, and automation/scripts — with its output piped and discarded. A Ctrl-C during it writes the retry file, which locks migration out for 24 hours. The internal skip list also misses owned-session workers and the interactive update child, so background workers can trigger or inherit the silent download.

snimu:
`--native-platform / --version probe contract`: Any drift in either probe output — install.sh --native-platform must print exactly 'os-arch' and the compiled binary's --version must print bare semver — silently keeps every npm user on Node forever: a probe mismatch is a permanent quiet deferral, with no error anywhere. No test pins the real outputs; the test stub prints the expected strings by construction. The producer side has matching sharp edges in #2140's install.sh: --native-platform prints without a trailing newline and exits 1 silently on unsupported platforms.

PR #2203, opened 2026-09-10
fail kernel bootstrap fast when the bundled runtime source is missing

kevinjosethomas:
- a session that outlived its deleted worktree could no longer find the bundled runtime source, fell back to installing an unpublished package name, and wiped the shared kernel venv that every other session was using before the reinstall failed with a bare exit code.
- bootstrap now resolves the runtime source before looking at the venv and fails with a clear message naming the searched paths and the recovery steps, leaving the existing venv untouched; the registry fallback is gone since the runtime is always installed from a local checkout.
- uv and python helper failures now carry the tail of their stderr, and a new runtime source override lets you install from an explicit checkout and doubles as the test seam for the missing-source path.

---

> [!NOTE]
> **Medium Risk**
> Changes shared kernel venv bootstrap ordering and install path resolution; incorrect behavior could still break multi-session Python kernel setup, though the intent is to reduce collateral damage when installs go missing.
> 
> **Overview**
> **Kernel bootstrap no longer destroys the shared venv when the bundled `prime-agent-runtime` checkout is gone** (e.g. a long-lived session after its worktree was removed). Bootstrap **resolves the runtime source directory before** comparing or rebuilding `~/.prime/agent/kernel-venv`, and **drops the old registry-name fallback**—installs always come from a local path with a content hash identity.
> 
> When no source is found, bootstrap **fails immediately** with searched paths, recovery hints (`PRIME_AGENT_RUNTIME_SOURCE`, restart from a valid install, or `PRIME_AGENT_KERNEL_PYTHON`), and an explicit note that the **existing venv was left untouched**.
> 
> **`PRIME_AGENT_RUNTIME_SOURCE`** lets devs point bootstrap at a specific runtime checkout (also wired into the bootstrap cache key). **`uv`/`python` helper failures** now append a bounded **stderr tail** instead of only an exit code.
> 
> Docs add the env var and a new RLM failure-mode row; tests cover override install, missing-source safety, and stderr propagation (Windows spawn expectations updated for stderr piping).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 147fc295215d8adcf6f5f5797a5d8cae7a6ec56c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Fail kernel bootstrap fast when bundled runtime source is missing
> - Kernel bootstrap now resolves and hashes the local runtime source before touching the shared venv. If no source is found, it throws a structured error listing searched paths and leaves the existing venv untouched.
> - Adds `PRIME_AGENT_RUNTIME_SOURCE` env var so callers can point bootstrap at a specific runtime checkout; when set, it is the sole candidate and the cache key changes accordingly.
> - Removes the registry-package fallback from `bootstrapVenv`; installs always use the resolved local source directory.
> - Subprocess helper `run` now captures stderr (up to a fixed tail) and appends it to failure errors instead of reporting only the exit code.
> - Risk: `resolveRuntimeIdentity` no longer returns a package-name fallback identity — any out-of-tree caller relying on that fallback will now fail when the local source is absent. Bootstrap subprocess spawn options changed to pipe stderr, so tests asserting stdio config in [kernel-windows-process.test.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/2203/files#diff-db1356e4f0322906d7eea82aba4053a2bb4193e6879a3bd43a0bbd34deb494ee) were updated.
>
> 
>
> <details>
> <summary>📊 <a href="https://app.macroscope.com">Macroscope</a> summarized 147fc29. 4 files reviewed, 1 issue evaluated, 0 issues filtered, 1 comment posted</summary>
>
> ### 🗂️ Filtered Issues
>
> </details>
>

### DIFF
--- a/packages/coding-agent/README.md
+++ b/packages/coding-agent/README.md
@@ -685,6 +685,7 @@ prime-agent --thinking high "Solve this complex problem"
 | `PRIME_AGENT_TRACES_API_KEY` | Prime API key used only for opt-in trace sharing |
 | `PRIME_AGENT_TRACES_BASE_URL` | Override the Prime Agent trace upload API base URL |
 | `PRIME_AGENT_KERNEL_PYTHON` | Use an existing Python environment with `prime-agent-runtime` instead of auto-bootstrapping `~/.prime/agent/kernel-venv` |
+| `PRIME_AGENT_RUNTIME_SOURCE` | Install `prime-agent-runtime` into the kernel venv from this source checkout instead of the copy bundled with Prime Agent |
 | `VISUAL`, `EDITOR` | External editor for Ctrl+G |
 
 The remaining `PI_*` variables in this table are compatibility names still read by the current runtime. They do not change the application name, command, or default `~/.prime/agent` configuration path.
--- a/packages/coding-agent/docs/rlm-runtime.md
+++ b/packages/coding-agent/docs/rlm-runtime.md
@@ -237,6 +237,7 @@ Provider credentials are resolved by the TypeScript host. The bounded model cata
 | Failure | Behavior |
 |---|---|
 | Managed runtime is missing | Kernel bootstrap rebuilds it; a custom `PRIME_AGENT_KERNEL_PYTHON` without a current `prime-agent-runtime` is rejected at kernel startup. |
+| Bundled runtime source is missing | Kernel bootstrap fails immediately, names the paths it searched, and leaves the existing venv untouched. This happens when the Prime Agent install a running process started from is deleted or moved (for example a removed worktree); restart from an intact install or set `PRIME_AGENT_RUNTIME_SOURCE`. |
 | Depth limit reached | The host rejects the spawn request; the error reply raises in Python. |
 | Unsupported options | Host rejects the request. |
 | Requested model unavailable | Spawn fails instead of substituting another model. |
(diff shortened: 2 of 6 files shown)


jimmyarphs2:
Confirmed against current head `147fc295`: this race is still reachable. The runtime source and hash are captured before `acquireBootstrapLock()`, then reused after the wait, so a removed source can still lead to teardown of the shared venv.

The minimum correction is to refresh both values immediately after acquiring the lock and use them for the locked readiness checks and rebuild:

```diff
-	const runtimeSourceDir = await requireRuntimeSourceDir();
-	const runtimeIdentity = await hashRuntimeSource(runtimeSourceDir);
+	let runtimeSourceDir = await requireRuntimeSourceDir();
+	let runtimeIdentity = await hashRuntimeSource(runtimeSourceDir);
 ...
 	const releaseLock = await acquireBootstrapLock(venv);
 	try {
+		runtimeSourceDir = await requireRuntimeSourceDir();
+		runtimeIdentity = await hashRuntimeSource(runtimeSourceDir);
 		if (await kernelReady(python, venv, runtimeIdentity, pythonSkills)) return python;
```

I also tested a regression that holds `kernel-venv.bootstrap.lock`, starts bootstrap, removes `PRIME_AGENT_RUNTIME_SOURCE` while bootstrap is waiting, releases the lock, and verifies bootstrap rejects without deleting a preserved file in the existing venv.

Verification on a branch rebased onto current `main`:

- `test/kernel-bootstrap.test.ts`: 26/26 passed
- `npm run check`: passed, including formatter, typecheck, installer check, and browser smoke check

PR #2210, opened 2026-09-11
fix(daemon): re-arm the herdr reporter when a pane attaches a session

kevinjosethomas:
## Problem

The built-in herdr reporter arms only at extension load, with `HERDR_*` env in scope (`withClientEnv` window). A daemon-resident session created headless or in a pane that has since closed keeps its stale identity after being attached from another pane, so it never shows in herdr's agents sidebar. The only workaround is a manual `/reload`, which the TUI refuses while a response is executing.

## Proposal

- Attach now rebinds the session's pane identity to the attaching client (last pane wins). `rebindClientEnv` overwrites `state.clientEnv` and propagates to subagents spawned before the rebind, so `pi.exec` (live read) and every runtime rebuild/reload keep load-time captures and subprocess env consistent with the new pane. The old adopt-if-absent contract remains for create-reuse.
- The already-loaded reporter cannot re-run its factory, so the daemon also emits `herdr:rebind` on the session's shared extension bus (new `ResourceLoader.emitExtensionEvent`). The reporter now registers its listener set even when unarmed and (re)arms its socket identity from the event, claiming the new pane with the live working/idle state immediately.
- Watchers (agents view, subagent viewers) and headless clients attach without env and never move the identity; plain detach leaves the last pane (herdr drops reports for panes it does not know).
- The supervisor no longer fabricates its own env for env-less worker attaches. That fallback adopted the daemon-starting pane onto env-less sessions via watcher attaches and would also have re-bound the reporter to a random pane.

Wire protocol is unchanged (attach already carried optional env), so this is backward-compatible; the env contract comments in `daemon-protocol.ts` and `daemon-mode.ts` are updated to the new semantics.

## Verification

- `npm run check` passes (biome, tsgo, linear-ticket, installer, browser-smoke).
- `npx tsx ../../node_modules/vitest/dist/cli.js --run test/herdr-agent-state.test.ts test/daemon-mode.test.ts test/daemon-client-env.test.ts test/daemon-extension-binding.test.ts test/daemon-supervisor-monitor.test.ts test/agent-connection-daemon.test.ts test/sdk-skills.test.ts`
- New tests: the reporter arms from a `herdr:rebind` event and reports live state, switches panes on later rebinds (last attach wins), ignores rebinds without a complete identity, and unsubscribes on shutdown; a daemon attach with `HERDR_*` env rebinds the session env and notifies extensions, an env-less attach changes nothing, and subagents inherit the rebind.
- The 4 daemon-mode passive-subagent and 4 resource-loader failures that reproduce in this environment also fail on pristine `origin/main` (pre-existing, unrelated).

---

> [!NOTE]
> **Medium Risk**
> Changes daemon attach semantics for Herdr pane identity and extension bus events; scope is integration/daemon attach paths with new test coverage, not auth or data handling.
> 
> **Overview**
> Fixes daemon-resident sessions that never showed in Herdr’s agent sidebar after being opened from a pane, because the built-in reporter only captured pane identity at extension load.
> 
> **Attach now rebinds pane identity** (`rebindClientEnv`): env-carrying attaches overwrite `clientEnv` with the attaching client’s allowlisted `HERDR_*` vars (last pane wins), propagate to subagents, and emit **`herdr:rebind`** on the session extension bus via new **`ResourceLoader.emitExtensionEvent`**. Env-less attaches (watchers, headless) leave identity unchanged. Session **create** still uses adopt-if-absent when reusing env-less sessions.
> 
> **Built-in herdr reporter** stays loaded even without Herdr env at load: it tracks working/idle/blocked while unarmed, skips socket traffic until armed, then **re-arms on `herdr:rebind`** and publishes live state to the new pane. Shutdown unsubscribes the rebind listener.
> 
> **Supervisor** no longer injects its own env on env-less worker attaches (which could wrongly bind sessions to the daemon’s pane).
> 
> Wire protocol unchanged; tests cover rebind, env-less attach, subagent propagation, and reporter behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit edeb993226ae94d433f221476a1e45f52a00556f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/coding-agent/.changes/herdr-reporter-rearm.md
+++ b/packages/coding-agent/.changes/herdr-reporter-rearm.md
@@ -1,0 +1,1 @@
+- Fixed the herdr agent indicator to follow the pane that attaches a daemon session, so a session opened from a pane appears in herdr without a manual `/reload`.
--- a/packages/coding-agent/examples/sdk/12-full-control.ts
+++ b/packages/coding-agent/examples/sdk/12-full-control.ts
@@ -48,6 +48,7 @@ Available: ipython. Be concise.`,
 	getAppendSystemPrompt: () => [],
 	extendResources: () => {},
 	reload: async () => {},
+	emitExtensionEvent: () => {},
 };
 
 const { session } = await createAgentSession({
--- a/packages/coding-agent/src/core/extensions/builtin/herdr-agent-state.ts
+++ b/packages/coding-agent/src/core/extensions/builtin/herdr-agent-state.ts
@@ -12,8 +12,10 @@
  * loader invokes per session load — inside the daemon's client-env window —
  * so each daemon session captures its own pane identity.
  *
- * The factory is a complete no-op when `HERDR_ENV` is not `"1"` (i.e. when
- * not running inside a Herdr pane), so it is safe to always load.
+ * The factory only opens sockets when armed with a full Herdr pane identity,
+ * so it is safe to always load. Outside Herdr it stays unarmed but keeps
+ * tracking session state, so the daemon can arm it later when a pane client
+ * attaches (see `HERDR_REBIND_EVENT`).
  */
 
 import { createConnection } from "node:net";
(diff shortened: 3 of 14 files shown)


kevinjosethomas:
Closing tentatively for now — branch fix/herdr-reporter-rearm stays pushed; reopen whenever ready.

PR #2230, opened 2026-09-11
fix(side-question): keep the session thinking level for /btw

snimu:
No-Ticket: side-question cache cost fix, discussed directly with snimu

`/btw` side questions clone the main conversation and send it to the model again. Today the side request also drops the thinking level to the cheap auxiliary level. On Anthropic-style APIs the thinking parameters are part of the message-level cache key, so a different level means the cloned conversation does not hit the cache: the first `/btw` on a long session re-reads the whole conversation at full input price instead of the much cheaper cache-read price.

This change sends the session's own thinking level for side questions, so the cache key for the cloned messages matches the main thread's requests.

Notes:
- No preference clamp is re-added. Providers already clamp the level per model at request time, exactly as they do for the main thread, so an unsupported level cannot be sent.
- `getAuxiliaryThinkingLevel` stays; refinement and auto-refine review still use it.
- Model-invisible change: the side thread still sees the same prompt, messages and (empty) tool list as before.

Test: `6010-auxiliary-reasoning.test.ts` now pins that the side question's request uses the session level while refinement and review keep the auxiliary level.

Validation (Prime sandbox): `npm run check` clean; `4509-side-questions.test.ts`, `6010-auxiliary-reasoning.test.ts`, `daemon-supervisor-side-question.test.ts` - 49 tests pass.

---

> [!NOTE]
> **Low Risk**
> Behavioral change to `/btw` cost and tool handling only; main session tools never execute in side threads, and reasoning may cost more than the old auxiliary level while cache savings offset input on long sessions.
> 
> **Overview**
> **`/btw` side questions now mirror the main session’s reasoning and tool declarations** so provider prompt caches (Anthropic-style keys on thinking params and tool prefixes) hit on the cloned conversation instead of forcing a full re-read at input price.
> 
> Side agents inherit the parent **thinking level** (no auxiliary clamp) and the parent **tool list**, while **`beforeToolCall` blocks every execution** with a fixed error message. The first-turn instruction explains `/btw`, deactivated tools, and that nothing feeds the main session. Runs can continue up to **three turns** when the model keeps attempting tools; completion picks the **last assistant message** in the side thread, and streaming keeps partial text when a turn is tool-only.
> 
> Regression coverage adds tool-declaration/blocking and turn-cap cases in `4509-side-questions.test.ts`, and `6010-auxiliary-reasoning.test.ts` expects auxiliary levels only for refinement/review—not side questions.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7289eb3ffac92aa9a105c05b995de45b1fc8fc97. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Keep session thinking level and tool declarations for `/btw` side questions
> - Side questions now use the parent session thinking level instead of an auxiliary lookup, so provider cache keys match the main conversation
> - Side questions receive the parent tool declarations but block every tool call through `beforeToolCall`, returning a fixed blocked-tool reason without executing tools
> - Side runs stop after three consecutive tool-call turns or when the assistant produces a non-tool response, whichever comes first
> - Answer updates now ignore assistant turns that contain no extracted text, so accumulated text is no longer wiped by tool-only updates
> - Behavioral Change: `sideQuestion.startSideQuestion` in [side-question.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/2230/files#diff-854e565ab4c061597eb4ed5ef465f9735fd0b98ef7186bc8424f1df97bd18265) sends tool declarations to the provider and blocks execution; first-turn instruction wording is updated; repeated tool-request turns are bounded by a three-turn maximum
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 7289eb3.</sup>
> 
>

### DIFF
--- a/packages/coding-agent/.changes/btw-side-question-thinking.md
+++ b/packages/coding-agent/.changes/btw-side-question-thinking.md
@@ -1,0 +1,1 @@
+- Fixed `/btw` side questions discarding the main conversation's prompt cache by lowering the reasoning level: side questions now keep the session's thinking level, so providers whose cache keys include thinking parameters reuse the cached conversation.
--- a/packages/coding-agent/.changes/btw-side-question-tools.md
+++ b/packages/coding-agent/.changes/btw-side-question-tools.md
@@ -1,0 +1,1 @@
+- Changed `/btw` side questions to declare the session's tools without allowing their use: the request now matches the main conversation's cached prefix byte for byte (tools included), any tool call gets an error result instead of executing, and the side thread's first turn explains that it is a `/btw` side conversation with tools deactivated.
--- a/packages/coding-agent/src/core/side-question.ts
+++ b/packages/coding-agent/src/core/side-question.ts
@@ -6,7 +6,6 @@ import {
 	type ProviderRetryPolicy,
 } from "./provider-retry.js";
 import { unwrapSemanticEdgeStreamFn } from "./semantic-edges.js";
-import { getAuxiliaryThinkingLevel } from "./thinking-levels.js";
 
 export type SideQuestionStatus = "running" | "complete" | "cancelled" | "error";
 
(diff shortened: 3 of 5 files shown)


snimu:
Closing to rebuild the stack in the intended order: the tools/cache fix must point at main, with the thinking-level change stacked behind it for independent discussion. GitHub's stack tracking blocked the base swap (and auto-marked #2231 merged into its base branch mid-reorder — no code reached main). Same commits, new PRs incoming.

PR #2243, opened 2026-09-11
[RSI] Reject slash-command typos with a suggested correction

sethkarten:
## What

A typo'd slash command — `/resuem`, `/compaq` — matched no registered surface (builtin, extension command, prompt template, skill), so the literal string went to the model as a user turn: a full round trip burned, the transcript polluted, the reply a confused assistant message. Genuine `/`-prefixed prompts (paths like `/etc/hosts`, URLs) must keep passing through.

## Change

- **Session-level guard** in `_normalizeSubmission`'s final normalization step: when the text parses as a slash command but matches **no** registered surface, and a registered command name is **close enough** (edit distance ≤ max(2, len/3) — the heuristic the CLI already uses for unknown shell commands), the prompt is rejected with `Unknown command: /resuem. Did you mean /resume?`. The interactive prompt error path already restores the editor text, so the user keeps their input to fix the typo.
- **No close match → pass through unchanged**, so path/URL-prefixed prompts are unaffected. This is the false-positive protection: the guard only fires when a known command is plausibly the intended one.
- The pure suggestion helper moved from `cli/command-registry.ts` to `core/slash-commands.ts` (`findSlashCommandSuggestion`) so core can use it without a layering violation; the CLI keeps a delegating re-export.

## Tests

`agent-session-slash-guard.test.ts` (+3, full scripted-session harness): a `/resuem` prompt rejects with the `/resume` suggestion **before any model call** (the stream function is never invoked); a no-near-match `/etc/hosts is where hostname lookups start` prompt passes through literally; a registered template (`/worktree-check`) expands instead of being guarded. 183 prompt-templates/args tests + 47 extensions/interactive tests green; tsgo + biome clean.

Fixes ENG-6102

---
*This PR was authored autonomously as part of a recursive self-improvement program (RSI).*

---

> [!NOTE]
> **Low Risk**
> Behavior change is limited to interactive prompt submission with template expansion; near-miss typos error instead of reaching the model, with explicit pass-through rules to limit false positives.
> 
> **Overview**
> **AgentSession** now blocks likely slash-command typos during prompt normalization (when template expansion runs): if the input parses as `/name` but is not a registered builtin, session command, template, extension command, or `skill:…`, and a registered name is **close enough**, it throws `Unknown command: /…. Did you mean /…?` **before** any model round trip. Genuine `/`-prefixed text with no plausible match, very long first tokens (>64 chars), and registered templates still pass through or expand as before.
> 
> Fuzzy matching is centralized in **`findSlashCommandSuggestion`** in `slash-commands.ts` (moved out of the CLI registry); short command tokens (≤3 chars) only suggest on a **single-character** typo so path-like words (e.g. `/tmp`) are not misclassified. The CLI keeps delegating to the same helper for unknown shell commands.
> 
> New **`agent-session-slash-guard.test.ts`** covers rejection with suggestions, pass-through cases, skill typos, and template expansion.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 85130f4bf2816d2b4d600b97858571c4470b0ad0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Reject mistyped slash commands with a suggested correction in `AgentSession`
> - Adds `AgentSession._throwIfUnknownSlashCommand`, which runs during submission normalization (when prompt-template expansion is enabled) and throws an error with a suggested command when the leading slash token is a close edit-distance match to a built-in, alias, session command, template, extension command, or skill.
> - Centralizes suggestion logic into a shared `findSlashCommandSuggestion` utility in [slash-commands.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/2243/files#diff-1c09ca743d0363fbc0f524e5ca07b9d5ac086e6907790d18a884c6a1957bd051); removes the local `editDistance` helper from [command-registry.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/2243/files#diff-ba6fd689c2d87db9c6afe1cdb902b1644cb862c4434a4dbb41442f7b2ef466a8).
> - Tightens matching for short tokens: inputs of three characters or fewer now allow only one edit, so path-like slash-prefixed prompts pass through to the model unchanged.
> - Long slash-prefixed names (>64 chars), exact built-ins, templates, and valid skills are exempt from the guard.
> - Risk: submissions that previously reached the model as plain prompts (e.g. `/resum` instead of `/resume`) now raise an error before the model call; callers of `AgentSession.prompt` that did not surface submission errors may need to handle the new rejection.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 85130f4.</sup>
> 
>

### DIFF
--- a/packages/coding-agent/.changes/eng-6102-slash-command-typo-guard.md
+++ b/packages/coding-agent/.changes/eng-6102-slash-command-typo-guard.md
@@ -1,0 +1,1 @@
+- Typo'd slash commands now fail fast with a suggested correction instead of being sent to the model as a prompt; genuine messages that merely start with a slash still pass through.
--- a/packages/coding-agent/src/cli/command-registry.ts
+++ b/packages/coding-agent/src/cli/command-registry.ts
@@ -1,4 +1,5 @@
 import { APP_NAME } from "../config.js";
+import { findSlashCommandSuggestion } from "../core/slash-commands.js";
 
 export interface CommandSpec {
 	path: readonly string[];
@@ -295,17 +296,7 @@ export function isHelpCommandRequest(path: readonly string[]): boolean {
 }
 
 export function findCommandSuggestion(input: string, candidates: readonly string[]): string | undefined {
-	let closest: { candidate: string; distance: number } | undefined;
-	for (const candidate of candidates) {
-		const distance = editDistance(input, candidate);
-		if (!closest || distance < closest.distance) {
-			closest = { candidate, distance };
-		}
-	}
-	if (!closest || closest.distance > Math.max(2, Math.floor(input.length / 3))) {
-		return undefined;
-	}
-	return closest.candidate;
+	return findSlashCommandSuggestion(input, candidates);
 }
 
 export function formatTopLevelHelp(): string {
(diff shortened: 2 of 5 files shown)


sethkarten:
Both findings fixed in b8f423a90:

1. Skill typos - candidates now include skill:-prefixed names, so /skill:pythno with a registered python skill rejects with "Did you mean /skill:python?". Regression test added (typo rejected before any model call).

2. Event-loop block - names over 64 characters skip the guard entirely: no registered command is anywhere near that long, so oversized /-prefixed prompts pass through as before without any fuzzy matching. Regression test asserts a 500+-char /path/... prompt reaches the model literally.

sethkarten:
Fixed in 85130f4bf: tokens of three characters or fewer now match only on a single-character edit distance, so "tmp" (two edits from "mcp") no longer triggers the guard while "log" (one insertion from "logs") still suggests it. Longer tokens keep the max(2, len/3) rule. Regression tests pin both: "/tmp notes for the cleanup" reaches the model; "/log rotate policies" rejects with the /logs suggestion. The shared helper gives the CLI's advisory unknown-command notice the same tighter floor.

xeophon (approved):

PR #2295, opened 2026-09-13
fix(coding-agent): prevent crash when leaving a chat during startup

xeophon:
Pressing left arrow while a chat is loading could dispose its daemon connection before startup requested the initial snapshot, crashing the session. Keep the connection alive until initialization finishes, mark shutdown immediately so startup skips onboarding and prompt delivery, and signal the agents-view handoff after teardown completes.

Load optional heartbeat metadata in the background so opening or leaving a chat does not wait for the global saved-session scan. Ignore its result after shutdown. No daemon wire changes.

Validation: `npm run check`; three focused startup-navigation regressions, also passed against both locally installed JavaScript artifacts (six checks). A probe using the real daemon heartbeat request and stubbed unrelated binding services measured 6,232 ms of binding wait before the change and under 1 ms after; the background scan still took 5,252 ms. This is not an end-to-end navigation benchmark.

---

> [!NOTE]
> **Medium Risk**
> Changes interactive session lifecycle, teardown ordering, and input submission guards; back navigation can block on stalled init until explicit shutdown.
> 
> **Overview**
> Fixes races when the user goes **back to the agents view** (or shuts down) while a chat session is still initializing.
> 
> **Startup vs. back navigation:** `run()` keeps the in-flight `init()` on `initializationPromise`. `returnToAgentsView` sets `isReturningToAgentsView`, waits for that promise (without failing on init errors), then tears down—so the daemon connection is not disposed before the initial snapshot path finishes. Startup onboarding, deferred notices, and startup prompt delivery are skipped once shutdown or agents-view return is in progress; prompt submit/restore paths treat back navigation like lifecycle cancel.
> 
> **Heartbeat tray:** Session rebind no longer **awaits** heartbeat catalog refresh; it runs in the background and in-flight refresh work is dropped if the user is leaving or shutting down.
> 
> Adds regression tests for non-blocking rebind, back during stalled catalog/snapshot load, and shutdown/SIGTERM interrupting back while init is stalled.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f293546f13d40d88ca94f1cf368de93e363efd16. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Fix crash when leaving a chat during startup in `InteractiveMode`
> - `InteractiveMode.run` stores the initialization promise before awaiting it.
> - `returnToAgentsView` waits for the initialization promise before teardown, preventing concurrent teardowns.
> - Heartbeat catalog refreshes are fire-and-forget and discarded if shutdown or navigation starts.
> - Prompt submissions are cancelled during back navigation to avoid restoring drafts into a tearing-down editor.
> - Risk: `returnToAgentsView` waits for the initialization promise; a stalled startup delays back navigation until an explicit shutdown or SIGTERM.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized f293546.</sup>
> 
>


xeophon:
Addressed the Codex onboarding finding in 74eacc565. `returnToAgentsView()` marks shutdown before awaiting initialization, and `run()` skips onboarding and startup prompt delivery when that flag is set. The agents-view handoff is signaled only after teardown completes.

The catalog and snapshot startup regressions both use an onboarding promise that would never resolve if called, and assert that onboarding is never entered. All three tests in `interactive-mode-startup-navigation.test.ts` pass.

sethkarten (approved):

PR #2321, opened 2026-09-14
feat(extensions): add cancellable keyed follow-ups

coleleavitt:
## Summary

Adds extension-owned, keyed follow-up admission and cancellation so lifecycle plugins can retract/coalesce pending continuations without touching user or other-extension queue entries.

- `queueFollowUp(key, content, { signal })` returns an action receipt
- `cancelFollowUp(key)` cancels only not-yet-delivering work owned by that extension
- owner identity is ephemeral and isolated across extensions/reloads
- reload/disposal cancels owned pending continuations
- exact-action signal cancellation cannot cancel a newer same-key successor
- ordinary agent-message cancellation semantics remain unchanged

## Why

A production `set_goal_until` incident produced 75 stale continuation nudges after repeated successful clears. The plugin had a stale async-result race, but the extension ABI also offered no keyed cancellation handle once a follow-up was admitted.

## Validation

- 222 passed / 8 skipped targeted lifecycle matrix
- 37 additional extension tests passed
- one unrelated refine-extension failure reproduces at baseline
- `tsgo --noEmit` passes
- Biome passes across 1,039 files
- terminal independent lifecycle review: PASS

## Safety boundary

This does not change existing `sendUserMessage`. New cancellation is extension-owner scoped. Restored actions intentionally lose ephemeral extension ownership and cannot be cancelled by a reloaded instance.

> [!NOTE]
> ### Add cancellable keyed extension follow-ups, RAVO refinement system, and pervasive tracing
> - Extensions can now enqueue one coalesced follow-up per owner/key and cancel it until delivery starts; an aborted admission removes the queued action without affecting other extensions' follow-ups
> - Adds the RAVO continual-harness refinement system: a run service, controller, reducer, archive, authority gate, error-budget ledger, and failure ledger that orchestrate inspect/plan/implement/evaluate/repair loops with `/ravo` slash-command support
> - Introduces a dependency-free tracing surface (`trace-context`, `otlp-span-exporter`) with W3C traceparent propagation and span instrumentation across the agent loop, tools, extensions, daemon, kernel, bash, and REPL; OTLP export installs from `OTEL_EXPORTER_OTLP_ENDPOINT` with bounded shutdown
> - Adds `runAgent` to the extension context API, enabling extensions to launch structured child-agent calls with scoped tools and cancellation propagation
> - Kernel crash recovery now rejects active requests with `KernelExitedError`, surfaces one-time recovery notices, and daemon connections can recreate lost sessions after daemon loss
> - Risk: `ModelRegistry.refresh` now restores sibling registries' providers instead of clearing them; reloading code that depended on the old full-reset behavior may need verification in [model-registry.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/2321/files#diff-2fd8c4a2672b4ce27e73d494479df612c8b30ebac3dda5a160e2d861c612f453). Orphan-process journal reading now fails closed on malformed records in [orphan-process-journal.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/2321/files#diff-31e05b95e0bcdcd7d6020871664b89d5ede314e555ae4166c109276a28f66229). Daemon schema revision advanced to 29; older daemons will not receive RAVO update pushes or deferred search-text requests.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 70198c1.</sup>
> 
>


coleleavitt:
Branch rebuilt onto current upstream main. Diff is now exactly seven files (492 additions/8 deletions), terminal independent review passed 302/302 relevant tests plus current-main typecheck/Biome.

PR #2374, opened 2026-09-16
[RSI, bug] fix(ai): recover stale Codex chains after metadata

sethkarten:
## Problem

`v0.9.5` contains the stale Codex continuation recovery from #2314, but `openai-codex/gpt-5.6-sol` sessions can still surface:

```
Codex error: Previous response with id 'resp_...' not found.
```

Local evidence from the current beta showed six user-visible failures across two sessions. Every rejected ID was a successful response recorded earlier in the same session, and four failures referenced an older response after a newer response had already completed. Async tool, child-message, refinement, and user-message interleavings can therefore invalidate a cached connection-scoped chain. The outer provider retry recovered each turn, but only after the empty zero-token error had already been persisted and shown.

## Root cause

The one-shot recovery from #2314 is gated on `!websocketStarted`, because retrying after assistant output begins could duplicate streamed content. However, `startWebSocketOutputOnFirstEvent` marked the stream started for every mapped WebSocket event, including lifecycle, telemetry, and Codex vendor metadata that `processResponsesStream` does not expose as assistant output.

If `previous_response_not_found` arrives after one of those internal events, the provider incorrectly treats the attempt as user-visible and skips its safe full-context retry. The non-transport Codex API error then escapes to the agent loop.

## Fix

- Delay the public `start` event until the first WebSocket event family that `processResponsesStream` can turn into assistant output or a terminal response.
- Keep lifecycle, telemetry, and vendor metadata internal, so a subsequent stale-continuation rejection remains eligible for the existing bounded retry.
- Clear any provisional `responseId` from the rejected attempt before retrying on a fresh socket with the full request body.
- Preserve the existing safety boundary: once visible assistant output starts, no chain-reset retry occurs.

## Regression coverage

The new test scripts this sequence:

1. Establish a cached continuation.
2. Emit `response.created`, `response.in_progress`, Codex metadata, and timing events.
3. Return `previous_response_not_found`.
4. Verify one fresh full-context retry without `previous_response_id`.
5. Verify the consumer sees exactly one `start`, a final `done`, the recovered response ID, and no SSE fallback.

A second pin covers the `responseId` reset: the retry-failure test now seeds the stale attempt with
`response.created` metadata and asserts the surfaced message carries no dead continuation anchor.

## Validation

- `GITHUB_BASE_REF=main npm run check` — passed (biome, test-policy, tsgo, installer, browser-smoke)
- `cd packages/ai && npx tsx ../../node_modules/vitest/dist/cli.js --run test/openai-codex-stream.test.ts` — 39 passed
- 8 neighbouring `packages/ai` suites (openai-responses x2, openai-completions x3, stream, stream-failure, xai-responses) — 168 passed, pre-existing skips only

## Test-line budget

`scripts/check-test-policy.mjs` caps net added test lines at the added meaningful source lines. The user directive for this audit (2026-09-18) is: "probably needs to remove redundant and unnecessary tests throughout. Remove unnecessary tests and we should have enough." Per the owner policy decision (2026-09-18): "nah dont grandfather them in. we should fix them now".

- Before: `net test additions 50 exceed source additions 13` (red).
- After (`GITHUB_BASE_REF=main`, head `f7574792e`): `Test policy check passed` - net test additions **+2** (74 added / 72 deleted) against 13 added source lines.

Ledger - no vector was dropped:

- **Copy-paste twin consolidation (zero vector loss).** `completes after response.completed even when the SSE body stays open` and `maps response.incomplete to stopReason length even when the SSE body stays open` held byte-identical 58-line fixtures (temp agent dir, encoder, never-closing `ReadableStream`, mocked release/prompt fetch, model, context) and differed only in the SSE payload and the expected stop reason. They are now one `it.each` with two rows, and both test names are preserved through the `$label` template. Probe: mutating `mapStopReason` so `incomplete` returns `stop` fails exactly the `response.incomplete` row while the `response.completed` row stays green, so the table still discriminates both vectors.
- **Wall-clock race removed.** The consolidated case dropped the per-test 1s `Promise.race` timeout; the suite's configured 30s `testTimeout` is the bound. While the race was still present the gate flagged a NEW `wall-clock-timer` violation from the merged body (reported at line 264), which this removal clears - the consolidation does not buy budget with a policy violation.
- **Coverage gap found and closed (dividend).** The second half of this PR's fix (`delete output.responseId` before the chain-reset retry) was unpinned: probe showed removing that line left all 39 tests green. The retry-failure test now seeds the stale attempt with `response.created { id: "resp_stale" }` metadata and asserts `second.responseId` is undefined. Probe: removing `delete output.responseId` now FAILS that test with `'resp_stale'` received, so both halves of the fix are pinned.

---

> [!NOTE]
> **Medium Risk**
> Changes Codex WebSocket streaming, continuation retry gating, and persisted `responseId` behavior—user-visible turn failures and conversation chaining can shift if the visible-event classifier is wrong.
> 
> **Overview**
> Fixes **stale `previous_response_id` recovery** on Codex WebSocket turns when lifecycle, telemetry, or vendor metadata arrives before `previous_response_not_found`.
> 
> The provider now treats only **output-producing** response events (terminal, output, reasoning, content, refusal, function families) as “visible stream started.” Internal events like `response.created`, `response.in_progress`, and Codex metadata no longer emit the public `start` event or block the one-shot **full-context chain reset**. On that retry, any provisional **`responseId` from the failed attempt is cleared** so a dead anchor is not kept.
> 
> Tests add the lifecycle/metadata stale-chain scenario, assert **`responseId` is undefined** when chain-reset retry fails after metadata, and consolidate duplicate SSE terminal-stream cases into one parameterized test.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e1ae595b68ff7bc691b6f9220da39260fb0c0dc7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Fix stale Codex chain recovery to ignore metadata events and clear `responseId` on retry
> - Adds `isCodexVisibleResponseEvent` classifier to distinguish output-producing response events (terminal, output, reasoning, content, refusal, function) from lifecycle, telemetry, and vendor metadata events
> - Renames the first-event wrapper to `startWebSocketOutputOnFirstVisibleEvent` so that only visible response events trigger the assistant-start transition; metadata events no longer mark a WebSocket attempt as visibly started
> - `streamOpenAICodexResponses` now deletes the accumulated `responseId` before retrying a stale continuation, so the failed attempt's metadata is not carried into the next attempt
> - Behavioral Change: `processWebSocketStream` applies visible-event gating before marking assistant output as started; lifecycle and metadata events that previously triggered an assistant-start event no longer do so. Check `startWebSocketOutputOnFirstVisibleEvent` in [openai-codex-responses.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/2374/files#diff-69623a8c4e616b30339182296d1b69ed62a9d7fcce14b744ed6e638ecf357abc) for consumers expecting start on any first event.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized e1ae595.</sup>
> 
>


snimu (approved):

PR #2398, opened 2026-09-16
[RSI, performance] start the daemon catalog on demand

sethkarten:
## What

The daemon supervisor no longer spawns the session catalog child eagerly at boot. The catalog process starts on demand, on the first session-file operation that needs it (agents view roster, `list --all`, saved-session rename/delete/archive, `mark_interrupted`), and stays resident once started.

## Why

Profile of a fresh interactive install on the PR benchmark's Linux x64 sandbox (from the published `memory-*.json` per-process data, main @ 66abc2a60):

| process | RSS |
| --- | --- |
| TUI (compiled bun) | ~297 MB |
| daemon supervisor (bun) | ~287 MB |
| session worker (bun) | ~243 MB |
| daemon catalog (bun, idle) | ~220 MB |
| uv (per-session skill sync) | ~140 MB |
| python kernel | ~27 MB |
| **summed idle RSS** | **~1,194 MB** |

The catalog is a full compiled runtime that sits completely idle when the user stays in a chat: none of its operations (saved-session list/resolve/rename/delete/archive) run during plain chat startup. It is pure idle overhead - about 220 MB, ~18.6% of the summed idle RSS - and its boot also competes with the TUI, supervisor, and session worker during the cold-start window.

`DaemonCatalogClient` already starts on demand inside every `request()`, memoizes the running child, and respawns after a crash, so the supervisor's eager `catalog.start()` at boot buys only spawn latency on the first catalog call. Removing that await makes the catalog lazy with a one-line semantic change; `stop()` already no-ops when never started, and a failed catalog start was already non-fatal at boot.

## Local before/after

Same machine protocol as the benchmark harness (PTY editor-probe readiness, summed RSS at ready+1s, 6 trials): linux x64 container, 4 vCPU, 8 GB, benchmark install fixture with the compiled bun binary, skills pre-synced. Absolute numbers are Rosetta-inflated; deltas are the signal.

| metric | before (main) | after | delta |
| --- | --- | --- | --- |
| idle RSS at ready+1s | 941 MB mean | 752 MB mean | **-189 MB (-20.0%)** |
| cold startup | 5,307 ms mean | 4,387 ms mean | **-920 ms (-17.3%)** |
| warm startup | 2,435 ms mean | 2,333 ms mean | -102 ms (within spread) |

Idle process tree after: TUI (250 MB) + supervisor (213 MB) + session worker (242 MB) + python kernel (40 MB); the catalog process is absent until a session-file operation runs. Verified on demand in the same container: opening the agents view spawns the catalog (`--version` child with the internal catalog env), the saved-session roster renders, and the catalog stays resident.

## Expected benchmark movement (informational)

- **Idle memory, summed RSS: ~-220 MB (~-18.6%)** on the main baseline of 1,194 MB - the idle catalog process disappears from the benchmark's fresh-chat sessions.
- **Cold startup: expected improvement** - the catalog bun boot no longer runs inside the cold-start window competing for the 4 vCPUs (local delta -17% under emulation; native x64 should be a few hundred ms).
- **Warm startup: ~unchanged** (no daemon boot in the warm path).
- **Install / bundle / disk / kernel metrics: unchanged.**
- **UI metrics: first agents-view open per daemon lifetime pays the catalog spawn** (~300-500 ms on native x64, once per daemon, not per open); `agents_view`/`agents_roster` may tick up by that amount on their first use per trial daemon, `agents_open` and later opens are unaffected. This is the explicit trade: a one-time spawn cost on a rare operation against ~19% resident idle memory at all times.

## Related finding (not addressed here)

The idle snapshot also consistently catches `uv pip install --editable` re-running the bundled Python-skill sync inside the measurement window (main re-runs the full 9-skill sync every trial because the benchmark kills the session before the sync completes and the version marker is written). That is worth a separate change (bootstrap the bundled skills at install time, or batch the per-skill uv calls); this PR deliberately stays scoped to the single biggest lever.

## Testing

- New regression test `daemon-supervisor-catalog-startup.test.ts`: supervisor boot does not start the catalog; the first catalog-backed command (`list_saved_sessions`) starts it exactly once; the response surfaces the not-connected gate as expected with the spawn mocked.
- Targeted suites: daemon-supervisor-monitor (117 incl. roster/eviction/heartbeats/catalog-entry), daemon-catalog-startup, daemon-catalog-process, daemon-supervisor-side-question, daemon-supervisor-lazy-subagents, daemon-agent-roster, daemon-supervisor-eviction, daemon-supervisor-heartbeats, daemon-supervisor-input-pause, daemon-deferred-session-frames, rlm-ledger - all green except one pre-existing `daemon-agent-roster` failure ("deletes a top-level saved session..."), verified failing identically on pristine main.
- `npm run check` (biome + tsgo + installer + browser-smoke) passes; pre-commit hook re-ran it.
- `daemon-supervisor-process.test.ts` cannot run locally on this machine (1s daemon-handshake budget under load; hangs identically on pristine main); end-to-end daemon behavior was instead validated on the compiled binary in the linux container across 10+ cold/warm starts plus an agents-view session-file round trip.

## Test-line budget

Fleet test-budget audit (2026-09-18): this PR exceeded the repo test-line budget gate (`node scripts/check-test-policy.mjs`, run by CI's Build and check job: net added test lines over changed test files may not exceed meaningful added source lines; this PR's source side is a single-line deletion, so the budget is 0). The user directive driving the audit: "probably needs to remove redundant and unnecessary tests throughout. Remove unnecessary tests and we should have enough." The owner policy decision authorized executing the cut menu in cost order: "nah dont grandfather them in. we should fix them now".

- Gate vs `main`: 46 net test additions vs 0 source additions (red) -> fold -> +3 vs 0 -> owner-authorized cut -> **net -5 vs 0: green** (0/9 in daemon-catalog-startup.test.ts, +4/-1 in daemon-supervisor-lazy-subagents.test.ts).
- Fold: both on-demand vectors moved into the existing real-boot test "dispatches authenticated peer queries and excludes disconnected workers" (real `supervisor.start()`, the test's existing catalog-start spy, a real client request); the PR's 46-line standalone test file was deleted. Every added line is load-bearing: `expect(catalogStart).not.toHaveBeenCalled()` after boot (boot must not start the catalog child), and `list_saved_sessions` + `expect(catalogStart).toHaveBeenCalledTimes(1)` (first catalog request starts it lazily, exactly once).
- Probe evidence: re-adding the pre-fix boot-time `catalog.start()` line makes the folded test fail on the boot vector ("expected 'start' to not be called at all, but actually been called 1 times").

### Lost-coverage ledger (owner-authorized cut, per vector)

- **"rejects immediately when the catalog exits during startup"** (`daemon-catalog-startup.test.ts`, -9 lines): pinned `DaemonCatalogClient.start()` rejecting when the spawned catalog child exits during startup ("exited during startup" error path). Cheapest option: smallest self-contained catalog-area test; this PR's subject (WHEN the catalog starts) is untouched by it. Surviving partial pin: the sibling "uses the dedicated entrypoint and allows a cold start past five seconds" keeps the start-path mechanics (entrypoint + cold-start tolerance) - the HOW; the two folded vectors keep the WHEN.
- **The standalone `daemon-supervisor-catalog-startup.test.ts` file** (-46 lines): not a coverage loss - both of its vectors are preserved in the fold with live pins (probe above).

---

> [!NOTE]
> **Medium Risk**
> Behavior change: code or flows that assumed the catalog was running immediately after supervisor start must trigger a catalog operation first; first agents/saved-session UI may pay a one-time spawn latency.
> 
> **Overview**
> The daemon supervisor **no longer starts the session catalog child at boot**—the eager `catalog.start()` call during `DaemonSupervisor.start` is removed. The catalog still comes up through `DaemonCatalogClient` on the **first session-file operation** that needs it (e.g. `list_saved_sessions` / agents roster, rename/delete/archive), then stays resident.
> 
> Tests now assert the catalog is **not** spawned on supervisor boot and is started **exactly once** when a catalog-backed client request runs; the standalone startup-failure-during-boot test was dropped as part of test consolidation.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3260539fc098cde6c97c2fa4078f274620b1b130. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Start the daemon catalog on demand instead of at `DaemonSupervisor.start`
> - Removes the awaited catalog startup call from [daemon-supervisor.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/2398/files#diff-18f2d5ec391785bebc6ddf3c1bd3b795e4ed6e02d0aa3852f5bf247d61d012f8) so the catalog no longer launches during supervisor boot.
> - The catalog starts lazily on the first session-file operation (e.g. `list_saved_sessions`), then stays resident.
> - Updates tests to assert `DaemonSupervisor.start` does not start the catalog and that the first session-list request triggers it exactly once; removes the test for an explicit startup exit during boot.
> - Behavioral Change: `DaemonCatalogClient.start` is no longer invoked at supervisor initialization; any code that assumed the catalog was running immediately after `DaemonSupervisor.start` must now trigger a session-file operation first.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 3260539.</sup>
> 
>


kevinjosethomas:
**Second-pass review findings — holding:**

- **Stale head dodges the test budget**: this branch predates `check:test-policy` on current main — it adds 46 test lines against 0 source additions (the source change only deletes a line), which today's gate rejects. The PR head's `package.json` doesn't run the check, which is why its CI is green. Resolve the budget honestly and rerun against current main before merge.
- Mechanism verified real: no import moved — `daemon-supervisor.ts` stops eagerly spawning the whole Bun catalog child, and `daemon-catalog-process.ts` already lazily starts/reuses it on first request. Recovery paths and roster seeding are unaffected.
- Scope caveats: the ~189 MB is summed per-process RSS at editor-ready+1s (double-counts shared pages), so it's narrower than "daemon RSS"; and the new test proves lazy dispatch only (mocks `start()`, expects "not connected") — a real first-use round-trip test would be stronger.

snimu (approved):

PR #2397, opened 2026-09-16
swarm: swarm harness entry kind with state-machine specification (validator + dag compiler)

sethkarten:
## Summary

Implements PR F of the swarm DAG feature ([spec: Swarm DAGs: declarative orchestration in Continual Harness](https://app.notion.com/p/3da72940136f81a88554e6ec7119270e)): a new `swarm` continual-harness entry kind that stores a declarative DAG of subagent nodes, plus its full write-time dry-run validator. **The executor (`rlm.swarm.run` / status / stop) is a follow-up PR and is not included here.**
## State machine model

The swarm specification is now a **state machine** stored in `arguments["machine"]`. The original DAG form (`arguments["dag"]`) stays as sugar and **compiles to machine form at canonicalization**: a compiled dag is a restricted machine — each state entered at most once, forward-only, with guard-less transitions (one per effective dependency edge) and `max_entries: 1`.

```jsonc
{
  "run": {
    "budget_ms": 600000, "failure_policy": "escalate", // fail_fast | continue | escalate (default)
    "max_parallel": 8,                                   // 1..64 (default 8)
    "max_transitions": 100                               // default 10 * states, hard cap 10000
  },
  "states": [                       // 1..1024, unique slug ids, >= 1 entry state
    {
      "id": "reviewing",
      "entry": true,                 // admission enters every entry state
      "max_entries": 4,              // bounded re-entry; further transitions are blocked
      "subagent": "reviewer",        // required unless the state is a wait state (same forms as V1)
      "lifecycle": "task",           // task | resident
      "inputs":  [{ "name": "pr_url", "type": "text", "from": "entry.pr_url" }], // binds the LATEST settle of the source
      "outputs": [{ "name": "verdict", "type": "json" }],
      "budget_ms": 300000, "retries": 2, "failure_policy": "escalate",
      "foreach": { "over": "items", "max": 64 },   // expands per entry
      // "wait" blocks are GATED: the rlm.watch.* host handlers arrive with the
      // communication series, so the validator rejects any wait block until then
    }
  ],
  "transitions": [
    { "from": "reviewing", "to": "fixing", "on": "settled",
      "when": { "output": "verdict", "path": "approved", "op": "eq", "value": false } }
  ]
}
```

[...]

PR #2401, opened 2026-09-16
swarm: state-machine swarm executor (guarded transitions, re-entry, wait states)

sethkarten:
## Summary

PR G of the swarm DAG feature: the executor. Builds on PR #2397 (swarm entry kind, `validate_swarm_spec` / `canonicalize_swarm_spec` / `topological_order`).

- **Python kernel side** (`prime-agent-runtime/src/rlm/swarm.py` + `rlm/__init__.py`): a `SwarmExecutor` exposed as the `rlm.swarm` namespace — `await rlm.swarm.run(spec_id, name=?)`, `rlm.swarm.status(run_id)`, `rlm.swarm.stop(run_id)`, `rlm.swarm.resume(run_id)`.
- **TS host side**: a `"swarm.progress"` host handler in `agent-session.ts` that injects one quiet steer-lane notice per run milestone (mirroring the `bash.completed` path), plus a hint sentence in the model prompt.
## State machine model

The swarm specification is now a **state machine** stored in `arguments["machine"]`. The original DAG form (`arguments["dag"]`) stays as sugar and **compiles to machine form at canonicalization**: a compiled dag is a restricted machine — each state entered at most once, forward-only, with guard-less transitions (one per effective dependency edge) and `max_entries: 1`.

```jsonc
{
  "run": {
    "budget_ms": 600000, "failure_policy": "escalate", // fail_fast | continue | escalate (default)
    "max_parallel": 8,                                   // 1..64 (default 8)
    "max_transitions": 100                               // default 10 * states, hard cap 10000
  },
  "states": [                       // 1..1024, unique slug ids, >= 1 entry state
    {
      "id": "reviewing",
      "entry": true,                 // admission enters every entry state
      "max_entries": 4,              // bounded re-entry; further transitions are blocked
      "subagent": "reviewer",        // required unless the state is a wait state (same forms as V1)
      "lifecycle": "task",           // task | resident
      "inputs":  [{ "name": "pr_url", "type": "text", "from": "entry.pr_url" }], // binds the LATEST settle of the source
      "outputs": [{ "name": "verdict", "type": "json" }],
      "budget_ms": 300000, "retries": 2, "failure_policy": "escalate",
      "foreach": { "over": "items", "max": 64 },   // expands per entry
      "wait": { "kind": "path", "target": "/tmp/x", "timeout_ms": 5000 }  // wait states only
    }
  ],
  "transitions": [
    { "from": "reviewing", "to": "fixing", "on": "settled",
      "when": { "output": "verdict", "path": "approved", "op": "eq", "value": false } }
  ]
}
```

- **Executor semantics:** ALL transitions whose guards pass fire from one settle (fan-out is legal; a guard-less transition always passes; switch patterns use mutually exclusive guards). A transition fires only if the target has entries left under `max_entries`; blocked transitions are recorded as `transition_blocked` ledger events and the machine continues. Re-entry re-binds inputs from the **current latest settles** of the input sources. Self-loops and back-edge cycles are legal — the V1 acyclicity check is removed entirely. Completion is **quiescence**: no state in flight (running/pending/waiting) and no firable transition from any unconsumed settle. `max_transitions` exceeded pauses the run once with a `budget_exceeded` milestone (resume-able).
- **Guards** (`when`, over the from state's latest settle output): ops `eq ne gt gte lt lte exists contains`; `path` resolves dotted paths inside json ports; numeric ops require numeric values, `contains` requires a list, `eq`/`ne` scalars.
- **Wait states** (`wait` block): no subagent, declared outputs, retries, or foreach, and never resident; entering registers an `rlm.watch.path` / `rlm.watch.agent` host watch and settles the implicit json port `event: {"detail", "timed_out"}` — from the watch outcome or its timeout through the injectable clock.
- **Unchanged from V1:** budgets (per entry: admission → settlement), retries (per entry attempt), foreach (per entry), state failure policies (fail_fast cancels in-flight, escalate pauses, continue finishes), resident teardown by `stop()`, notices, and the ledger's arrived/shown/read stages.

This PR reworks the executor onto the machine form: `run()` canonicalizes the stored spec (dag sugar compiles), admission enters the entry states, and the control loop evaluates transitions on every settle. **All 32 V1 dag executor tests run unchanged through the compiler path** (43 tests total with the machine suites). New ledger event kinds: `state_entry` (with entry index), `transition_fired`, `transition_blocked` (plus `wait_settled` accepted by the replay checker for legacy ledgers); `status()` adds per-state `entries_used` / `max_entries`, per-entry breakdowns, and a `transitions_fired` usage count over a 200-event window.

**Review-fix round:** wait states are gated at the validator (the `rlm.watch.*` host handlers arrive with the communication series #2351/#2356), so this PR removes the executor's wait registration/timeout/settle/cancel paths entirely — quiescence, re-entry, and fan-out are unchanged. A `max_transitions` pause mid-settle now records the transition index where it landed and a resume continues **after** it (previously the settle re-fired already-fired transitions, duplicating spawned work); `eq`/`ne` guards compare JSON-strictly (a bool never equals a number, numbers compare numerically); `contains` requires a non-empty list value at validation and is defensively false at runtime; the `max_transitions` pause fires its own `max_transitions_exceeded` milestone kind (no collision with the run-budget `budget_exceeded` — both notice-able); the control-loop stall names the pending entry whose input source never settled (with a test); EVENT_WINDOW grows 50 → 200 (the pr-manager happy path is ~43 events before any retry); refinement `validateEdit` accepts machine-form swarm edits (either `arguments.dag` or `arguments.machine` object, never both). Machine inputs may declare `"optional": true` to bind a null sentinel instead of waiting when their source state never settled — the enabler for the closed pr-manager review/fix loop in the eval PR (compiled dags never emit it, so dag semantics stay V1-exact).

Spec: [Swarm DAGs: declarative orchestration in Continual Harness](https://app.notion.com/p/3da72940136f-81a8-8554-e6ec7119270e) (sections "Executor" and "Runtime contract").

## Executor semantics

**Ownership split.** The RLM supervisor owns the children (admission via `rlm.spawn`, settlement via `rlm.collect`, cancellation via `rlm.delete_subagent`); the executor owns run state in kernel memory. Every host call resolves through the module-level `rlm` functions / `host_request` at call time, so tests patch `rlm.host_request` directly.

**Dry run = write time + run time.** `create_swarm` validates the graph at write time. `run()` re-validates and canonicalizes the stored dag, then resolves every node's subagent reference — a string is a harness subagent entry **id or title** (entry content becomes the prompt template; optional `metadata.model` / `metadata.thinking` carry spawn settings), an inline object uses its own fields. All reference failures are reported together in one `ValueError` and **nothing starts**. The resolved node count and `max_parallel` are reported in the result. Actual admission limits (concurrency, recursion depth, provider rate limits) are enforced at spawn time through the backoff path; that split is the V1 "runtime tree limits" dry run.

**Nonblocking admission.** `run()` initializes per-state state, creates an entry for every entry state (a `state_entry` ledger event each), then prepares and admits instances up to `max_parallel` (respecting foreach expansion), records handles, and returns `{"run_id", "spec_id", "name", "nodes", "max_parallel", "started", "pending"}`. A kernel asyncio task (the same `loop.create_task` pattern as the job-watch poller, guarded so a dead bridge fails the run gracefully instead of wedging) continues the work; the calling model turn ends immediately.

**Control loop (iterative, no recursion).** Polls in-flight instances with `await rlm.collect(ids, timeout_ms=2000)`. For each settled instance: capture the answer, `duration_ms`, mark done, append a ledger event; when an entry's last instance settles, the entry settles and captures the state's declared output ports (json ports parsed as in binding, text ports the captured string). Every settle is then evaluated once: ALL guard-passing outgoing transitions fire (fan-out is legal), each fire enters its target unless it is out of `max_entries` (recorded as `transition_blocked` and the machine continues), and re-entry re-binds inputs from the current latest settles. Completion is quiescence: no entry in flight (pending/running) and no unevaluated settle. (Wait states are specified but gated: the `rlm.watch.*` host handlers arrive with the communication series #2351/#2356, so the validator rejects `wait` blocks and this PR ships no wait execution paths.) Yields once per iteration so an instantly-settling host cannot hot-spin. Tested at 100-node chain and 1000-node wide-fan scale against a fake host.

**Input binding.** For each input `{name, type, from: "<node>.<output>"}`: text ports take the upstream captured answer; json ports parse the upstream text — preferring the trailing fenced ` ```json ` block whose object contains the output name, else the whole text, else the node errors. Rendering replaces `{input_name}` placeholders in a single pass; inputs without a placeholder are appended in a trailing `## Inputs` section, so no bound value is ever dropped. Binding failures never retry (a deterministic binding error would recur on every re-render); the node fails and its failure_policy applies.

**foreach.** When the `over` json input resolves to a list of K items, the node spawns K instances (each item bound as that input's value), K clamped to `foreach.max`. A foreach node is done when all instances settle; its captured answer is its instances' answers joined with blank lines. A node fails at its FIRST permanently failed instance — it does not wait for its remaining instances — so `fail_fast` cancels in-flight siblings while they are still running, and a failure that settles before its successes can never strand the node in a mixed terminal state.

**Retries and failure policies.** A child failure (collect `error` status/field) re-spawns the same rendered prompt while `attempts <= retries`; then the node `failure_policy` applies: `fail_fast` → cancel every running child of the run via `rlm.delete_subagent` and mark the run failed; `continue` → mark the node error and keep going (dependents without a data edge still run; data dependents fail at binding); `escalate` (default) → mark the node error and pause the run, spawn nothing further until `resume()`.

**Budgets (injectable clock).** Wall-clock budgets measure **admission to settlement**. Per node: if the spawn-to-settle span exceeds `budget_ms`, the attempt is treated as failed — the budget is spent, so no retry; the failure_policy applies. Run-level: elapsed time past `run.budget_ms` pauses new spawns (children already in flight keep running) with a `budget_exceeded` milestone; the budget milestone reports once per run, and resuming after it is an explicit operator decision.

**Rate limits.** If spawn admission fails with a rate-limit/429-style error, the executor retries with exponential backoff — doubling delays capped at 60s, at most 5 admissions per attempt — then fails the node through its failure_policy; the loop never crashes. In the `run()` admission phase (and in `resume()`, exactly the same way) a rate limit does not sleep inside the calling turn: the node stays pending and the control loop retries it with backoff (keeps both calls nonblocking).

**stop() and races.** `stop()` sets a transitional `stopping` state before its first await, so the control loop can neither admit new children nor finalize the run while the cancellations are in flight; the loop re-checks state before completion, `_finalize` only runs while the run is `running`, and the executor-error path never overwrites a concurrent stop. A stopped run therefore stays `stopped` (never `done` + a spurious `finished` notice). Repeated `stop()` is idempotent and appends only one `run_stopped` ledger event.

**Event ledger.** Kernel memory, per run: spawned/attempts, settled (with `duration_ms`), answer captured, retries, errors, and run milestones. Stages follow the spec: `arrived` (child answer settled and captured), `shown` (milestone notice injected), `delivered` (parent read via `status()` — marked on each call). `status()` returns state states (with per-state `entries_used`/`max_entries`, per-entry breakdowns, and per-instance details), the trailing 50 events, `elapsed_ms`, and usage (spawns/settled/tool uses/running/`transitions_fired`). `status`/`stop`/`resume` raise `ValueError` on an unknown run id.

**Milestone notices.** One quiet runtime notice per milestone kind per run (`finished` / `failed` / `paused` / `budget_exceeded`), never per node, sent via `host_request("swarm.progress", {run_id, kind, node?, detail})`. The TS handler (`createSwarmProgressHostHandler` + `createSwarmProgressMessage`) validates the payload and injects a quiet custom message (customType `swarm_progress_notice`, steer lane, `queueIfBusy`/`resumeIfIdle`, `suppressAutonomousContinuation`, mirroring `bash.completed`) that wakes the parent with `[swarm-progress run:<id>] finished|failed|paused|budget-exceeded: <detail>`. A dead bridge cannot be told; the ledger keeps the milestone and `status()` still surfaces it.

**Resident nodes (V1 wake sources).** A resident node is admitted like any other node and then stays alive under the parent session; its wake source is its own subagent prompt/tooling (watches, heartbeats, or continuous work inside the child) — no structural dry-run rule in V1. A run's declarative work completes when all non-resident nodes settle; resident children keep running (they cannot outlive the parent session) until `rlm.swarm.stop(run_id)` tears them down.

## Known limitations (documented deliberately)

- **Runs do not survive a kernel restart.** Run state lives in the executor's kernel-memory registry; after a restart the registry is gone. Children are supervisor-owned and keep running — `rlm.list_subagents()` can still see and stop them.
- **Answer-preview cap.** Answers are captured from `rlm.collect` previews, which the host caps at 160 characters (`compactRlmText`); the executor additionally caps at 200 defensively. Input binding and downstream prompts work on these previews; full child outputs stay in the child's own session.
- A paused run does not observe child settlements until `resume()` (completed children keep their results for a later `collect`).
- Run completion with any node error reports state `failed` (continue policy finishes the graph, but the run is marked failed so the parent sees the error); stop() reports `state: "stopped"` and lists every non-terminal node as cancelled.
- **Node budgets are settlement-only.** A child that never settles is never budget-failed, so the loop keeps polling it indefinitely (V1 limitation); the natural follow-up is a per-instance watchdog. Run budgets share this: they pause new spawns, never kill in-flight children.
- **Cancellation shortcuts (V1).** If `delete_subagent` fails, the instance is still reported `cancelled` in the ledger (the child keeps running supervisor-side; the `cancel_failed` event records the error), and a deferred (rate-limit) instance of a cancelled node stays `pending` in `status()` output.

## Tests

- `prime-agent-runtime/test/test_swarm_executor.py` (32 tests, deterministic, no live model; patched `rlm.host_request` async fake routing by request type with per-child-id scripted outcomes, collect/delete suspension gates, injected fake clock and sleeps): dry-run rejection (invalid dag via a bypassed store entry, missing references listing all failures, unknown spec), admission starting exactly the in-degree-0 nodes with `max_parallel` respected, propagation with exact rendered prompts (text, fenced-json, whole-text json, bad json → node error without spawning, unplaced inputs appended, two-parent fan-in in one collect batch, answer-capture cap slice), foreach expansion + clamp + zero items, all three failure policies (fail_fast cascade cancels running children; continue finishes remaining nodes; escalate pauses + notices + `resume()` continues) plus mixed-instance foreach per policy (fail_fast cancels in-flight siblings; escalate pauses; continue finishes with node error), retries until attempts exhausted, node and run budgets via the fake clock, stop() cascade + states + idempotency, a gated stop-race test proving a stop mid-collect keeps `stopped`, finalizes nothing, and admits no new children, 429 admission deferral → backoff → success and backoff exhaustion → node error (including the `resume()` deferral path), 100-node chain and 1000-node fan scale under a generous wall bound, status delivered-marking + unknown-run `ValueError`, resident node spawn/stay/stop.
- `packages/coding-agent/test/swarm-executor.test.ts` (12 tests, mirroring the async-bash completion test): bracket-grammar notice content, `convertToLlm` pass-through, `startsAgentRun` wake, payload validation and forwarding, invalid payload rejection.
- `prime-agent-runtime`: `uv run python -m unittest discover -s test` — 384 tests, only the two pre-existing `test_bash` failures (`test_windows_without_bash_raises_teaching_error`, `test_unconfigured_journal_stays_permissive`, both reproduced on the clean base branch).
- `packages/coding-agent`: `npx vitest --run test/swarm-executor.test.ts test/refinement.test.ts` — 83 passed. Repo root: `npx tsgo --noEmit` and `npx biome check` on changed files — clean.

Stacked: base branch is swarm/swarm-spec-harness-entry (#2397); merge that first.

Draft only - review withheld per workflow.


sethkarten:
Review round complete (for the morning review): the review found 2 must-fix bugs - foreach nodes with mixed instance outcomes never reached the failure policy, and stop() raced the control loop (run could finalize as done after an explicit stop). Both are fixed in commit 8344cdbec with 7 regression tests (suite now 32, all green; gates clean): foreach now fails at the first permanently failed instance (fail_fast cancels in-flight siblings immediately), stop() uses a transitional stopping state with finalize guards, and resume() no longer sleeps in the model turn. Both fixes were verified by re-running the original reproductions. Remaining nits are documented V1 limitations in the PR body (settlement-only budgets, cancellation shortcut reporting).

sethkarten:
Integration note (verified by merging all open swarm branches together): main now enforces a deferred-boot contract - repl tests assert Version: ImageMagick 7.1.2-31 Q16-HDRI aarch64 8309dc92a:20260903 https://imagemagick.org
Copyright: (C) 1999 ImageMagick Studio LLC
License: https://imagemagick.org/license/
Features: Cipher DPC HDRI Modules 
Delegates (built-in): bzlib freetype heic jng jpeg lcms ltdl lzma png tiff webp xml zlib zstd
Compiler: clang (21.0.0)
Usage: import [options ...] [ file ]

Image Settings:
  -adjoin              join images into a single multi-image file
  -border              include window border in the output image
  -channel type        apply option to select image channels
  -colorspace type     alternate image colorspace
  -comment string      annotate image with comment
  -compress type       type of pixel compression when writing the image
  -define format:option
                       define one or more image format options
  -density geometry    horizontal and vertical density of the image
  -depth value         image depth
  -descend             obtain image by descending window hierarchy
  -display server      X server to contact
  -dispose method      layer disposal method
  -dither method       apply error diffusion to image
  -delay value         display the next image after pausing
  -encipher filename   convert plain pixels to cipher pixels
  -endian type         endianness (MSB or LSB) of the image
  -encoding type       text encoding type
  -filter type         use this filter when resizing an image
  -format "string"     output formatted image characteristics
  -frame               include window manager frame
  -gravity direction   which direction to gravitate towards
  -identify            identify the format and characteristics of the image
  -interlace type      None, Line, Plane, or Partition
  -interpolate method  pixel color interpolation method
  -label string        assign a label to an image
  -limit type value    Area, Disk, Map, or Memory resource limit
  -monitor             monitor progress
  -page geometry       size and location of an image canvas
  -pause seconds       seconds delay between snapshots
  -pointsize value     font point size
  -quality value       JPEG/MIFF/PNG compression level
  -quiet               suppress all warning messages
  -regard-warnings     pay attention to warning messages
  -repage geometry     size and location of an image canvas
  -respect-parentheses settings remain in effect until parenthesis boundary
  -sampling-factor geometry
                       horizontal and vertical sampling factor
  -scene value         image scene number
  -screen              select image from root window
  -seed value          seed a new sequence of pseudo-random numbers
  -set property value  set an image property
  -silent              operate silently, i.e. don't ring any bells 
  -snaps value         number of screen snapshots
  -support factor      resize support: > 1.0 is blurry, < 1.0 is sharp
  -synchronize         synchronize image to storage device
  -taint               declare the image as modified
  -transparent-color color
                       transparent color
  -treedepth value     color tree depth
  -verbose             print detailed information about the image
  -virtual-pixel method
                       Constant, Edge, Mirror, or Tile
  -window id           select window with this id or name
                       root selects whole screen

Image Operators:
  -annotate geometry text
                       annotate the image with text
  -colors value        preferred number of colors in the image
  -crop geometry       preferred size and location of the cropped image
  -encipher filename   convert plain pixels to cipher pixels
  -extent geometry     set the image size
  -geometry geometry   preferred size or location of the image
  -help                print program options
  -monochrome          transform image to black and white
  -negate              replace every pixel with its complementary color 
  -quantize colorspace reduce colors in this colorspace
  -resize geometry     resize the image
  -rotate degrees      apply Paeth rotation to the image
  -strip               strip image of all profiles and comments
  -thumbnail geometry  create a thumbnail of the image
  -transparent color   make this color transparent within the image
  -trim                trim image edges
  -type type           image type

Miscellaneous Options:
  -debug events        display copious debugging information
  -help                print program options
  -list type           print a list of supported option arguments
  -log format          format of debugging information
  -version             print version information

By default, 'file' is written in the MIFF image format.  To
specify a particular image format, precede the filename with an image
format name and a colon (i.e. ps:image) or specify the image type as
the filename suffix (i.e. image.ps).  Specify 'file' as '-' for
standard input or output. must not pull asyncio or secrets. prime-agent-runtime/src/rlm/swarm.py imports asyncio at module level; defer it into the four functions that use it (SwarmExecutor ctor default sleep, _start_loop, _control_loop, _loop_body) when refreshing onto main.

sethkarten:
Consolidated into #2452 (DAG swarm workflows as a single review/merge unit) per the DAG-first review plan. All commits, tests, and review history from this PR are included there; this branch is unchanged and can be deleted.

sethkarten:
Status correction: the consolidation above was reverted - this PR is the active review unit again, restored as part of the stacked series (this PR -> its dependents up to #2485). Ignore the "Consolidated into #2452" note; #2452 is closed and its branch deleted.

PR #2419, opened 2026-09-16
[RSI, bug] Harden the daemon supervisor monitor timing probe against runner load

sethkarten:
## Problem

`daemon-supervisor-monitor.test.ts` carries a known CI flake: the test "keeps the supervisor monitor armed after a replacement binds but exits before claiming" failed across at least three unrelated PR runs on loaded runners, always passing on rerun and locally.

## Root cause

The test drives the supervisor availability-check chain with fake timers (`vi.advanceTimersByTimeAsync(100)` in a loop capped at 200 steps) and then asserts the exact probe count. The chain is not pure fake time: each availability check awaits `isDaemonShutdownAdmissionActive()` before probing and rescheduling, and that takes **real** filesystem locks (proper-lockfile mkdir/stat/rmdir inside the supervisor registry guard). Fake timers cannot advance those; probes only run once the real fs ops resolve.

Each `advanceTimersByTimeAsync` call yields only about one real event-loop turn, so the old 200-step cap was effectively a tiny **real-time** budget (measured ~2-5 ms locally). Any loaded-runner fs stall blows that budget: the loop exits with `probeCount` short and the exact assertion fails. Confirmed with a deterministic reproduction — injecting a one-shot 150 ms real delay into the registry guard makes the old helper fail with the exact CI signature (`expected +0 to be 2`).

## Fix

Replace the fixed 200-step advance loop with a poll-with-timeout that keeps stepping the fake clock until the expected probes land, bounded by:

- an explicit step cap (100,000 × 100 ms of fake-time headroom), and
- a real wall-clock cap (10 s per phase, well inside the package's 30 s `testTimeout`), using `Date.now` captured before `vi.useFakeTimers()`.

Every semantic assertion is unchanged and equally strong: probe sequence `[false, true, false, false]`, exactly one launch after the first phase, exactly two launches and four probes after the full sequence, `supervisorAbsentSince` cleared at bind, monitor timer armed at both checkpoints. The failure message now reports observed probes, steps, wall time, `absentSince`, and monitor-timer state instead of a bare count.

Test-infra only; no production code changed.

## Validation

- Targeted suite 10/10 consecutive passes (108 tests each).
- Simulated load: 29/29 passes — 3 rounds x 6 concurrent copies, one mixed round with the other daemon-supervisor suites running concurrently, and one round with 16 CPU burners plus 6 concurrent copies.
- Deterministic injected 150 ms real-fs stall: old helper fails, hardened helper passes (A/B verified independently by implementer and reviewer).
- `npm run check` clean (biome + tsgo + installer + browser-smoke).

---

> [!NOTE]
> **Low Risk**
> Test-only changes to timing helpers; no production behavior modified.
> 
> **Overview**
> Fixes a flaky CI failure in **daemon-supervisor-monitor** where fake-timer stepping could finish before real filesystem-lock work in the availability-check chain completed, leaving probe counts short on loaded runners.
> 
> The probe-driving helper no longer stops after a fixed number of `advanceTimersByTimeAsync(100)` iterations. It keeps advancing fake time until the expected probes appear, with a large fake-time step budget and a **10s real wall-clock** cap (using `Date.now` captured before `vi.useFakeTimers()`). Failed expectations now report probe progress, steps, wall time, and monitor state instead of a bare count mismatch.
> 
> Assertions on probe sequence, launch counts, and monitor arming are unchanged.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b6ac5d014d99401b55820835a4966584271e9a3c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Harden daemon supervisor monitor timing probe test against runner load
> Updates [daemon-supervisor-monitor.test.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/2419/files#diff-c28db4d48d5180cce23d4192cce2b049f88dfb6b9bf98622bafc3a9eb018f65d) to capture the real wall-clock implementation before enabling fake timers, so availability-check probes that wait on real filesystem-lock operations are accounted for.
>
> - Reworks the probe-driving helper to keep advancing fake timers until the expected probe count is reached, with a 100,000-step limit and a 10-second real-time limit
> - Expands the failed-expectation message with probe progress, elapsed wall time, the absence timestamp, and whether the monitor timer is armed
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 0daeb84.</sup>
> 
>


sethkarten:
Fixed: the helper comment now states the contract in one line (fake clock, step and wall bounds, observed-state assertion) instead of narrating the loop and the removed bare-count assertion; the Date-capture note above already covers the filesystem-lock constraint. Comment-only change; no behavior difference.

sethkarten:
Superseded by #2336 — evidence: main\'s #2336 landed ~2h after this PR\'s head and rewrote the same region, deleting `advanceUntilProbes` (the only code this PR patched) and replacing it with `settleSupervisorAvailabilityCheck` awaiting the new production completion signal `supervisorAvailabilityCheckSettled` — exactly the "await a concrete event or deferred promise" form AGENTS.md prescribes over a polling loop. Main\'s test keeps every assertion this PR protected (probe sequence, launches, absentSince, monitor-timer checkpoints) plus an explicit probe-count assertion, so no coverage is lost. A rebase of this branch onto current main yields an empty diff (file blob byte-identical to main). The test-line budget rule (net test additions must not exceed source additions) would also fail this +16/-4, 0-source change. Recommendation: close as superseded. Diagnostic residue worth a separate small change if wanted: main\'s settle await is unbounded, so a stalled chain surfaces as an opaque 30s timeout rather than a rich message — a bounded Promise.race guard against the completion signal would restore that, sized inside the test budget.

PR #2433, opened 2026-09-17
[RSI, performance] Append catalog metadata without full transcript parses

sethkarten:
## What

Catalog metadata operations (rename, archive, mark_interrupted) appended one line to a closed session file by opening a full `SessionManager`, which parses and indexes the entire transcript. That is an O(session) stall plus a memory spike proportional to session size on every routine UI action.

This adds `appendEntryToExistingFile` and three wrappers (`appendSessionInfoToExistingFile`, `appendSessionStateToExistingFile`, `appendCustomMessageToExistingFile`) that:

- repair crash damage exactly like a full open (`repairJsonlDamage`),
- validate the session header from a bounded 64 KiB leading window,
- resolve the current leaf id from a bounded 256 KiB tail window,
- append a single line.

They fall back to `SessionManager.open` whenever only a full open can place the entry correctly: a header past the window, a file version this build does not write (older or newer), an oversized tail, or a tail too small to resolve the leaf. `SessionManager` now shares the entry builders with the fast path, so both write the same line shape.

The daemon catalog commands `rename`, `archive`, and `mark_interrupted` route through the wrappers.

Measured on an 8.2 MB, 20k-entry fixture: rename drops from 18.3 ms to 0.5 ms median.

## Behavior change

A missing or header-invalid session file now fails these updates with a clear error (`Cannot append to missing session file`, `Session file has no valid session header`) instead of being recreated or silently rewritten as a fresh session. In worker recovery the interruption notice stays advisory: a notice that cannot be written is logged, and recovery still reaps orphaned processes and resolves its journal.

## Count-based evidence

`test/session-manager/append-to-existing-file.test.ts` counts fs calls through passthrough spies, so every claim is a call count rather than a timing:

- 18k-entry fixture larger than 1 MB: `readFileSync` is never called, exactly one `appendFileSync` call happens, and the appended bytes parse as one `session_info` entry chained to the previous leaf.
- Fallback: a v1 file forces the full open (`readFileSync` called), the entry still lands, and the migration rewrites the header to v3; an oversized tail with no assistant entry falls back, the live manager suppresses the notice, and the file is left byte-identical.
- Clean errors: missing and headerless files throw (`Cannot append to missing session file`, `Session file has no valid session header`) and leave the file untouched (nothing created, appended, or rewritten).

## Test-line budget

Fleet policy caps net test additions per PR at the PR's meaningful added source lines (`scripts/check-test-policy.mjs`, run by `npm run check`); the fleet-wide audit directive was: "probably needs to remove redundant and unnecessary tests throughout. Remove unnecessary tests and we should have enough." This PR could not reach that budget by consolidation alone, and the owner decided against grandfathering the gap: "nah dont grandfather them in. we should fix them now" - the coverage cuts below are owner-authorized, ratio over coverage, with nothing removed silently.

- Before: net test additions 345 vs source 138 (gate red).
- Honest floor (consolidation only, commit a9e03959b): 319 vs 138 (red; a bounded audit of the owned daemon test files found no true duplicates left to delete honestly - the area was pruned by #2336).
- After owner-authorized cuts (commit d3af14e4a): 122 vs 138 (green; `npm run check` passes at branch level against main).

### Lost-coverage ledger

Per-vector record of every executed cut: what the deleted test pinned, why it was the cheapest option, and the surviving partial pin. PROBE entries were verified by mutating the named source and observing the named test.

1. **v4 (future-version) fallback row** - pinned the version-mismatch guard's non-migrating variant. Cheapest: the guard branch (`header.version !== CURRENT_SESSION_VERSION`) stays pinned by the v1 row (PROBE PM-D: removing the guard makes the v1 fallback test fail - the pin is alive).
2. **empty-file refusal vector** - pinned the `!header` arm of the header throw. Cheapest: the identical throw line is pinned by the malformed-header vector. Partial pin: malformed vector; the distinct `!header` arm is lost.
3. **18k reopen round-trip** - pinned that the fast-appended line loads and applies through a real open. Cheapest: the same round-trip observable is pinned live-side. Partial pin: `session-state.test.ts` (open -> `appendSessionInfo` -> list -> name applies).
4. **torn-tail repair phase** - pinned repair-before-append on the fast path. Cheapest: the repair function is pinned at open. Partial pin: `file-operations.test.ts` repairs crash damage at open. Lost: the fast path's own repair call.
5. **state fast-write phase** - pinned `appendSessionStateToExistingFile` (archive path). Cheapest: entry construction is pinned live-side. Partial pin: `session-state.test.ts` live `appendSessionState` via the shared `buildSessionStateEntry`.
6. **leading-blank row** - pinned blank-prefix fast-path acceptance. Cheapest: the skip-to-first-parseable branch is exercised by the refusals fixture. Partial pin: refusals' malformed fixture; the acceptance observable is lost.
7. **header-only row** - pinned `readTailLeafId`'s empty-file null-leaf branch. Cheapest per line among the shape rows. No surviving pin (null-leaf branch unpinned).
8. **catalog subprocess integration test** - pinned rename/archive/mark_interrupted dispatch wiring in `daemon-catalog-process.ts`, in-place append + untouched transcript, and client-visible error propagation. Cheapest: the fast functions stay unit-pinned and the supervisor arg contract survives. Partial pins: unit tests here; `daemon-supervisor-monitor.test.ts` "marks each busy worker session interrupted independently" (PROBE PM-C: reverting the catalog rename case to `SessionManager.open` now yields zero failures - the dispatch wiring is confirmed unpinned).
9. **notice fast-write phase** - pinned `appendCustomMessageToExistingFile`'s fast write (custom entry shape, details, no-assistant). Cheapest: the same function's fallback+suppression path stays pinned. Partial pin: the fallback-suppression phase (PROBE PM-E: enlarging the tail window makes the suppression phase fail - the window-trigger pin is alive). Lost: the fast-write entry-shape pin.
10. **supervisor notice-failure tolerance test** ("completes recovery when an interrupted session notice cannot be written") - pinned `recoverUncertainWorkerOperations`' error tolerance: a notice write that fails is logged, the orphan is still reaped, the journal still resolves. Cheapest per line: one unique vector across ~35 lines; the happy path, arg contract, and reap mechanics survive elsewhere. Partial pins: "marks each busy worker session interrupted independently" (same flow, happy path, per-session args); the orphan-journal `it.each` tables (reap mechanics); the unit refusals test (the same missing-file failure class the supervisor logs) (PROBE PM-B: rethrowing `markInterrupted` errors in the supervisor now yields 108/108 pass - the tolerance pin is confirmed gone).
11. **Not cut - consolidated** (zero vectors lost): PR helper blocks one-lined; the suppression test folded into the fallback test as a phase instead of deleted; fixture and comment compaction. Surviving pins probe-verified alive: version guard (PM-D), tail window (PM-E), missing-file guard (PM-F), name trim (PM-RA) - removing each makes the named test fail.

## Validation

- `packages/coding-agent` targeted suites all green: append-to-existing-file, daemon-supervisor-monitor (full), append-hot-path, file-operations, flat-storage, session-state, daemon-catalog-startup, daemon-session-id, tree-traversal, build-context.
- `npm run check` (branch level vs main): biome, `check:test-policy`, `tsgo --noEmit`, installer, and browser-smoke all pass.

---

> [!NOTE]
> **Medium Risk**
> Changes session-file mutation semantics (errors instead of silent recreate) and append behavior on legacy or edge-case transcripts via bounded reads with full-open fallback; recovery now swallows catalog write failures, which could hide transcript corruption from operators.
> 
> **Overview**
> **Catalog metadata** (rename, archive, worker-interruption notices) on closed session files no longer opens a full `SessionManager` parse for every UI action. New **bounded append APIs** validate the session header from a 64 KiB leading window, resolve the leaf `parentId` from a 256 KiB tail, repair JSONL damage, and write one JSONL line—falling back to `SessionManager.open` only when version migration, an oversized tail, or an ambiguous header requires it. Live appends share the same entry builders so line shape stays consistent.
> 
> The **daemon catalog** routes `rename`, `archive`, and `mark_interrupted` through these wrappers. **Missing or header-invalid files** now throw explicit errors instead of being recreated or rewritten as empty sessions. During **worker recovery**, a failure to write the interruption notice is logged and recovery still reaps orphans and resolves the journal (the notice is advisory).
> 
> Tests add fs call-count coverage for the fast path, fallback, and clean failures; supervisor recovery tests share consolidated fixtures.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7ca34f88a2ef34fab305911eb83a21c636ec402c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Add bounded append APIs to `SessionManager` for catalog metadata
> - `SessionManager` uses `readFirstParseableLine` and `readTailLeafId` to validate headers and resolve the parent leaf with bounded file reads.
> - New append APIs (`appendSessionInfoToExistingFile`, `appendSessionStateToExistingFile`, `appendCustomMessageToExistingFile`) write a single JSONL line for current-version files, falling back to a full open for unsupported versions or unresolved tails.
> - `daemon-catalog-process.ts` uses these APIs for rename, archive, and interruption-marking. `DaemonSupervisor.recoverUncertainWorkerOperations` treats interruption-notice failures as advisory, logging errors and continuing recovery.
> - Risk: `appendEntryToExistingFile` throws explicit errors for missing or header-invalid files instead of recreating them; callers in `daemon-catalog-process.ts` handle these as standard command errors.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 7ca34f8.</sup>
> 
>


snimu (approved):

PR #2465, opened 2026-09-19
Hold goal/autonomous continuations while background bash() handles run

sethkarten:
## Motivation

A goal-active session that ends its turn with a background `bash()` handle still running was immediately re-prompted by the timer-driven continuation, because the existing hold (`_autonomousContinuationAwaitsRlmWork` / `_goalContinuationAwaitsRlmWork`) only paused for unsettled RLM descendant work. The continuation raced the handle's own wake-up (the bash-done completion follow-up) and burned a continuation turn while the agent was just waiting on the command. The user direction: "goal shouldnt do continuation when a background bash() shell is running. background bash() shells should message agent when completed" — that completion follow-up already exists; the scheduler just never waited for it.

## Implementation

Mirrors the existing RLM-child hold, extending it to background bash handles:

- **Liveness query (reused, not duplicated):** `AgentSession._hasLiveBackgroundBashHandles()` reads `KernelClient.hasBackgroundWork` — the kernel's bash-activity tracking that already powers the bash-done follow-ups.
- **Hold:** `_getGoalContinuationMessages` and `_holdAutonomousContinuationForRlmWork` now hold the timer-driven continuation while any background handle is live, without consuming goal/autonomous continuation budget.
- **Resume:** `ReplKernelManager` notifies the session (`KernelManagerOptions.onBackgroundWorkSettled`, threaded through `IpythonKernelProvisioner`) when its live handle set empties — i.e. after the last handle's completion notice is admitted — or when the kernel tears down. The session then fires the same `_maybeResumeGoalContinuationAfterRlmWork` / `_maybeResumeAutonomousContinuationAfterRlmWork` pair the RLM child settlement sites use, so the owed continuation resumes behind the notice exactly like child exit notices. Both resume paths keep the deferral while handles remain live, and both notification sites are guarded so a host callback failure can neither break the kernel event path nor abort teardown.
- **Keep-alive valve:** while only background handles are pending, the autonomous valve re-polls after another window instead of waking the parent with a subagent keep-alive message (the completion follow-up is the real wake-up).
- **Never gated:** user prompts and explicit `/goal` commands are untouched; a user message always wakes the session immediately.

Docs updated in `docs/long-running-agents.md` and `docs/rlm.md`.

## Verification

- `npm run check` (biome, test policy, tsgo, installer, push guard, browser smoke) passes; the test-policy line budget is 39 net test lines vs 39 meaningful source lines.
- Targeted tests green: `autonomous-continuation-subagent-gate` (28), `goal-continuation-quiescence` (8), `2053-background-bash-passivation` incl. real-kernel rows (9), `async-bash-completion` (6).
- Mutation probes: each new row fails when its guard is reverted (autonomous hold, goal hold + resume deferral, valve re-poll, settlement callback incl. the teardown branch).
- Empirical end-to-end probe (real kernel through the real session provisioner wiring, faux provider): a goal-active session that spawns `bash('sleep 2')` and ends its turn holds the continuation (`continuationsUsed: 0`, no goal context message) until the handle settles; the bash-done notice then wakes the session and the held continuation is delivered exactly once, behind the notice. The same probe fails on pristine main with two continuations delivered immediately while the handle still runs — the reported bug.
- Dual-reviewer pre-push gate (deepseek-v4.1-flash, Bugbot + Macroscope personas, each with their own live-session probes over the real wiring): both PASS. Their non-blocking findings (unguarded host callback, weak valve re-arm pin, missing teardown pin, docs wording) are fixed in this commit and mutation-verified.
- Pre-existing isolation flake (not from this change): running `agent-session-goal.test.ts` after `agent-session-autonomous.test.ts` in one vitest invocation fails 4 `initialGoal` seeding tests; reproduced identically on pristine `origin/main`, each file passes alone. Known pre-existing gap (also not from this change): the threshold-compaction goal continuation queue path bypasses both the RLM quiescence gate and this new handle gate.

---

> [!NOTE]
> **Medium Risk**
> Changes long-running goal/autonomous scheduling and kernel teardown callbacks; incorrect gating could stall continuations or wake too early, though behavior mirrors an existing subagent hold with targeted tests.
> 
> **Overview**
> Timer-driven **goal** and **autonomous** continuations now **pause** while any live background `bash()` handle is running, matching the existing hold for unsettled RLM subagent work. Continuation budget is **not** spent during the wait; the last handle settling is the wake-up—the **"Background command finished"** notice (when owed) is delivered first, then the held continuation resumes behind it.
> 
> **AgentSession** treats kernel `hasBackgroundWork` as the liveness signal via `_hasLiveBackgroundBashHandles()`, extending hold/resume gates in goal continuation, autonomous hold, and subagent keep-alive (re-arm the timer instead of injecting a keep-alive when **only** bash handles are pending). **ReplKernelManager** adds `onBackgroundWorkSettled`, invoked when the bash-activity map empties or on teardown, wired through **IpythonKernelProvisioner** to resume owed continuations. Docs in `long-running-agents.md` and `rlm.md` describe the behavior; tests cover hold, resume deferral, settlement callback, and teardown.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0545530ba3a25e15df2adf4ef0bd06ef7a060ab9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Hold goal and autonomous continuations while background bash handles run
> - Timer-driven goal and autonomous continuations now pause without consuming continuation budget while a kernel background bash handle is active, resuming after the last handle settles.
> - Adds an `onBackgroundWorkSettled` callback wired from `IpythonKernelProvisioner` through `ReplKernelManager` so the session wakes deferred continuations when the final handle clears or the kernel tears down.
> - The autonomous keep-alive timer (`_fireAutonomousSubagentKeepAlive`) no longer admits a keep-alive turn when only background handles remain; it re-arms and waits for settlement.
> - Behavioral Change: `_getGoalContinuationMessages` and `_holdAutonomousContinuationForRlmWork` now return no continuation and skip budget increments when `_hasLiveBackgroundBashHandles` is true; a never-exiting handle keeps the continuation deferred indefinitely until a user prompt arrives.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 0545530.</sup>
> 
>


snimu (approved):

PR #2474, opened 2026-09-20
[RSI, performance] perf(coding-agent): jitter provider retry backoff to de-synchronize retry waves

snimu:
## Problem

`providerRetryDelay` (packages/coding-agent/src/core/provider-retry.ts:85-97) computed `min(max(baseDelayMs * 2^(attempt-1), retryAfterMs ?? 0), MAX_TIMER_DELAY_MS)` with no jitter. Every concurrent session retrying a provider outage shares the same policy defaults (baseDelayMs 2000, maxRetries 3 — provider-retry.ts:139-144), so the whole fleet retries in lockstep on the same 2s/4s/8s schedule: synchronized retry waves (thundering herd) against a shared provider during outages and 429s. Jitter existed only in the wait-for-recovery pings (provider-retry.ts:208-211), never in the quick retry loop used by the session auto-retry loop (agent-session.ts:12700) and every one-shot consumer (`completeWithProviderRetry`: compaction.ts:562,828; side-question.ts:164; refinement.ts:1190,1279; branch-summarization.ts:300; daemon-session-summarizer.ts:172).

## Change

- `providerRetryDelay` now applies full +/-25% multiplicative jitter (reusing the existing `providerWaitJitter`) to the computed exponential backoff only, so concurrent sessions spread out their retries.
- A server-specified Retry-After is honored exactly: when it is >= the computed backoff it passes through unchanged (bit-identical to the previous output on that path), and otherwise it stays a floor the jittered backoff never under-shoots. Same in-file precedent as `providerWaitDecision`, which schedules server-reported reset times exactly and jitters only blind pings.
- The exceeds-cap check (`retryAfterMs > maxRetryDelayMs` -> abort) stays first and unchanged; the `MAX_TIMER_DELAY_MS` overflow clamp stays last (the 90-day overflow regression test passes unchanged).
- Jitter is injectable via an optional `rng` parameter (default `Math.random`); both call sites are untouched and tests stay deterministic.

## Tests/Validation

- `packages/coding-agent/test/provider-retry.test.ts`: one new deterministic test (19 -> 20) covering the jitter boundaries via injected rng (attempt 1: 1500 at rng=0, 2500 at rng=1), exact Retry-After pass-through (5000 stays 5000 even at max jitter), and the server-wait floor (jittered 6000 floored to 7000).
- Sandbox (Prime sandbox, node:24): `vitest --run test/provider-retry.test.ts` 20/20; retry-adjacent suites (agent-session-bus, agent-traces, suite/agent-session-retry-events) 104/104 with the pass-through 50/40/40 assertions unchanged; root `npx tsgo --noEmit` clean; `npm run check` fully clean (biome, test policy, tsgo, installer, push-guard, browser-smoke).
- Test policy: net test additions (+8) vs meaningful source additions (+8) within budget; no new timers/sleeps/polls in tests.

Comments minimized per review (comment-only follow-up commit).

---

> [!NOTE]
> **Low Risk**
> Retry timing only—no auth, data, or API contract changes; optional `rng` keeps existing three-argument callers compatible.
> 
> **Overview**
> **Provider retry backoff is no longer deterministic across sessions.** `providerRetryDelay` applies the existing ±25% `providerWaitJitter` to the exponential backoff so concurrent agent sessions spread retries instead of hitting the same 2s/4s/8s schedule during shared outages.
> 
> **Server `Retry-After` behavior is preserved:** when the server wait is at or above the computed backoff, that delay is returned unchanged (no jitter). When the server wait is lower, the jittered backoff is floored at that minimum. Exceeds-cap and `MAX_TIMER_DELAY_MS` clamping are unchanged.
> 
> An optional injectable `rng` on `providerRetryDelay` supports deterministic unit tests for jitter bounds and pass-through cases; production call sites keep the default `Math.random`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 37a6ca016419aa1cc90097f604f91ce880a09a92. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Add ±25% jitter to `providerRetryDelay` exponential backoff
> - `providerRetryDelay` now applies +/-25% random jitter to the computed exponential backoff to de-synchronize retry waves across consumers.
> - When the server supplies a `Retry-After` wait at or above the backoff, it is used as-is without jitter. When the server wait is smaller than the backoff, the jittered backoff is floored at that server wait.
> - An injectable random-number generator was added to `providerRetryDelay` to enable deterministic test coverage of jitter bounds and server-wait handling.
> - Risk: callers that injected the old `providerRetryDelay` signature will need to update to pass the new RNG parameter; see [provider-retry.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/2474/files#diff-6b31923dc9faefc23eb791f813ba9596050d1fd8c0c238f2e6f1738de59691fc).
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 37a6ca0.</sup>
> 
>


sethkarten (approved):

PR #2491, opened 2026-09-21
fix(agent): end truncated proxy streams and send serviceTier through the proxy

snimu:
No-Ticket: found by a read-only pattern audit of main, not tracked in an issue.

Two proxy transport bugs from `packages/agent/src/proxy.ts`.

1. A truncated proxy response hung the caller. When the SSE body ended cleanly without a `done` or `error` event, the code called `stream.end()` with no result. Iteration stopped, but `EventStream.result()` never settled, so an agent awaiting the turn waited forever. The truncated case now terminates with an error event, like the direct transport contract in `packages/ai/src/types.ts` requires.

2. `serviceTier` was never sent to the proxy. The agent always passes it, but the proxy serialization list was written before the option existed, so priority/flex/default behavior silently differed between direct and proxy transport. The list is now a total map over `SimpleStreamOptions`: every option is marked serialized or client-local, and the request builder must list every serialized option. A new stream option now fails to compile until someone classifies it, so this class of drift cannot repeat.

Tests: `packages/agent/test/proxy.test.ts` pins both behaviors. Each was proven to fail with the fix reverted (truncated stream hangs to timeout; serviceTier assertion reports `undefined`).

Validation in a Prime sandbox: `packages/agent` vitest 66/66 pass, root `biome check` clean, root `tsgo --noEmit` clean. Root `npm run check` passes locally against `origin/main` as the test-policy base.

---

> [!NOTE]
> **Low Risk**
> Scoped proxy client fixes with tests; truncated streams now error instead of hanging, which aligns with the direct transport contract.
> 
> **Overview**
> Fixes two **proxy transport** bugs in `streamProxy`.
> 
> When the proxy SSE body ended without a **`done`** or **`error`** event, the stream closed without a terminal result, so **`result()`** never settled and turns could hang indefinitely. The client now tracks terminal events and, if the response ends early, fails with a **"Proxy stream truncated before completion"** error like direct transport.
> 
> **`serviceTier`** is included in the JSON sent to the proxy (it was previously omitted from the hand-maintained option list). Option forwarding is driven by an exhaustive **`PROXY_SERIALIZED_OPTIONS`** map over **`SimpleStreamOptions`**, so new shared options must be marked serialized or client-local at compile time instead of being dropped silently.
> 
> Adds **`packages/agent/test/proxy.test.ts`** for truncated-stream settlement and **`serviceTier`** serialization, plus brief agent/coding-agent changelog notes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e24445f57632345f2fd7136cea405c6ca8a4bd8a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### End truncated `streamProxy` streams and forward `serviceTier`
> - `streamProxy` in [proxy.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/2491/files#diff-fcc58abef0d544fa1f81af99ccc4d284f18cf45f0e77b88a8ff020b93a799892) tracks terminal `done` or `error` events. It emits an error with a stream-truncated message if the proxy response closes without one.
> - `buildProxyRequestOptions` now forwards the `serviceTier` stream option. `PROXY_SERIALIZED_OPTIONS` is an exhaustive map, requiring new shared options to be classified at compile time.
> - Behavioral Change: Truncated proxy streams now emit an error result and stop reason instead of ending silently.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized e24445f.</sup>
> 
>


sethkarten (approved):

PR #2499, opened 2026-09-21
fix(kernel): compare recorded Python skills by key and pin the marker swap

snimu:
No-Ticket: follow-up from the RSI review batch (findings F2/F3 of the kernel skill-sync review); no tracker issue exists.

## Problem

`pythonSkillsMatch` in `packages/coding-agent/src/core/kernel/bootstrap.ts` compared the recorded skill list against the caller list index by index. The two lists have different orders: `normalizePythonSkills` sorts the caller list by package path, while `syncPythonSkills` writes the marker in install order (`sortPythonSkillsForInstall`, dependencies first). When a skill depends on a sibling skill that sorts after it, the marker holds `[dependency, dependent]` and the caller list holds `[dependent, dependency]`. The ready gate never matched, so every session took the cross-process bootstrap lock, ran the skip-only skill sync (no uv work) and rewrote the marker. The zero-cost warm start from #2405 / #2409 never applied to that case.

The tmp+rename marker write (`writeBootstrapVersion`) had no tests. A plain `writeFile`, a removed retry loop, or a swallowed final error all passed the suite.

## Change

- `pythonSkillsMatch` now compares by `(importName, packagePath)` key: same number of skills, and every caller skill has a recorded entry with the same `pyprojectPath` and `pyprojectHash`. Order no longer matters. Existing markers on user machines match without a rewrite.
- A small `pythonSkillKey` helper replaces the six inline `` `${importName}\0${packagePath}` `` templates that already served as this key in the sync loop and merged marker write.
- Tests in `kernel-bootstrap.test.ts`, all inside existing tests plus one `rename` fault hook (`vi.mock("node:fs/promises")`, passthrough when idle):
  - the fast-path test now bootstraps a skill with a sibling dependency, pins the marker mtime after the sync and asserts it is untouched by a no-skill call and a later skills call;
  - the warm-venv test becomes the give-up pin: three failed swaps reject with the bootstrap error, the previous marker stays byte-identical, no `.bootstrap-version.tmp` is left;
  - the base-marker test arms two swap failures, so the base marker still lands before the first install through a retried swap.
- Changelog fragment `packages/coding-agent/.changes/kernel-skill-marker-ready-gate.md`.

## Mutants killed

| mutant | failing test |
|---|---|
| old index-by-index compare | `keeps a synced venv with a sibling dependency zero-cost across later and no-skill calls` (marker mtime changed) |
| plain `writeFile` instead of tmp+rename | `keeps the previous marker and surfaces the error when every marker swap fails` (resolves instead of rejecting) |
| `BOOTSTRAP_MARKER_SWAP_ATTEMPTS = 1` | `retries a busy marker swap before the first install and keeps a failed skill out of the record` (rejects) |
| `return` instead of `throw lastError` | `keeps the previous marker and surfaces the error when every marker swap fails` (resolves) |

## Existing tests changed

No coverage relative to `main` is lost; three existing tests host the pins:

- `preserves recorded Python skills when a no-skill bootstrap call reuses a warm venv` is rewritten into the give-up pin (`keeps the previous marker and surfaces the error when every marker swap fails`). Its behavior, a no-skill call on a warm venv leaves the recorded skills alone, stays pinned by the fast-path test below: a rewritten marker changes the mtime that test asserts, and the following skills call would re-sync and change the uv log.
- `keeps a skill-synced venv fast for real sessions after a no-skill bootstrap call` becomes `keeps a synced venv with a sibling dependency zero-cost across later and no-skill calls`: same flow and assertions, plus the sibling dependency and the marker mtime check.
- `writes the base marker before the first install and keeps a failed skill out of the record` becomes `retries a busy marker swap before the first install and keeps a failed skill out of the record`: same assertions, with two swap failures armed.

## Validation

Validated in a Prime sandbox: `tsgo --noEmit` clean, `biome check --error-on-warnings` clean, `test/kernel-bootstrap.test.ts` 22/22 (also with `--sequence.shuffle`); each mutant above fails only its named test. `check:test-policy` passes (12 net test lines against 14 source lines).

> [!NOTE]
> ### Compare recorded Python skills by key and retry marker swap in `bootstrap.ts`
> - Fixes marker validation to match recorded skills by composite key instead of positional array order, so venvs with sibling-dependency skills use the warm-start path instead of re-syncing
> - Adds a shared `pythonSkillKey` utility used at normalization, marker-merge, and sync call sites to replace inline key construction
> - Retries `node:fs/promises.rename` on transient errors (e.g. EBUSY) before and after skill install; on persistent failure the original marker is preserved and the temp file is cleaned up
> - Risk: `pythonSkillsMatch` now accepts markers whose install order differs from the caller's path order while still rejecting missing, extra, or metadata-mismatched skills — check [bootstrap.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/2499/files#diff-e87b9a04ab9b4ac02e8dd33af2f2f58320d8b7f6be6853acf4a3e05a34623cae) if any caller depended on positional ordering
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 47e1228.</sup>
> 
>

---

> [!NOTE]
> **Medium Risk**
> Changes kernel venv readiness and bootstrap marker semantics; wrong matching could skip needed syncs or accept stale markers, though existing on-disk markers should still match by key.
> 
> **Overview**
> Fixes kernel sessions that **re-ran skill sync and rewrote `.bootstrap-version` every time** when a Python skill depended on a sibling whose package path sorted differently than install order.
> 
> **`pythonSkillsMatch`** now treats the marker and caller lists as sets keyed by `importName` + `packagePath`, checking matching `pyprojectPath` and `pyprojectHash` without requiring the same ordering. **`pythonSkillKey`** centralizes that key for normalization, merged marker writes, and the sync loop.
> 
> **Tests** mock `fs/promises.rename` to simulate busy marker swaps: sibling-dependency venvs stay zero-cost on later and no-skill `ensureKernelPython` calls (uv log and marker mtime unchanged), exhausted swaps reject without corrupting the on-disk marker, and transient failures still allow the base marker before first install.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 47e12284791cdb4035e328310f5a628fb01f9b16. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


sethkarten (approved):

PR #2501, opened 2026-09-21
test(coding-agent): pin the RSI review findings within the test-line budget

snimu:
No-Ticket: test-only follow-up from the RSI review batch; no tracker issue exists for it.

## What this PR does

This PR adds missing regression tests. Five behaviors that shipped in recent merged PRs could currently be reverted without a single test failing; each one now has a test that fails if the fix is reverted. No production code is changed.

- **Compaction fingerprint** (#2400): after compaction, the summary is stamped with a fingerprint of the saved state. Resuming with unchanged state reuses that summary instead of stacking a new one, even when the rendered text has drifted.
- **Snapshot-clearing loop** (#2394): when new state is saved after a compaction, the old compaction snapshot is removed from the prompt instead of being left behind.
- **Same-instance reload cache drop** (#2414): reloading a session file on the same instance clears the in-memory branch cache, so the reload actually serves the new file's entries.
- **Catalog notice tolerance** (#2433): during worker recovery, if writing one "session interrupted" notice fails, recovery still finishes and every other session still gets its notice.
- **Header guard** (#2416): a damaged session file whose first entry is a tool result is rejected, even if a valid-looking session header appears later in the file.

All five checks are folded into existing tests that already build the same fixtures, and every original assertion of those tests is kept. That keeps the PR inside the repo's test-line budget (net test lines must not exceed source lines, and this PR adds no source lines) and makes it net negative overall: 42 test lines added, 47 deleted.

## Coverage deliberately not restored

Some gaps from those reviews are still left uncovered, cut to fit the budget, lowest-value ones first:

- **Catalog subprocess routing** (#2433): rename/archive/interrupt requests being routed through the fast append wrappers. Testing this takes ~33 lines plus a forked process for a two-line dispatch. The rename and interrupt wrappers are covered by other tests; the archive wrapper shares their code path and has no direct test.
- **Fast-append parent ID** (#2433): that a fast append computes the same parent entry as a full open. The only case not already covered elsewhere is handled by the existing rename test.
- **Old compaction snapshots** (#2400): snapshots written before #2400 carry no fingerprint, so the old text comparison decides and the digest gets re-delivered exactly once. Only transcripts compacted before #2400 are affected.
- **Supervisor log line** (#2433): the log line that names the session file whose notice was refused. The test checks that recovery completes, not what is logged.

Note for the merger: the first commit (f3649c80c) still carries an inert `Test-Budget-Exception:` trailer from before the rework; please leave that line out of the squash commit message.


sethkarten (approved):

PR #2504, opened 2026-09-21
fix(coding-agent): refresh the model catalog whenever the model picker opens

sirouk:
Opening the model picker with Ctrl+L or a bare `/model` shows the model catalog cached for up to 60 s (`MODEL_CATALOG_REFRESH_TTL_MS`). Models registered since the last fetch, for example by an extension calling `pi.registerProvider`, are missing until the cache expires. `/model <search>` already forces a refresh.

**Fix:** `showConfigurationMenu()` always calls `refreshModels(true)`. The picker still shows the cached list first and updates in place, so opening it does not block.

**Test:** a regression test in `2182-picker-selection.test.ts` opens the picker without a search on a fresh cache and expects `getModelCatalog` to be called and the new model to appear. It fails on the old line. The model picker and catalog tests pass (60/60), and `tsgo --noEmit` is clean.

> [!NOTE]
> ### Refresh model catalog whenever model picker opens in `InteractiveMode`
> Previously the model picker only refreshed the catalog when an initial search term was present. Now [interactive-mode.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/2504/files#diff-b96137f89d422953665b334785273ae6cdbcb53ff37d8828db43f85038658316) unconditionally requests a refresh when the picker opens, so the list is never stale.
>
> - Adds a regression test in [2182-picker-selection.test.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/2504/files#diff-5482049b9a6135dde2a8ecaaa5cfe59366d5a144316e1689506e84566b235d0a) confirming a newly returned model appears after refresh when opening with no search term.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized a7d6dbb.</sup>
> 
>


kevinjosethomas (approved):
lgtm, thanks @sirouk! do you mind fixing failing ci?

we're removing the tests LOC line CI but for now you can just remove the test

PR #2494, opened 2026-09-21
Remove duplicated Prime HTTP and daemon wire code

snimu:
No-Ticket: cleanup wave on main; duplicate-implementation findings from a read-only pattern audit, no tracker item.

## What this changes

Three duplicated implementations get one owner each. The goal is deletion: 253 lines removed, 247 added, most of the additions being the new owner module.

**1. Prime HTTP boundary helpers (new owner `packages/coding-agent/src/core/prime-http.ts`)**

`core/agent-traces.ts` and `core/prime-inference-auth.ts` each had their own `isRecord`, `stringField`, `numberField`, `stringEnv`, error-body parsing and fetch-with-timeout. Both now import one copy.

Resolved differences:

- Timeout error: the owner takes the timeout error from the caller. Agent traces still pass their typed `TraceUploadTimeoutError` (the retry classifier depends on it); Prime auth still passes `Error("Prime Inference request timed out")`.
- Abort reason: the owner uses the agent-traces semantics (abort the request with the timeout error, forward the caller abort reason). Prime auth maps any caller abort to `Error("Login cancelled")` through the optional `cancelledError`, which other auth code matches on by message.
- URL type: `string | URL` (the wider of the two).
- JSON fallback: one parse path that returns the raw text when the body is not a JSON object.
- The Prime auth timer is now `unref`ed, like the traces timer, so a pending request does not hold the event loop open.

Prime auth keeps a 6-line local wrapper (`fetchPrimeAuth`) that binds those two error objects, so its four call sites stay unchanged.

**2. Daemon client connect lifecycle**

`daemon-client.ts` and `daemon-worker-client.ts` had parallel socket-connect blocks (timer, listener cleanup, connect/error settling, destroy-on-timeout). Both now call `awaitSocketConnect`, exported from `daemon-client.ts` (the worker client already imports from that module, so this adds no new dependency edge).

`isDaemonResponse` was copied verbatim in both clients. It now lives in `daemon-protocol.ts`, which already owns the exported wire guards.

**3. SessionSummary guard**

The agents view accepted any `{id, sessionId}` as a complete `SessionSummary`; the supervisor also required `cwd`. The strict guard is now exported from `daemon-protocol.ts` and used by both. The weak copy is deleted.

## Wire compatibility

Backward compatible; nothing on the wire changed. No protocol version bump, no schema revision bump, no capability gate: the command and event unions, envelope shapes and compatibility maps are untouched. These are refactors plus two client-side validation/cleanup changes described below.

[...]

PR #2508, opened 2026-09-21
Release the daemon worker client socket after a failed connect

snimu:
No-Ticket: split of #2494 per review feedback ("seems like it should be multiple prs"); part 1 of 3, no tracker item.

## The bug

`DaemonWorkerClient.connect` set `this.socket` and `this.channel` before the socket connected, and left them set when the connect errored or timed out. The dead socket and channel stayed referenced until `close()`, and the next `connect()` on the same instance threw "Daemon worker client is already connected" instead of opening a fresh socket. The control-plane `DaemonClient` already cleared its reference on failure; the worker client did not.

Impact today is latent: both production callers (`daemon-supervisor.ts` `connectWorker` and `daemon-routed-client.ts`) build a fresh client per attempt and close it on failure, so nothing currently retries `connect()` on the same instance. The fix makes the worker client behave like the control client and removes the trap for the next caller that does.

## The fix

Both clients had a parallel inline "settle the socket connect" promise (timer, listener cleanup, connect/error settling, destroy-on-timeout). That block now lives once as `awaitSocketConnect` in `daemon-client.ts` (the worker client already imports from that module, so no new dependency edge). It takes an `onFailure` hook; `DaemonClient` passes its existing `clearSocketReference`, and `DaemonWorkerClient` gets the same `clearSocketReference` (new, 8 lines) so a failed connect releases the socket and channel.

No behavior drift for the control client: same order of cleanup, reference clearing, destroy and reject on timeout; same wrapped error on connect failure. The `error`/`close` listeners are still attached only after a successful connect, so no close notification changes.

## Test

One new test in `test/daemon-client.test.ts` ("drops the socket reference when the connect attempt fails, so the client can retry"): first connect fails, the retry reaches a fresh socket. It lives in that file because it reuses the file's `node:net` mock; there is no dedicated worker-client test file. Fails on main, passes here.

## Validation

- `scripts/check-test-policy.mjs` against origin/main - passes (18 net test lines vs ~44 source lines).
- vitest: `daemon-client`, `daemon-client-env`, `daemon-routed-client`, `daemon-worker-windows-timeouts`, `agent-connection-daemon` - 5 files, 155 tests, all passing.

Sibling PRs from the same split: https://github.com/PrimeIntellect-ai/prime-agent/pull/2509, https://github.com/PrimeIntellect-ai/prime-agent/pull/2510. Supersedes #2494.

> [!NOTE]
> ### Clear stale socket state on failed `DaemonWorkerClient.connect` attempts
> - Adds shared `awaitSocketConnect` utility to consolidate socket connect, timeout, and cleanup logic across daemon clients
> - `DaemonWorkerClient.connect` now clears its socket and channel references when a connect fails or times out, via the new `clearSocketReference` method, so a later retry creates a fresh socket instead of throwing "already connected"
> - Refactors `DaemonClient.connect` to use the same `awaitSocketConnect` helper, preserving its existing timeout and error messages
> - Behavioral Change: failed or timed-out worker connections now destroy the failed socket and reset `DaemonWorkerClient` state; previously the stale socket remained and blocked retries
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized ae8c02c.</sup>
> 
>


sethkarten (approved):

PR #2509, opened 2026-09-21
Move daemon response and session-summary wire guards into daemon-protocol.ts

snimu:
No-Ticket: split of #2494 per review feedback ("seems like it should be multiple prs"); part 2 of 3, no tracker item.

## What this changes

Two wire guards had more than one definition in the daemon client/supervisor/agents-view code:

- `isDaemonResponse` was copied verbatim in `daemon-client.ts` and `daemon-worker-client.ts`. It now lives in `daemon-protocol.ts`, which already owns the exported wire guards, and both clients import it.
- `isSessionSummary` existed twice with different strictness: the supervisor required `id`, `sessionId` and `cwd`; the agents view accepted any `{id, sessionId}` and produced a typed `SessionSummary` with a missing required field. The strict guard moves to `daemon-protocol.ts` and both the supervisor and the agents view use it. The weak copy is deleted. The agents view keeps its local `isRecord` (it has other users).

`cli/daemon-command.ts` keeps its own, stricter summary check for CLI output (it validates many more fields); it is unchanged.

## Wire compatibility

Nothing on the wire changed. No protocol version bump, no schema revision bump, no capability gate.

## Behavior change / question for reviewers

The agents view now rejects session summaries without `cwd` (it raises "Daemon returned an invalid session summary", the same as for any other malformed summary). Every current source fills `cwd`: worker summaries come from `summaryForActiveSession` (`sessionManager.getCwd()`) and already pass the supervisor's strict guard before reaching a client; inactive summaries come from `summaryForInactiveSession`, and `session-manager.ts` always fills `SessionInfo.cwd` with a string. So this should only reject malformed data. Question: is there any path you know of that ships a summary without `cwd` (older daemon, partial roster entry) where the view should degrade instead of raising?

## Test

One new test in `test/daemon-protocol.test.ts` pins the strict guard: a summary with `cwd` passes, one without does not.

## Validation

- `scripts/check-test-policy.mjs` against origin/main - passes (6 net test lines vs ~23 source lines).
- vitest: `daemon-protocol`, `daemon-client`, `daemon-routed-client`, `daemon-supervisor-monitor`, `agents-view-mode`, `agents-view-state`, `agents-view-roster`, `agents-view-empty-sessions`, `4606-update-restart-coordinator` - 9 files, 343 tests, all passing.

Sibling PRs from the same split: https://github.com/PrimeIntellect-ai/prime-agent/pull/2508, https://github.com/PrimeIntellect-ai/prime-agent/pull/2510. Supersedes #2494.

> [!NOTE]
> ### Move daemon response and session-summary wire guards into `daemon-protocol.ts`
> - Consolidates the duplicate `isDaemonResponse` and `isSessionSummary` guards from the daemon client, worker client, supervisor, and agents view into shared helpers in [daemon-protocol.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/2509/files#diff-a45bb1297653325fee7a1902ea1e109747510137380c53e92cbbe113436f0d29)
> - The shared `isSessionSummary` now requires a string `cwd` field; the agents view's prior local guard only checked `id` and `sessionId`, so it now rejects session summaries without a working directory
> - Behavioral Change: agents view now rejects session-summary payloads missing a string `cwd`; daemon client, worker client, and supervisor guard behavior is unchanged
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 4ec137a.</sup>
> 
>


sethkarten (approved):

PR #2510, opened 2026-09-21
Move shared Prime HTTP helpers into core/prime-http.ts

snimu:
No-Ticket: split of #2494 per review feedback ("seems like it should be multiple prs"); part 3 of 3, no tracker item.

## What this changes

`core/agent-traces.ts` and `core/prime-inference-auth.ts` each had their own copy of `isRecord`, `stringField`, `numberField`, `stringEnv`, error-body parsing and a fetch-with-timeout. This PR gives them one owner, the new `core/prime-http.ts`, and both files import it. Net: one new 104-line module, ~160 lines of duplicates deleted.

Resolved differences between the two copies:

- Timeout error: the owner takes the timeout error from the caller. Agent traces still pass their typed `TraceUploadTimeoutError` (the retry classifier depends on it); Prime auth still passes `Error("Prime Inference request timed out")`.
- Abort reason: the owner uses the agent-traces semantics (abort the request with the timeout error, forward the caller abort reason). Prime auth maps any caller abort to `Error("Login cancelled")` through the optional `cancelledError`, which other auth code matches on by message.
- URL type: `string | URL` (the wider of the two).
- JSON fallback: one parse path that returns the raw text when the body is not a JSON object (same result as before for both callers).

Prime auth keeps a 6-line local wrapper (`fetchPrimeAuth`) that binds those two error objects, so its four call sites stay unchanged.

## Behavior change

The Prime auth timeout timer is now `unref`ed, like the traces timer, so a pending request does not hold the event loop open by itself. No user-visible change expected.

## Tests

No new tests. The existing `prime-inference-auth`, `login-dialog` and `agent-traces` suites already cover the cancel, abort and error-body paths that the shared helper now serves.

## Validation

- `scripts/check-test-policy.mjs` against origin/main - passes (0 net test lines vs ~93 source lines).
- vitest: `agent-traces`, `prime-inference-auth`, `login-dialog` - 3 files, 62 tests, all passing.

Sibling PRs from the same split: https://github.com/PrimeIntellect-ai/prime-agent/pull/2508, https://github.com/PrimeIntellect-ai/prime-agent/pull/2509. Supersedes #2494.

> [!NOTE]
> ### Move shared Prime HTTP helpers into `core/prime-http.ts`
> - Extracts duplicated Prime HTTP utilities (record/field validation, response parsing, timeout handling) from `agent-traces.ts` and `prime-inference-auth.ts` into the new [prime-http.ts](https://github.com/PrimeIntellect-ai/prime-agent/pull/2510/files#diff-cbcb7cfb825ad1d59ccee74988f45ab02c99ceecdd07018f108f0cbbbaffc2a3) module.
> - Updates trace uploads and Prime Inference auth requests to import the shared helpers instead of using local copies.
> - Behavioral Change: shared `fetchWithTimeout` calls `unref` on its timeout timer, so a pending Prime HTTP request no longer keeps the Node.js process alive by itself. Timeout and cancellation errors still map to the existing Prime-specific errors at the HTTP boundary.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 999cc09.</sup>
> 
>


sethkarten (approved):

